PR #35, opened 2014-04-29
Significant performance and memory improvement in cleanContent

calguy1000:
I encountered a problem with parsing an arcGis generated PDF File that contained one page and numerous images, but no text (per say).  the cleanContent routine would take minutes to do it's task and timeout.

Numerous tests indicated that
a: preg_match_all with PREG_OFFSET_CAPTURE would not always return an array of 2 elements (so the offset was sometimes not returned)... which would result in the corruption of the beginning of $content instead of the cleaning of the regex.
b: the foreach loop and substr_replace were extremely slow and memory intensive.
c: using preg_replace_callback with a simple closure was extremely more optimal (down to ~2 seconds from ~166 seconds) and significantly less RAM.

Please check my code.

### DIFF
--- a/src/Smalot/PdfParser/Object.php
+++ b/src/Smalot/PdfParser/Object.php
@@ -135,21 +135,19 @@ class Object
         $char    = $char[0];
         $content = str_replace(array('\\\\', '\\)', '\\('), $char . $char, $content);
 
-        // callback function to clean text
-        // callback is memoory efficient, and faster than a foreach loop.
-        $_clean = function($part) use ($char, &$_count) {
+        // this function is more memory efficient, and significantly faster than
+        $_clean = function($part) use($char) {
             return str_replace($part[1],str_repeat($char,strlen($part[1])), $part[0]);
         };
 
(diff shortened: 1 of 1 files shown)


camlafit:
Hi

This patch improve my use case. I do some import text to sphinx (about 5k files). 
I can analyze more files with this patch but I obtain in some case an error on Object.php on line 224  with an Error memory exhausted.

rijvirajib:
What's the status of this merge?

keesschepers:
I would also like to know the status. I haven't tested it yet but it's likely going to solve a lot of problems we are currently facing with memory consumption.

smalot:
Unfortunately this shortcut contains some errors.
If you have strings with escaped round brakets, I'm afraid it won't work.

keesschepers:
Could you explain more in detail what will go wrong? The regular expressions look the same, I am not seeing where it should go wrong. Did you test it?

k00ni:
Closed due to no activity since 5 years. Please reopen, if you wanna solve it.

PR #46, opened 2014-09-18
see my pull request

waldemarnt:
I have a problem with varius pdf types, and i need change this, to make it functionally, plz see this pull request.


smalot:
Partially integrated

PR #50, opened 2014-10-29
add getTextArray methods to Page.php and Object.php -

dslawrence:
to enable retrieving text in separate array elements rather than one long string. This will make it easier for people who are extracting data from PDFs to distinguish one text snippet from another.


smalot:
Thanks

PR #55, opened 2015-01-15
Handle circular references in xObjects

### LINKED ISSUE #20: getText() -> Maximum function nesting level of '100' reached, aborting!
Hi,

Today I received feedback about a failing PDF. It generated a function nesting error.
I wrote a small patch that masked the problem, which is kinda dirty.

I can only upload images to a ticket, so I uploaded the files somewhere else:
The error call stack: http://nederland.ligtin.eu/sthoolen/20140217-pdfparser-error.html
The dirty patch: http://nederland.ligtin.eu/sthoolen/20140217-pdfparser-quickfix.patch
The PDF: http://www.vgogh.nl/documenten/huwelijkse_voorwaarden/files/assets/downloads/publication.pdf

Hopefully this problem can be dealt with in a somewhat more decent way? ;-)

aik099:
Sometimes it happens that elements on a page have reference to themselves.

Example PDF that have circular reference that causes infinite loop: https://www.dropbox.com/s/ks78qbatmik1d2k/nesting_level_issue.pdf?dl=0

Closes #20

### DIFF
--- a/src/Smalot/PdfParser/Object.php
+++ b/src/Smalot/PdfParser/Object.php
@@ -244,6 +244,7 @@ class Object
         $current_font        = new Font($this->document);
         $current_position_td = array('x' => false, 'y' => false);
         $current_position_tm = array('x' => false, 'y' => false);
+        $object_hash         = spl_object_hash($this);
 
         foreach ($sections as $section) {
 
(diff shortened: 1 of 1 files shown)


smalot:
Thanks

PR #59, opened 2015-03-17
Parser.php - added object buffering to avoid memory exhaustion

mc-comanescu:
This fixes an issue when parsing larger pdfs, and runs out of memory.

### DIFF
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -84,9 +84,11 @@ class Parser
     public function parseContent($content)
     {
         // Create structure using TCPDF Parser.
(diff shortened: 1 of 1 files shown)


aik099:
How exactly this helps. Output buffering would only help if we have some output during PDF parsing, but we don't.

mc-comanescu:
Hi,
It's a hack that I think allows php to flush some of the memory out. By using ob_start you tell the interpreter you will not ouput to the webserver. Thus php can optimize/flush what it currently outputted so far, and make memory available for what it needs to do next. Furthermore since the object doesn't produce output, and it's buffered, the interpreter can take shortcuts in execution. I just erase the buffer, once the heavy part of the execution finishes and resume normal outputting.

PR #67, opened 2015-05-26
Indirect recursion detection in xObjects

aik099:
In #55 I've implemented basic recursion protection that catched `A > A` cases (object referencing itself).

Recently I've stumbled on PDF document that had `A > B > A > B > A` case (object referencing himself through another object).

This PR fixes that case.


smalot:
Nice patch !
Thanks

Issue #73, opened 2015-07-20
Call to undefined method Smalot\PdfParser\Object::getPages()


black-silence:
Fatal error in smalot/pdfparser/src/Smalot/PdfParser/Document.php line 220

This is the first and only case I found. PDF loads fine in zathura so I don't think it's broken.
PDF probably generated by a software called PaperPortPDF. 
%PDF-1.5

Can't give you access to the pdf (of course... meh) but I can give you debug output if required.

MarcusSchwarz:
This definitely needs some more attention, I have the same problem here

rastapopougros:
I have a similar error but it's for `Call to undefined method Smalot\PdfParser\Encoding::getPages()` and it's on`/src/Smalot/PdfParser/Document.php on line 218`.

So it's on this line :
```php
/*216*/ /** @var Pages $object */
/*217*/ $object = $this->objects[$id]->get('Pages'); 
/*218*/ $pages  = $object->getPages(true);
```

The comment tells us it would be a `Pages` object… But it seems not true, the fatal error says `Encoding`.

Is `$object` could be an `Encoding` object ?
And inside the class Encoding, there is absolutely no method called `getPages()`.

black-silence:
Same line. I guess $id is wrong but I haven't debugged enough to check why

rastapopougros:
Hello, 
So, there was a commit to fix this "wrong object" bug ?

smalot:
Can you test again with the current master branch ?

denismusto:
Hi, I solved this problem commenting the follow code:  

/*if (mb_check_encoding($text, "UTF-8")) {
     $unicode = true;
}*/

In \Smalot\PdfParser\Font.php, line 371 - 373.

PR #74, opened 2015-07-20
workaround for wrong object returned

black-silence:
workaround for issue #73


black-silence:
Travis fail looks like composer problem, crashed before checking my change. Not fair.

PR #81, opened 2015-09-17
Package tecnick.com/tcpdf is abandoned

j0k3r:
As `composer up` says:

> Package tecnick.com/tcpdf is abandoned, you should avoid using it. Use tecnickcom/tcpdf instead.


smalot:
Thanks

PR #83, opened 2015-09-24
Using the first font from the document as default font

zCarot:
I have PDF from FastReport that can't be parsed with `new Font($this->document)`
But parser works fine with first font object included in PDF

I know that is not best solution, but it's better than current blank font.

P.S. If you need my pdf file for testing, I can send it privately as it contains personal information.


smalot:
Thanks

PR #87, opened 2015-11-12
Spaces in tables, fixes #76

### LINKED ISSUE #76: No spaces inserted when cursor is moved horizontally
I have a pdf that shows text that is aligned in columns (probably a pdf export of an excel file):

| First Header | Second Header |
| --- | --- |
| Cell 1 | Cell 2 |
| Cell 3 | Cell 3 |

There is no character spacing between cell 1 and cell 2 causing the text to be output as:

> First headerSecond header
> Cell 1Cell 2
> Cell 3Cell 4

The spacing in the pdf is apparently done using the Text Matrix "Tm" command. I adjusted the "Tm" handling to take a change in the x coordinate into consideration to get spacing between the cells.

Object.php / getText ->

>  case 'Tm':
>    $args = preg_split('/\s/s', $command[self::COMMAND]);
>    $y    = array_pop($args);
>    $x    = array_pop($args);
> 
>  // ADDED START
>  if ($current_position_tm['x'] !== false) {
>    $delta = abs(floatval($x) - floatval($current_position_tm['x']));
>    if ($delta > 10) {
>      $text .= " ";
>    }
>  }
>  // ADDED STOP
> 
>  if ($current_position_tm['y'] !== false) {
>    $delta = abs(floatval($y) - floatval($current_position_tm['y']));
>    if ($delta > 10) {
>      $text .= "\n";
>    }
>  }
>  $current_position_tm = array('x' => $x, 'y' => $y);
>  break;

Would that be a correct way to fix the problem?

Stichoza:
Fix by @archonia provided in #76. This fix works for me as well. So maybe we could merge this if everything still works fine? I haven't tested a lot, but there was no problem for me while 3-4 different files with tables.

Anyways, we can discuss it :)

fixes #76

### DIFF
--- a/src/Smalot/PdfParser/Object.php
+++ b/src/Smalot/PdfParser/Object.php
@@ -328,6 +328,12 @@ class Object
                         $args = preg_split('/\s/s', $command[self::COMMAND]);
                         $y    = array_pop($args);
                         $x    = array_pop($args);
+                        if ($current_position_tm['x'] !== false) {
+                            $delta = abs(floatval($x) - floatval($current_position_tm['x']));
+                            if ($delta > 10) {
+                                $text .= " ";
+                            }
+                        }
                         if ($current_position_tm['y'] !== false) {
                             $delta = abs(floatval($y) - floatval($current_position_tm['y']));
                             if ($delta > 10) {


AlexKovalevych:
@Stichoza thank you for your PR, but in some cases space won't help to parse the table. Here is an example of a single line from the pdf table:

```
0.11 12 310 -200.17 -200.17
```

Its not possible to figure out how many columns we have here, since column values can have spaces in them. Consider using a different character to separate columns, maybe `\t`:

```
0.11	12 310	-200.17	-200.17
```

That way you can easily parse the line, like `explode("\t", $line);` and get columns correctly.
I've created an appropriate PR, which is close to this one: https://github.com/smalot/pdfparser/pull/97

k00ni:
@AlexKovalevych: You mentioned your PR #97. How close is it to this PR?

rubenvanerk:
It is extremely close. Merged #97 uses a `"\t"` as a separator instead of a space which this one uses.
IMO, #97 makes this PR obsolete.

k00ni:
Thats a tricky one. So its about the text output after parsing the PDF. We have the option to either separate table cells using a white space or tab.

Maybe a tab would be better, because you could search for it later on and use it to work with the data of the table. A white space won't allow any differentiation. So i approve your point and mark this obsolete too.

Issue #71, opened 2015-06-26
Stops with no error indication when low memory limit


mazhuravlev:
Stops on Parser::parseFile() when file is big and PHP memory_limit is low. No error is indicated.

aik099:
The actual parsing is done by TCPDF and large file isn't what eats memory, but how file is structured internally. In general it would be a good idea to increase memory limit in any case before doing PDF parsing.

srakas:
I've got the same error while parsing PDF file of size 115KB, while memory limit is set to 1GB.

When I debug the code, it seems to be happening inside TCPDF when method getRawObject() is being called. Any suggestions how to solve this?

keesschepers:
We have a pretty similar problem, even when setting memory limits up to 8GB. We are building a webpage crawler which should also extract PDF documents with your library but especially with big PDF's (like 40mb) it's crashing on Smalot/PdfParser/Object.php on line 143 which is the preg_match_all. I assume that too many occurrences are found and the array consumes too much memory.

Is it possible to split up the content of the document into chunks and then ready every X bytes and cleanup stuff or is there a risk of data corruption in that case?

lkalinka:
Here is example of PDF that is causing such problems:
 http://www.cranendonck.nl/document.php?m=64&fileid=42797&f=6e910448e0030fb42fbfcc848a7e6258&attachment=0&c=33386 
its very small PDF but it can eat whole available memory (8GB) when i trying to read(parse) this file, that make server crash.

Could you take a look and reproduce/fix that ?

marathonstudios:
This issue has popped up both times I've used this project in the last month - a fix is really essential.

StanProg:
> I know this is super old, but could any of you give [#787](https://github.com/smalot/pdfparser/pull/787) a try? Thanks in advance 🙏

I tried it, but there was no improvement.

File size: 1454805 bytes (1.4 MB)
Bytes taken from parseContent()/parseFile(): 288030720 (288 MB) + Out of memory

PR #92, opened 2016-01-13
Fixing infinite loop caused by wrong XREF reference, migration to newer bundle

lkalinka:
Migration to bundle that will overwrite old one - suggested by author there: https://sourceforge.net/p/tcpdf/discussion/435311/thread/996c7864/
Fix for parsing pdf files tht described in this ticket: https://github.com/smalot/pdfparser/issues/71

### DIFF
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -39,7 +39,7 @@ use Smalot\PdfParser\Element\ElementNull;
 use Smalot\PdfParser\Element\ElementNumeric;
 use Smalot\PdfParser\Element\ElementString;
 use Smalot\PdfParser\Element\ElementXRef;
-use Com\Tecnick\Pdf\Parser\Parser;
+use Com\Tecnick\Pdf\Parser\Parser as TCPDF_Parser;;
 
 /**
  * Class Parser
@@ -86,7 +86,7 @@ class Parser
     {
         // Create structure using TCPDF Parser.
         ob_start();
-        @$parser = new Parser();
+        @$parser = new TCPDF_Parser();
         list($xref, $data) = $parser->parse(ltrim($content));
         unset($parser);
         ob_end_clean();
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -39,6 +39,7 @@ use Smalot\PdfParser\Element\ElementNull;
 use Smalot\PdfParser\Element\ElementNumeric;
 use Smalot\PdfParser\Element\ElementString;
 use Smalot\PdfParser\Element\ElementXRef;
+use Com\Tecnick\Pdf\Parser\Parser;
 
 /**
  * Class Parser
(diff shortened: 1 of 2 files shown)


lkalinka:
Hmm this new library comes with php > 5.3, so there should be a 7 version that will accept that, maybe?

marathonstudios:
This change fixes some instances of the problem, but not all. I also noticed that the new version of the parser fails to return many of the metadata fields that the old parser was able to find.

smalot:
Happy to discover the "tecnickcom/tc-lib-pdf-parser" library.
Another good work of Nicola Asuni

smalot:
Needs more work   ;-)

nuclearmedia:
What's the status on this?

src/Smalot/PdfParser/Parser.php
```diff
@@ -85,8 +86,8 @@ public function parseContent($content)
     {
         // Create structure using TCPDF Parser.
         ob_start();
-        @$parser = new \TCPDF_PARSER(ltrim($content));
-        list($xref, $data) = $parser->getParsedData();
+        @$parser = new Parser();
```
nuclearmedia: This class is already called Parser. This is why this is failing.

src/Smalot/PdfParser/Parser.php
```diff
@@ -85,8 +86,8 @@ public function parseContent($content)
     {
         // Create structure using TCPDF Parser.
         ob_start();
-        @$parser = new \TCPDF_PARSER(ltrim($content));
-        list($xref, $data) = $parser->getParsedData();
+        @$parser = new TCPDF_Parser();
```
nuclearmedia: This is also already in use.

Hackmancoltaire:
What's the status on this fix? It's been some months, and this seems to be breaking larger file parsing.

joncojonathan:
Is there a status on this fix please?

smalot:
This fix is not working.
The returned parsed data is currently not compatible with the previous parser.
Issue seems to be located on `XrefStream` class at line 292.

Don't forget to use atoum unit test command line:
```
./vendor/bin/atoum -d src/Smalot/PdfParser/Tests/ -ncc
```

k00ni:
I can't decide if including the new library makes the whole situation better or worse, after reading comments here and on https://sourceforge.net/p/tcpdf/discussion/435311/thread/996c7864/. 

There i vote **against** these changes, because:

* @Hackmancoltaire wrote: `It's been some months, and this seems to be breaking larger file parsing.`
* @JonathanZWeber wrote: `I also noticed that the new version of the parser fails to return many of the metadata fields that the old parser was able to find.`
* TCPDF's successor, TC-Lib-PDF, is also not actively maintained anymore (see https://github.com/tecnickcom/tcpdf#important)

Ref:  #291

I leave this pull request open for a while, but will close it if there is no activity.

k00ni:
We remove TCPDF and ported some of its code. FYI: #299

Thank you for your efforts!

Issue #95, opened 2016-03-04
Wrong Unicode Detection


sparx82:
Hi

I have a PDF here (generated from my companies SAP system) which has hexadecimal encoded text but IS NOT unicode. This leads to the effect that the output of the parser is Windows-1252 coded text while it should be UTF-8.

Some more details:
Font.php line 361 sets $unicode = true for hexadecimal coded text but this is not true for my pdf. If comment this line, my PDF is parsed correctly as the conversion is then done by Font.php line 491.

`$this->has('ToUnicode')` is false
`$this->has('Encoding')` is true

```
$encoding = $this->get('Encoding');
if ($encoding instanceof Encoding)
```

is false

So everything `decodeContent` should do is encode the text to Unicode but this is prevented by $unicode beeing true.

I don't know the PDF standard well enough to say whether this is a bug in the pdfparser or the PDF is not according to the standard at all.

Unfortunately I cannot attach the PDF because it contains pretty private data and all my attempts to create a similar one using Python failed so far. But I keep trying. And I'm happy to help with debug information if necessary.

sparx82:
Ok, I was able to create a PDF which has the same problem.

Stream ist `<E6F6FC>` and it should look like this:
![image](https://cloud.githubusercontent.com/assets/17567952/13546280/54415574-e2a9-11e5-92b7-e41658abdb44.png)

 And this PDF causes the problem:
[SpecialCharTest.pdf](https://github.com/smalot/pdfparser/files/159810/SpecialCharTest.pdf)

smalot:
Hi sparx82,
Should your patch fix this issue ?
Could you update too the unit test classes ?
Many thanks for your patch

sparx82:
Hi smalot,

I added my test PDF and extended the tests to check for correct decoding of ANSI/Unicode PDF content.
The patch should fix the issue, at least as good as the mb_check_encoding PHP function works. Which should be ok according to the comments on the PHP site.

sparx82:
Updated pull request:
https://github.com/smalot/pdfparser/pull/96

Golmote:
Hi! You should close this issue if #96 fixed it.

PR #96, opened 2016-03-05
Fix for Issue #95

sparx82:
Check if `$text` is really Unicode before setting `$unicode = true`


smalot:
Many thanks for your work !

PR #97, opened 2016-03-09
separate table columns with tabs on parsing

AlexKovalevych:


defkrie:
@AlexKovalevych thanks for your commit, its helpful in my case

smalot:
Thanks guys

PR #114, opened 2016-09-09
Fix on Page->getText() and Page->getTextArray()

gbrlmza:
In some PDFs, the content of the page is an instance of ElementNull, in those cases getText() and getTextArray() fails on getting the text: `Call to undefined method Smalot\PdfParser\Element\ElementNull::getText()`

Sample PDF with this kind of empty pages attached
[Sample-PDF-with-empty-pages.pdf](https://github.com/smalot/pdfparser/files/464316/Sample-PDF-with-empty-pages.pdf)


smalot:
Many thanks

PR #140, opened 2017-02-26
Update to newer atoum versions

jubianchi:
atoum 2.8 allows us to remove the bootstrap when it's only used to
require the autoloader (see atoum/atoum#605)

atoum 3.0 allows us to run tests on latest PHP versions.


smalot:
Thanks

PR #155, opened 2017-06-19
Added Document getTextArrayByPageAndSection

motin:
I was missing a way to remove parts of headers and footers as well as to detect natural separation of content. By including the section index in the returned array, natural separation of text content / sections are found, and by including the page index, it is easier to remove parts of headers and footers.


k00ni:
Closed due to no activity.

PR #170, opened 2017-09-29
Parser fix

dotomaz:
I had problems with some pdf-s on HPUX. This regex caused a core dump and the entire request died. If I change the regex with a simple parser it works.

### DIFF
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -170,7 +170,13 @@ class Parser
                         $match = array();
 
                         // Split xrefs and contents.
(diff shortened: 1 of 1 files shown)


smalot:
Can you provide the $content value when you encountered a core dump ?
I'm not sure it's a good idea to replace a regex with a simple work around.

dotomaz:
[example.txt](https://github.com/smalot/pdfparser/files/1348182/example.txt)

It dies on this one. This is the fourth stream part in the pdf. First three pass without error. I can't upload the entire pdf because it contains personal information.

This only hapens on HPUX. It work ok on windows. Haven't tried linux.

smalot:
I don't understand why a core dump occurs on HPUX.
Not sure it's a good idea to replace a preg_match by a simple strpos ...

PR #180, opened 2017-12-12
Bugfix: Call to undefined function Smalot\PdfParser\mb_check_encoding()

kstenschke:
Fixed error: Error: Call to undefined function Smalot\PdfParser\mb_check_encoding()
see also: https://github.com/smalot/pdfparser/issues/73


smalot:
`mb_check_encoding` is a native php function.
It seems your php is not correctly installed or a php module is missing

doganoo:
@smalot maybe you can add a check for the extension? Either via composer or directly in the source code

smalot:
Which extension should be checked in composer ?

PR #184, opened 2018-01-30
page object not null verification

doganoo:
I have a weird case where the pdf document is valid but the page not. The page instance in the pages array is null. Therefore, i propose to verify the instance before calling the "getText()" method.

### DIFF
--- a/src/Smalot/PdfParser/Document.php
+++ b/src/Smalot/PdfParser/Document.php
@@ -257,6 +257,16 @@ class Document
         $pages = $this->getPages();
 
         foreach ($pages as $index => $page) {
+            /*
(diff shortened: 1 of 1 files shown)


klausi:
Found the same problem. Symptom:
```
PHP Fatal Error: Call to a member function getText() on null in Smalot\PdfParser\Document->getText() (line 260 of vendor/smalot/pdfparser/src/Smalot/PdfParser/Document.php)
```

This is kind of a hotfix, the `getPages()` method should never return NULL pages.

doganoo:
@klausi yes you are right, but at that time a hotfix was ok in my case.

It seems like that @smalot does not maintain this library anymore. Thats why I have forked this project and merged my pull request: https://github.com/doganoo/pdfparser

Issue #61, opened 2015-03-29
Empty page, nothing parsed


rvdsarr:
Hey guys,

I have a PDF, and whenever I try to parse it, it comes back as an string with a few whitespaces. Same happends on the demo page. 

I'd like to share the PDF, but just with the dev's if thats alright. 

Thanks.

Roy

AlexKovalevych:
Having similar problem. After debug i figured out that it silently failed here:

```
 [Symfony\Component\Debug\Exception\ContextErrorException]
  Notice: iconv(): Wrong charset, conversion from `Windows-1252' to `UTF-8//TRANSLIT//IGNORE' is not allowed
```

Trace:

```
Exception trace:
 () at /var/www/vendor/smalot/pdfparser/src/Smalot/PdfParser/Font.php:498
 Symfony\Component\Debug\ErrorHandler->handleError() at n/a:n/a
 iconv() at /var/www/vendor/smalot/pdfparser/src/Smalot/PdfParser/Font.php:498
 Smalot\PdfParser\Font->decodeContent() at /var/www/vendor/smalot/pdfparser/src/Smalot/PdfParser/Font.php:386
 Smalot\PdfParser\Font->decodeText() at /var/www/vendor/smalot/pdfparser/src/Smalot/PdfParser/Object.php:318
 Smalot\PdfParser\Object->getText() at /var/www/vendor/smalot/pdfparser/src/Smalot/PdfParser/Page.php:217
```

Not sure what to do with that error, yet..

gud:
I've the same problem. Each time a pdf file contains at least one blank page the parser stop and nothing is returned to the caller script, no error either. This is really blocking because it is not possible to ignore the file neither taking any action to bypass the file.

AlexKovalevych:
I figured out the problem - i've used alpine `php-iconv` package which is obviously not working well. After i built from the source `libiconv` library and rebuilt `php-iconv` everything worked well.

gud:
This not works on an apache 2.4.9 /php 5.5.12 server on windows 7 (WAMP) BUT works well on a Linux server with PHP Version 5.3.29

mplx:
I ran into the same problem when switching from a Debian host to a alpine based docker container.

[...]

PR #191, opened 2018-03-15
Changed from iconv to mb_convert_encoding

AndreasFurster:
See https://github.com/smalot/pdfparser/pull/168
And https://github.com/smalot/pdfparser/issues/61

I couldn't test all cases, but it worked in my case and should work in others.


smalot:
Done with another merge request.
Can you check the master branch ?
Thanks for your interest

PR #199, opened 2018-05-19
PHP fatal error: Call to a member function getContent() on null in Smalot\PdfParser\Font->loadTranslateTable() (line 141 of src/Smalot/PdfParser/Font.php).

klausi:
The problem is that Header::resolveXRef() sometimes can return null, but all callers expect that an Element comes out. We can return an ElementMissing to resolve that.


klausi:
And I had to fix another PHP fatal error, see the last commit. This is now the 4th fatal error I had to fix in this library, hopefully that is all now :-)

PR #205, opened 2018-06-20
removing error control operator and updated PHPDoc

doganoo:
as multiple users complained that the parseFile() method dies silently, I have removed the error control operator @ and added information about exceptions in PHPDoc.


smalot:
Thanks !

Issue #18, opened 2014-02-15
Input not checked enough PdfParser/Font.php L253


kaystrobach:
```
PHP Warning:  pack(): Type H: illegal hex digit y in smalot/pdfparser/src/Smalot/PdfParser/Font.php on line 253
```

With some documents i get tons of this warnings, with others none ...

robotrobot:
Yup, getting this warning a lot

wackerm:
can confirm this warnings..

arunkumars3483:
Still issue exists. pack(): Type H: illegal hex digit v

lucas-gp:
I get the same problem 

The arunkumars3483 solution solve the problem for me too, thanks you @arunkumars3483 

Please update :+1:

Pesche007:
Had the same issue. No idea what the code of arunkumars3483 does - but it fixes the issue for me.

PR #224, opened 2019-01-13
fix illegal hex

waspinator:
issue #18


xelan:
The change proposed in #18 and applied here breaks the part name parsing, as the actual info is lost during the process.
crc32() returns an integer (checksum of the part) which is formatted... as an integer. And then any spaces occuring in the integer (duh?) are replaced with nothing before passing the result to the hex packing... :confused: 
```
$part = pack('H*', str_replace(' ', '', sprintf('%u', CRC32($part))));
```

PR #231, opened 2019-03-21
Add default timezone

rbairwell:
According to https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf page 87 section '7.9.4 Dates', date/timestamps should be, by default, in UTC/GMT. If no timezone is specified to \DateTime::createFromFormat it defaults to the system date/time which could pose problematic.


smalot:
Thanks Richard

PR #237, opened 2019-04-06
Update Parser.php

### LINKED ISSUE #233: Invalid type: ">>".
With a lot of PDFs gettings this error. When open the PDF and resave it (with Adobe Acrobat), the problem is gone..

smalot:
Fix #233


k00ni:
What is the status here?

PR #245, opened 2019-05-13
Fixed issue when encoding can not be casted to string.

leitom:
A lot of the pdf's I have tried to parse will fail when the encoding can not be converted to a string.

### DIFF
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -86,7 +86,12 @@ class Font extends PDFObject
 
         $details['Name']     = $this->getName();
         $details['Type']     = $this->getType();
-        $details['Encoding'] = ($this->has('Encoding') ? (string)$this->get('Encoding') : 'Ansi');
+
+        try {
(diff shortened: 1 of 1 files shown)


davispuh:
This is fixed with #257 PR. It's because `Encoding` property was `Header` class and not `Encoding` class

src/Smalot/PdfParser/Font.php
```diff
@@ -86,7 +86,12 @@ public function getDetails($deep = true)
 
         $details['Name']     = $this->getName();
         $details['Type']     = $this->getType();
-        $details['Encoding'] = ($this->has('Encoding') ? (string)$this->get('Encoding') : 'Ansi');
+
+        try {
+            $details['Encoding'] = ($this->has('Encoding') ? (string) $this->get('Encoding') : 'Ansi');
+        } catch (\Exception $e) {
+            $details['Encoding'] = 'Ansi';
```
k00ni, line 93: I dislike the try-catch approach, because it will hide any exception thrown. One will never know if an exception was thrown was relevant or not.

k00ni:
Solved by #318.

PR #256, opened 2019-09-04
Update PDFObject.php

payal-indpro:
To avoid extra spaces while getting the text

### DIFF
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -292,7 +292,7 @@ class PDFObject
                             )
                         ) {
                             // horizontal offset
-                            $text .= ' ';
+                            //$text .= ' ';
                         }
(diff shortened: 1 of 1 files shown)


smalot:
The "extra spaces" problem depends on each font used.
Indeed, each font uses a pixel matrix which indicates spaces between chars and words.
I never been able to identify the right space.
The current fix will patch only some cases and make fail those who are currently working

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -292,7 +292,7 @@ public function getText(Page $page = null)
                             )
                         ) {
                             // horizontal offset
-                            $text .= ' ';
+                            //$text .= ' ';
```
k00ni, line 295: What did you changed here? It looks the same. Furthermore please remove commented-out code.

k00ni:
Closed due to no activity.

Issue #42, opened 2014-07-08
Several Issues with greek characters.


IoannisLoukeris:
Hello.

I am using this library to extract info from several PDF files but although all English and French files are converted very easily and with no flaws, PDFs with Greek characters are hit and miss. Some are perfect. Some have some letters wrong some are completely garbled. 

You can see two very characteristic specimens at:

http://dev.life-read.gr/pdfs/5RVIZ15j14RjqGJ_3Cy0b.pdf (letters μ Μ Δ δ ω and Ω not rendered correctly through out the text)
http://dev.life-read.gr/pdfs/cRDA1T9U9a6m3lt_9dkV9.pdf (every letter after the fifth letter of the alphabet seems to be shifted by some places left. so letters α β γ δ ε are rendered correctly along with any accented letters but every thing else is garbled)

The generators for these are NITRO PDF and MS Word according to metadata extracted with pdfparser.

Thank you,

John

davispuh:
Try #257 PR it should be fixed there.

Connum:
I was able to get the PDF files for testing by removing the `dev.` from the provided URLs.
However, even going back as far as the 0.9.x releases from around October 2013, before this issue was opened, I wasn't able to reproduce any of the issues described. I don't speak or read Greek, but the letters were rendered fine and I didn't see any shifting or unusual spaces.

On the other hand, nothing improved with #257 either, as far as I can see...

So, unless @IoannisLoukeris has anything to add after 6+ years, I'd say it's safe to close this issue.

k00ni:
I will close this, but please reopen if the problem persists.

Issue #241, opened 2019-04-23
Encoding Error?


garyr87:
I'm having a problem when trying to read a Bank Statement PDF.  It appears as though its not picking up something.  Below is what is being returned from the getText() function.  Any assistance would be appreciated.  I've tried converting the text to different types however it still stays a black triangle with a ? mark in it.
 
@���������������	
2 4	Page	of @���������������	
12417220202000000062	
3 4	Page	of @���������������	
4 4	Page	of	
March 01, 2019	through	March 29, 2019	
Account Number:	
This Page Intentionally Left

Thanks much!

davispuh:
Try #257 PR maybe it works better.

k00ni:
@garyr87 , did you try it with #257? 

Can you upload the PDF here so we can use it to reproduce the error? It must be free to use, because we will use it in the test suite.

Connum:
With no sample PDF provided after a ping almost 3 months ago after almost 1.5 years, and a seemingly inactive account... I'd suggest closing this issue.

Issue #202, opened 2018-06-01
UTF-8 autodetection not working?


ThomasLandauer:
On some documents, I *need* to do `utf8_encode()` on the result of `getText()`, on some documents I *must not* do `utf8_encode()`, and on some it doesn't work either way ;-)

Looks like the encoding is somehow "autodetected"? How is this mechanism working and how can I adjust it?

davispuh:
Try #257 PR it should be fixed there.

Connum:
@ThomasLandauer I know it's been 2 years, but could you please try to provide a PDF with which I could reproduce this issue, or verify that it has been fixed in the meantime? I would very much like to find out whether #257 could have an impact here and would be worth being revived.

ThomasLandauer:
I would say the issue is still there. Here's an example: [test.pdf](https://github.com/smalot/pdfparser/files/5304153/test.pdf)

Expected outcome:
> „fööbär“

Actual outcome:
> �f��b�r�

I created the PDF with [dompdf](https://github.com/dompdf/dompdf) - here's the code (Symfony Controller):

```php
$dompdf = new Dompdf();
$dompdf->setPaper('A4', 'portrait');
$dompdf->loadHtml('<html><head><meta charset="utf-8"></head><body>„fööbär“</body></html>');
$dompdf->render();
$pdf = $dompdf->output();
file_put_contents('test.pdf', $pdf);

$parser = new \Smalot\PdfParser\Parser();
$pdf = $parser->parseFile('test.pdf');
$text = $pdf->getText();
return new Response("<html><head><meta charset='utf-8'></head><body>$text</body></html>");
```
Deleting both `<meta charset='utf-8'>` doesn't make a difference.
I have smalot/pdfparser v0.16.2

Connum:
I can confirm that this issue persists with the current master. It is however fixed by #257, which was closed due to inactivity. I'm working on reviving this with a new PR that addresses the failing and missing tests. Thanks for the sample, I'll be using it for the unit test!

k00ni:
@ThomasLandauer thank you for your feedback. Just to be sure, what character set / encoding is used in your Symfony application for the output?

ThomasLandauer:
> Thanks for the sample, I'll be using it for the unit test!

Yeah, fööbär forever! :-)

> Just to be sure, what character set / encoding is used in your Symfony application for the output?

[...]

Issue #152, opened 2017-05-24
Croatian letters not showing correctly


cicnavi:
Hello, thanks for creating this great package.
When converting from PDF to text, some letters commonly used in Croatian language are converted to some letters which are never used in Croatian language.
For example, letter "đ" becomes "Ď", letter "ž" becomes "ţ". The correct way would be that the letter "đ" stays "đ", and the same is with letter "ž". Is it possible to correct this?
Here is the PDF I tried to extract text from:
[Dumic_Mak.pdf](https://github.com/smalot/pdfparser/files/1024775/Dumic_Mak.pdf)
Thanks,
Marko

davispuh:
Try #257 PR it should be fixed there.

Connum:
I'm in the process of reviving #257 in a new PR. However, this issue is not solved by it.

Interestingly, when copy/pasting a word like `istraživanja` from the PDF in Acrobat DC into any text editor or input field, it becomes: `istraţivanja`. An using the search functionality of Acrobat DC or when viewing the PDF in a browser, one can search for `istraţivanja` and the word looking like `istraţivanja` will be highlighted.

Even more interestingly, when searching for `istraživanja` explicitly, there is only one instance of this word that is actually being highlighted (on page 28), despite the word appearing several times throughout the document. And this one instance is also correct when decoded by pdfparser:
![grafik](https://user-images.githubusercontent.com/13076806/94605168-0316a680-0299-11eb-820e-29403a1b0699.png)

**So it seems like the font that this PDF uses is actually substituting characters for each other. What looks like a `ž` is actually a `ţ` and therefore, it's correctly parsed as a `ţ` by pdfparser.** 

I don't think we can work around that in any way... The only thing that comes to my mind would be analyzing each glyph of the font and apply some kind of character recognition, but of course that's not feasible.

k00ni:
Thank you @Connum for checking that. I follow your argument, therefore issue is closed. If the problem persists, please reopen.

PR #257, opened 2019-09-17
Properly decode ANSI encodings

davispuh:
This PR fixes several bugs which prevented from correctly decoding ANSI encodings.
For example with this PR now can correctly get text of  PDF file which doesn't have `ToUnicode` but has `WinAnsiEncoding` with differences for Windows-1257 code page

Most likely these issues are fixed: #42 #44 #88 #151 #152 #202 #241


fbett:
Thx for this PR.

This also fixes a PHP 7.4 error:
````
chr() expects parameter 1 to be int, string given at smalot/pdfparser/src/Smalot/PdfParser/Font.php:494
````

**Comment by smalot** [comment]
That's a great work but some unit tests should be really appreciated

**Comment by smalot** [comment]
I just run "unit test" on your PR and some are broken.

Can you try on your side ?
```
./vendor/bin/atoum -d src/Smalot/PdfParser/Tests/ -ncc
```

davispuh:
I actually didn't even check them so most likely they need fixing.

k00ni:
Hey @davispuh, there are a few things before we can merge this:

* [ ] merge in current `master` branch
* [ ] run `php vendor/bin/php-cs-fixer fix` in the terminal (applies our new coding styles)
* [ ] fix failing tests

I hope we can finish this, this code seems important to other issues/PR's too.

davispuh:
I'm not using this library anymore and I don't really have time to complete this so you can just make new PR and fix whatever needs to be fixed.
That test fails because `$currentFont` seems to be `null` but I don't really see how that could happen.

k00ni:
Closed due to no progress. I checked some of the issues mentioned here, there was no indication that this PR helped them.

davispuh:
> there was no indication that this PR helped them.

That's weird, I didn't saw anyone actually testing this. It did fix a lot of issues for me. Anyway I guess not my problem anymore.

Issue #260, opened 2019-10-10
Wrong characters parsed with getText()


OneSoftIO:
I have a code:

`$parser = new Parser();`
`$pdf = $parser->parseFile(storage_path($file->full_path));`

It's a simple code to parse file content, after doing ->getText() output is:

> 1\t\fü\x00\x03\x00\x1C\x00\x18\x00\x0F\x00\x16\x007\t0\x00\b\x00 \t\fü\x00\x03\x00\x1C\x00\x18\x00\x0F\x00\x16\x007\t\fü\x00\x03\x00\x1C\x00\x18\x00\x0F\x00\x16\x007\t\n
> Costi di spedizione\t\fü\x00\x03\x00\x14\x00\x17\x00\x0F\x00\x1C\x009\t\fü\x00\x03\x00\x14\x00\x17\x00\x0F\x00\x1C\x009\t\fü\x00\x03\x00\x14\x00\x17\x00\x0F\x00\x1C\x009
> 

Bus some part of pdf is being read normally. 

Can you, please, help?

machekhin:
I`ve make a little fix here:
https://github.com/smalot/pdfparser/pull/269

You can change this file in your project while this pull request will be merged

Connum:
@OneSoftIO / @machekhin could you please provide a test file to reproduce this with? Because the fix for this breaks other PDF parsing processes and resulted in at least three new issues so far...

machekhin:
@Connum , sorry, I can`t show You this file even I find it, 'cause it`s a commercial document.
Anyway, it was a trouble with parse 'tab'-char \t. The previouse version had manual replace method and some special chars (like \f or \t) broke parcing.

Connum:
That's unfortunate... I'm still unsuccessful in trying to produce a PDF with a tab or form feed character that will crash or fail with pdfparser prior to the fix...

machekhin:
Maybe it was really \f. It is absent in the previous version:

https://github.com/smalot/pdfparser/pull/269/commits/4f4fd108270c32c0e81ea3dceaac2249162f27dc

Connum:
> 
> 
> Maybe it was really \f. It is absent in the previous version:
> 
> [4f4fd10](https://github.com/smalot/pdfparser/commit/4f4fd108270c32c0e81ea3dceaac2249162f27dc)

Yeah, probably, as `\t` was already present in the replacement code before.
[...]

PR #269, opened 2020-01-11
Fix \f crush

machekhin:
Escaped f (\f) in text crushes parcing process;
The better way is using base php function *stripcslashes*
FIx for https://github.com/smalot/pdfparser/issues/260


smalot:
Thank you

PR #290, opened 2020-05-18
Define lowest deps

j0k3r:
Define the lowest deps possible for the lib to work:
- tecnickcom/tcpdf has a security advisory before 6.2.22 (see https://packagist.org/packages/tecnickcom/tcpdf/advisories?version=2463879)
- atoum/atoum doesn't work with version 3.0.0 (got few issues) and the version 2.8.0 doesn't work with PHP 7

Also:
- add a new build with `--prefer-lowest` for composer, so the lib is tested with the lowest deps possible
- rise minimum version of PHP to 5.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,18 +9,18 @@ php:
   - 7.2
   - 7.3
   - 7.4
-  - hhvm
+  - nightly
 
 jobs:
   fast_finish: true
   allow_failures:
-    - php: hhvm
+    - php: nightly
   include:
     - php: 7.2
       env: COMPOSER_FLAGS="--prefer-lowest"
 
 install:
   - composer self-update
-  - composer update --prefer-source -o $COMPOSER_FLAGS
+  - composer update --prefer-dist --no-progress --no-suggest --optimize-autoloader $COMPOSER_FLAGS
 
 script: ./vendor/bin/atoum -d src/Smalot/PdfParser/Tests/ -ncc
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,24 +1,26 @@
...
-  - composer self-update || true
-  - composer update
-  - composer --prefer-source --dev install
+  include:
+    - php: 7.2
+      env: COMPOSER_FLAGS="--prefer-lowest"
+
+install:
+  - composer self-update
+  - composer update --prefer-source -o $COMPOSER_FLAGS
```
k00ni, line 24: Why using `--prefer-source` here? Wouldn't it be faster if we use `--prefer-dist` instead?

.travis.yml
```diff
@@ -1,24 +1,26 @@
...
-#  - 5.3
-#  - 5.4
-#  - 5.5
   - 5.6
   - 7.0
   - 7.1
   - 7.2
   - 7.3
   - 7.4
   - hhvm
```
k00ni, line 12: HHVM has no PHP support anymore and could therefore be ignored IMHO (source from 2018: https://hhvm.com/blog/2018/09/12/end-of-php-support-future-of-hack.html)

.travis.yml
j0k3r: Good idea, I just used what was previously defined.

.travis.yml
j0k3r: Good idea too, I'll drop it and add nightly instead

PR #282, opened 2020-03-31
Pdf data extractor

izabala:
### Outdated - Discussion moved to #297

---

The Pdf files save text data in a "Boxes" style, where any box has information like the coordinates (X,Y) where the box is and if direction where the text flow. Knowing this, you could know that a text that has the same Y value (or near) could be in the line, same with the X value. I am saying the same or near, because the (X,Y) value will change from one Pdf to another of the same kind of pdf.

When you used the getText() function from the Smalot/PdfParser, the text came disordered. And some times is really difficult to understand which is the info you are trying to get. But, using the X,Y coordinates of the text box, help to find the right info. That information is getted internally by the Smallot/PdfParser packages, but is not public so, you can not really used it.

I used to parse PDF file in Python using a module called pdfminer or pdfminer3k. In that modules, you can get the text of the PDF with the text boxes info, that helps you in the parse process.

I am changing smalot/pdfparser package to imitate that behavior, but I need to change a little bit (make a property from protected to public), and I am using internal status of the object that could be changed in the future, so I rather ask here for change so EveryOne can use it, and that changes in internal data and status, will not affect it.

The following method was added to help get the information is needed:
- extractRawData: it gets all the pdf in raw mode.
- extractDecodedRawData: decode the data extracted by extractRawData if it is encoded.
- getDataCommands: it filters to keep just de commands thats are related to the Text Matrix information (for getting coordinates) and the texts of the document.

These 3 method give the base for the following 2 methods that are needed to Parse the file by coordinates:
- getDataTm: gets the Text Matrix (Tm) for every text in the document. It returns an array, where every item has a Tm value (an array of 6 values) and the string of the text that has that Tm value.  The Text matriz (Tm) has 6 values:a b c d e f.  The e and the f values are the X, Y coordinates of the text.  The a, b, c and d values are involved in Scaling, Rotation and Skew of the text.  You use all this info to understand where is the Text that you need to parse, this is really useful for documents that has the same format.
-getTextXY: gets the text that are exactly or near the given coordinate.  It returns an array where every item has the Tm value of the text and the text itself.

---
(EDIT by @k00ni) **The following list contains points to discuss and open TODOs:**

* [ ] src/Smalot/PdfParser/Element.php: instead of making `value` public, please make it protected again and add a getter-function to retrieve its value.
* [x] `&$command`: see related comment - solved by explanation of @izabala
[...]

PR #296, opened 2020-05-28
For Pdf data extractor #282 but with the master merged and the TODO solved

izabala:
The Pdf files save text data in a "Boxes" style, where any box has information like the coordinates (X,Y) where the box is and if direction where the text flow. Knowing this, you could know that a text that has the same Y value (or near) could be in the line, same with the X value. I am saying the same or near, because the (X,Y) value will change from one Pdf to another of the same kind of pdf.

When you used the getText() function from the Smalot/PdfParser, the text came disordered. And some times is really difficult to understand which is the info you are trying to get. But, using the X,Y coordinates of the text box, help to find the right info. That information is getted internally by the Smallot/PdfParser packages, but is not public so, you can not really used it.

I used to parse PDF file in Python using a module called pdfminer or pdfminer3k. In that modules, you can get the text of the PDF with the text boxes info, that helps you in the parse process.

I am changing smalot/pdfparser package to imitate that behavior, but I need to change a little bit (make a property from protected to public), and I am using internal status of the object that could be changed in the future, so I rather ask here for change so EveryOne can use it, and that changes in internal data and status, will not affect it.

The following method was added to help get the information is needed:

extractRawData: it gets all the pdf in raw mode.
extractDecodedRawData: decode the data extracted by extractRawData if it is encoded.
getDataCommands: it filters to keep just de commands thats are related to the Text Matrix information (for getting coordinates) and the texts of the document.
These 3 method give the base for the following 2 methods that are needed to Parse the file by coordinates:

getDataTm: gets the Text Matrix (Tm) for every text in the document. It returns an array, where every item has a Tm value (an array of 6 values) and the string of the text that has that Tm value. The Text matriz (Tm) has 6 values:a b c d e f. The e and the f values are the X, Y coordinates of the text. The a, b, c and d values are involved in Scaling, Rotation and Skew of the text. You use all this info to understand where is the Text that you need to parse, this is really useful for documents that has the same format.
-getTextXY: gets the text that are exactly or near the given coordinate. It returns an array where every item has the Tm value of the text and the text itself.
(EDIT by @k00ni) The following list contains points to discuss and open TODOs:

 src/Smalot/PdfParser/Element.php: instead of making value public, please make it protected again and add a getter-function to retrieve its value.
 &$command: see related comment - solved by explanation of @izabala
 can you merge in the current master branch and run PHP-CS-Fixer to comply with our new coding styles? FYI: #292


izabala:
Closing and opening another one... for having it OK with my fork.

PR #299, opened 2020-06-01
Removed TCPDF library and ported used code - Outside tests welcome!

### LINKED ISSUE #291: How to handle outdated/abandoned lib "tecnickcom/tcpdf"?
We use `tecnickcom/tcpdf` which seems **abandoned** according to:

https://github.com/tecnickcom/tcpdf

> A new version of this library is under development at https://github.com/tecnickcom/tc-lib-pdf and as a consequence this version will not receive any additional development or support. This version should be considered obsolete, new projects should use the new version as soon it will become stable.

But its successor `tc-lib-pdf` is not ready yet:

> UPDATE: CURRENTLY ALL THE DEPENDENCY LIBRARIES ARE ALMOST COMPLETE BUT THE CORE LIBRARY STILL REQUIRES A SIGNIFICANT AMOUNT OF WORK TO BE COMPLETED.

... and is not actively developed. There are voices to help out (https://github.com/tecnickcom/tc-lib-pdf/issues/53), but nothing happened so far.

How much is this library relying on tecnickcom/tcpdf?

k00ni:
I removed `tecnickcom/tcpdf` as composer dependency and ported the code, which was used from TCPDF to:

* src/Smalot/PdfParser/RawData/FilterHelper.php
* src/Smalot/PdfParser/RawData/RawDataParser.php

**Reasons for this change:**
* TCPDF is not actively maintained
* with this change, PDFParser is fully standalone and controls all aspects of PDF parsing

closes #291

### Changes in comparison to TCPDF

* `TCPDF_PARSER` uses a config array named `$cfg`, which was not directly utilized by PDFParser. I kept most of its default values, except Option $cfg['die_for_errors']. It was removed, because in case of an error, one could set to either throw an exception or immediately stop the script. Latter is not recommended and throwing an exception allows better error-handling in general.
* use PHPDoc comments
* made `$this->pdfdata` and `$this->xref` local variables which allows more decoupled code and better control over their scope
* in `FilterHelper::decodeFilterFlateDecode` it uses `@gzuncompress`. I removed the `@` and added code so that non-catchable warnings by `gzuncompress` are now catchable.

### Changes to PDFParser

* Added parameter `$cfg` to `Parser.php` constructor which allows to set the following configuration values (Ported from TCPDF).
  * ignore_filter_decoding_errors (Default: `true`):  if `true` ignore filter decoding errors
  * ignore_missing_filter_decoders (Default: `true`): if `true` ignore missing filter decoding errors
* `TCPDF_PARSER` was used between `ob_start` and `ob_end_clean` which is problematic. Using `ob_*` functions may interfer with other libraries, such as PHPUnit. This library might not use PHPUnit, but other libraries could (like in a PHPUnit test). The removal of the `ob_*` functions results in more clean and overall compatible code.
* small refinements along the way

### TODOs

* [x] we may find some fixes in one of the pull requests at https://github.com/tecnickcom/TCPDF/pulls and can include them - *none available*
* [x] merge #300 into `master` branch first (atoum tests => PHPUnit)
  * [x] port new FilterHelper test case
  * [x] refine FilterHelper test to check exception thrown
* [x] Further tests from other people would be great! Can anyone can check with custom PDFs and post a comment, if errors occur? (@rubenvanerk tested PDFs successfully)

What do you think?


k00ni:
Main reason for the removal is the fact that TCPDF is **not maintained** anymore. There are merges from time to time, sure, but as stated in its README.md: `[...] as a consequence this version will not receive any additional development or support`. I assume these 12 pull requests will be there for another year.

Its successor is also not finished and not maintained. Furthermore, there are signs that it doesn't cover all areas its predecessor does, e.g. https://github.com/smalot/pdfparser/pull/92.

So we will run into a dead-end at some point. My suggestion would be to cut it off. This gives us (and our contributors) the option to take care of bugs ourselves, without the need to wait for TCPDF maintainer to integrate new fixes.

> Are you going to continually check for bug fixes from TCPDF on the code you've added? I won't.

I have it on my Github watch list. If there are any new release, i will get notified and can check. Also the code i ported is only a small chunk of library. Most of it isn't used in PDFParser anyway.

What do you think?

j0k3r:
Ok let's do it then.

rubenvanerk:
I have tested this with about 5 different pdfs and haven't run into any issues yet. I will continue to use these changes and let you know if I get any errors.

k00ni:
Thank you @rubenvanerk!

k00ni:
**If anyone catches problems with TCPDF, please post a comment here so that we can check if they work with the current version now**.

---

Just for documentation: the following file fails to parse with TCPDF but runs with our latest master. 

https://www.uni-konstanz.de/typo3temp/secure_downloads/57669/0/ec158dc3ad67761676c92b7f69f7864bac6240f8/Reise_Info.pdf

Using the outdated version with TCPDF library i will get the following error:

> 1) App\Tests\Integration\Service\PDFHelperTest::test
Exception: TCPDF_PARSER ERROR: **Invalid PDF data: missing %PDF header.**
/var/www/html/vendor/tecnickcom/tcpdf/tcpdf_parser.php:807
/var/www/html/vendor/tecnickcom/tcpdf/tcpdf_parser.php:108
/var/www/html/vendor/smalot/pdfparser/src/Smalot/PdfParser/Parser.php:91
/var/www/html/tests/Integration/Service/PDFHelperTest.php:26

PR #300, opened 2020-06-01
Port atoum tests to PHPUnit

k00ni:
This pull request replaces atoum tests with PHPUnit tests.

We have code coverage now (via scrutinizer):

### Reasons for this pull request

* use PHPUnit, because it more known and better integrated as atoum
* enable code coverage analysis
* aligned and fixed some tests (e.g. removed code which was commented out)

### Changes

* test related files are located in `test/`
* tests itself are located in `tests/`
* replaced many atoum tests with PHPUnit ones
* put tests and test related files into a separate namespace (using PSR-4)
* remove usages of full class paths, used `use Class` instead
* use functions to get class-instances during tests, like for `Parser`. That makes tests more decoupled from the actual class constructor and in case it changes, only 1 place has to be adapted.

### TODOs:

* [x] port remaining tests
* [x] integrate in Travis
* [x] generate code coverage
  * [x] ~~prepare for coveralls?~~ - use https://scrutinizer-ci.com/docs/build/code_coverage instead
* [x] fix issues raised by @j0k3r


composer.json
```diff
@@ -21,14 +21,20 @@
...
     },
     "autoload": {
         "psr-0": {
             "Smalot\\PdfParser\\": "src/"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "Test\\Smalot\\PdfParser\\": "test/",
+            "Tests\\Smalot\\PdfParser\\": "tests/"
```
j0k3r, line 35: Why do you have `test` & `tests`?
Keep only `tests`, it's enough.

.travis.yml
```diff
@@ -23,4 +23,4 @@ install:
   - composer self-update
   - composer update --prefer-dist --no-progress --no-suggest --optimize-autoloader $COMPOSER_FLAGS
 
-script: ./vendor/bin/atoum -d src/Smalot/PdfParser/Tests/ -ncc
+script: vendor/bin/phpunit --coverage-clover coverage/clover.xml
```
j0k3r, line 26: I think we shouldn't generate coverage for each build.
Only one is enough and make the whole test suite faster.

You can re-use the same solution I used to define `$COMPOSER_FLAGS`

j0k3r:
I'm wondering why does Scrutinizer doesn't show up along with Travis build 🤔

phpunit.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.1/phpunit.xsd"
+         backupGlobals="false"
+         backupStaticAttributes="false"
+         bootstrap="test/bootstrap.php"
```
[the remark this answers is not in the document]
k00ni: The purpose of the `bootstrap.php` file is to be a place for test preparations. There aren't any, but maybe in the future. But OK, will change it.

composer.json
k00ni: `test` contains test related classes, like `TestCase` or maybe traits etc. later on. Its the place for other helper classes too.

`tests` only contains the tests itself.

I saw this separation on a few projects and find it easier to maintain this way.

.travis.yml
k00ni: You mean a separate build for coverage, like yours with `$COMPOSER_FLAGS`?

j0k3r: Yep

k00ni:
> I'm wondering why does Scrutinizer doesn't show up along with Travis build

I don't know. But in case you are interested in the latest results: https://scrutinizer-ci.com/g/smalot/pdfparser/?branch=feature/switch-to-phpunit

src/Smalot/PdfParser/Page.php
```diff
@@ -382,18 +374,16 @@ public function extractDecodedRawData($extractedRawData = null)
...
-    /*
+    /**
      * Gets just the Text commands that are involved in text positions and
      * Text Matrix (Tm)
      *
      * It extract just the PDF commands that are involved with text positions, and
      * the Text Matrix (Tm). These are: BT, ET, TL, Td, TD, Tm, T*, Tj, ', ", and TJ
      *
      * @param array $extractedDecodedRawData The data extracted by extractDecodeRawData
-                           if it is null, the method extractDecodeRawData is called.
```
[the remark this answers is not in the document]
k00ni: Ok.

.travis.yml
```diff
@@ -18,9 +18,12 @@ jobs:
   include:
     - php: 7.2
       env: COMPOSER_FLAGS="--prefer-lowest"
+  include:
```
j0k3r, line 21: And it should be ok

```suggestion
```

k00ni: Fixed.

phpunit.xml
```diff
@@ -0,0 +1,39 @@
...
+
+    <testsuites>
+        <testsuite name="all">
+            <directory>tests</directory>
+        </testsuite>
+    </testsuites>
+    <logging>
+        <log type="coverage-html" target="./coverage/html.html" showUncoveredFiles="false" />
+        <log type="coverage-clover" target="./coverage/clover.xml"/>
+    </logging>
```
j0k3r, line 33: You should comment these lines too.
Otherwise, the coverage will always be calculate no matter the `COVERAGE_FLAGS`

k00ni: Ok.

k00ni:
I am still not sure when it runs code coverage. Even if you have `logging` in the `phpunit.xml` it wont create coverage. You always had to use the `--coverage-*` option when calling PHPUnit.

.travis.yml
k00ni: Done.

composer.json
k00ni: Any objections here?

j0k3r:
With `logging` in the `phpunit.xml`, the test suite is slower and it generates the coverage.

Without `logging`:
image

With `logging`:
image

You can see it generates the coverage.

k00ni:
Good catch! I just saw, that i used a Docker container without xdebug installed.

OK anything left or can i merge it?

PR #301, opened 2020-06-02
Ignore some files for the end user

j0k3r:
As of now, the whole repository is download to people who are using the lib.

Some files don't need to be available for the end user in a production environment:
- all _dotfiles_ which are usually only for development
- tests files which are only for test
- samples PDF files (saving 1,2Mo!)

Also adding [`.editorconfig`](https://editorconfig.org/) so everyone should have the same behavior in its editor.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+; top-most EditorConfig file
+root = true
+
+; Unix-style newlines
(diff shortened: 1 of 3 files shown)


j0k3r:
@k00ni of course :)

PR #302, opened 2020-06-04
Use SVG icons

j0k3r:
For a better rendering.

Also:
- remove the `composer.lock` badge (which I found useless) and use the License badge instead.
- add scrutinizer badges


k00ni:
Good addition. :+1:

PR #304, opened 2020-06-04
Add PHPStan (+ extend php-cs-fixer to complain)

j0k3r:
It'll fix bug :)

Also:
- fix namespace for some tests
- add custom build for linter stuff (php-cs-fixer & phpstan)
- convert some else into _return early_ to avoid useless block


j0k3r:
I don't know why checks aren't updated with Travis & Scrutinizer 🤷‍♂️

.travis.yml
```diff
@@ -18,11 +16,16 @@ jobs:
...
     - php: 7.4
       env: COVERAGE_FLAGS="--coverage-clover coverage/clover.xml"
 
 install:
-  - composer self-update
   - composer update --prefer-dist --no-progress --no-suggest --optimize-autoloader $COMPOSER_FLAGS
 
-script: vendor/bin/phpunit $COVERAGE_FLAGS
+script:
+  - if [ "$LINTER_RUN" = "run" ]; then php vendor/bin/php-cs-fixer fix --verbose --dry-run ; fi;
```
k00ni, line 28: I assume it will raise errors if it founds coding-style violations?

k00ni:
Travis looks good though: https://travis-ci.org/github/smalot/pdfparser/builds/694681387

I ran into this problem on another repository and don't know why its sometimes not included in a PR.

.travis.yml
```diff
@@ -18,11 +16,16 @@ jobs:
   include:
     - php: 7.2
       env: COMPOSER_FLAGS="--prefer-lowest"
+    - php: 7.3
```
k00ni, line 19: Just a thought: instead of splitting PHP 7.3 and 7.4 and using it for different purposes, why not say that the latest stable PHP version is used with `COMPOSER_FLAGS` and 'LINTER_RUN`? That way we only have to maintain one entry.

.travis.yml
j0k3r: Yup

PR #306, opened 2020-06-08
Allow PHP 8

j0k3r:
And enable tests for it using PHPUnit Bridge from Symfony (I wasn't able to find a proper restriction in composer to allow PHPUnit to work from PHP 5.6 to PHP 8).

Also:
- use custom config for PHPStan (to specify phpunit path (because of simple-phpunit))
- force PHPUnit version in Scrutinizer as long as the PHP version
- move `COMPOSER_FLAGS` outside linter stuff to avoid side behavior (because of lower deps version)
- remove php-cs-fixer for non linter build because it enforce PHP < 8
- force PHPUnit version to avoid error with lower PHP version

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,11 +1,14 @@
-# For more information about code coverage
-# See: https://scrutinizer-ci.com/docs/build/code_coverage
 build:
-  nodes:
-    coverage:
-      tests:
-        override:
-          - command: vendor/bin/phpunit --coverage-clover coverage/clover.xml
-            coverage:
-              file: coverage/clover.xml
-              format: clover
+    nodes:
+        analysis:
+            tests:
+                override:
+                    - php-scrutinizer-run
+                    -
+                        command: SYMFONY_PHPUNIT_VERSION=7.5 php vendor/bin/simple-phpunit --coverage-clover coverage/clover.xml
+                        coverage:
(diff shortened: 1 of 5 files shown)


j0k3r:
Again, checks weren't triggered properly:
- travis: https://travis-ci.org/github/smalot/pdfparser/builds/695926021
- scrutinizer: https://scrutinizer-ci.com/g/smalot/pdfparser/inspections/1ead038c-d749-44af-848f-c4623676ccfa

.scrutinizer.yml
```diff
@@ -1,11 +1,14 @@
...
-            coverage:
-              file: coverage/clover.xml
-              format: clover
+    nodes:
+        analysis:
+            tests:
+                override:
+                    - php-scrutinizer-run
+                    -
+                        command: SYMFONY_PHPUNIT_VERSION=7.5 php vendor/bin/simple-phpunit --coverage-clover coverage/clover.xml
```
k00ni, line 8: Just curious: was that line break intended?

j0k3r: Just to be more readable :)

j0k3r:
> So you switched to PHPUnit-bridge to allow PHPUnit run for PHP 5.6 until 8?

Yeah kind of. PHPUnit bridge determine which version of PHPUnit it should install depending on the PHP version.

PR #307, opened 2020-06-15
Fix problems in decodeFilterASCII85Decode

randeraad:
Only check for the start sequence at the start of the data.
This fixes a problem where the encoded data ends with a < (3Ch), with the EOD characters this becomes <~>
In this case the first characters are removed when they shouldn't.

Fix problem in preg_match, \x74 (t) should be \x7A (z)

### DIFF
--- a/src/Smalot/PdfParser/RawData/FilterHelper.php
+++ b/src/Smalot/PdfParser/RawData/FilterHelper.php
@@ -146,8 +146,8 @@ class FilterHelper
         // all white-space characters shall be ignored
         $data = preg_replace('/[\s]/', '', $data);
         // remove start sequence 2-character sequence <~ (3Ch)(7Eh)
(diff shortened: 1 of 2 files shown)


randeraad:
I tried to make this two pull requests but they got merged to one, can't find the option to remove this pull request.

k00ni:
Hi @randeraad, first of all thank you for your PR.

There seems to be a problem with the branch your code is coming from. Github says to me ` unknown repository`. When i click on it i come to this repository and not yours. So i assume you deleted something.

Just to be sure, can you just create new PR?

Furthermore, can you add a test which shows that there was a problem before?

randeraad:
I tried to remove this pull request by deleting my fork.
I'll create a new PR when I've added the tests.

k00ni:
Thank you. Looking forward to it.

PR #310, opened 2020-06-19
Fix array access on integer for php7.4

### LINKED ISSUE #267: In Parser.php line 154: Notice: Trying to access array offset on value of type int
Hi
I have an error :
`In Parser.php line 154: Notice: Trying to access array offset on value of type int 
`

PaulBehrendtVentoro:
fixes "Notice: Trying to access array offset on value of type int" for certain PDFs


k00ni:
Looks good to me. Thank you @PaulBehrendtVentoro!

*I let this open for a day or two in case someone wants to comment.*

PaulBehrendtVentoro:
btw. this does not fix the "Missing Catalog" thing as your test indicates.
So it shouldn't close #267 .

k00ni:
> btw. this does not fix the "Missing Catalog" thing as your test indicates. So it shouldn't close #267.

You are right. But Github thinks that and will close issue #267 on merge. Maybe its because in your initial post you wrote "fixes #267"?

PaulBehrendtVentoro:
yeah i didnt know that. looks alright now

k00ni:
**Notice to myself:** after merging, remove branch `issue/267`.

PR #312, opened 2020-06-23
fix missing catalog bug (+ some code refactoring)

### LINKED ISSUE #267: In Parser.php line 154: Notice: Trying to access array offset on value of type int

PaulBehrendtVentoro:
fixes #267

### DIFF
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -465,7 +465,9 @@ class RawDataParser
      */
     protected function getIndirectObject($pdfData, $xref, $objRef, $offset = 0, $decoding = true)
     {
-        // --- build indirect object header
+        /*
+         * build indirect object header
+         */
         // $objHeader = "[object number] [generation number] obj"
         $objRefArr = explode('_', $objRef);
         if (2 !== \count($objRefArr)) {
@@ -473,8 +475,10 @@ class RawDataParser
         }
         $objHeader = $objRefArr[0].' '.$objRefArr[1].' obj';
 
-        // --- check if we are in position
-        // ignore whitespace characters at offset
+        /*
+         * check if we are in position
+         */
+        // ignore whitespace characters at offset (NUL, HT, LF, FF, CR, SP)
         $offset += strspn($pdfData, "\0\t\n\f\r ", $offset);
         // ignore leading zeros for object number
         $offset += strspn($pdfData, '0', $offset);
@@ -483,7 +487,9 @@ class RawDataParser
             return ['null', 'null', $offset];
         }
 
-        // --- get content
+        /*
+         * get content
+         */
         // starting position of object content
         $offset += \strlen($objHeader);
         $objContentArr = [];
@@ -503,7 +509,9 @@ class RawDataParser
         // remove closing delimiter
         array_pop($objContentArr);
 
-        // --- return raw object content
+        /*
+         * return raw object content
+         */
         return $objContentArr;
     }
 
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -471,7 +471,7 @@ class RawDataParser
         }
         $objref = $obj[0].' '.$obj[1].' obj';
         // ignore leading zeros
-        $offset += strspn($pdfData, '0', $offset);
+        $offset += strspn($pdfData, "0\n", $offset);
         if (strpos($pdfData, $objref, $offset) != $offset) {
             // an indirect reference to an undefined object shall be considered a reference to the null object
             return ['null', 'null', $offset];
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -470,8 +470,8 @@ class RawDataParser
             throw new Exception('Invalid object reference for $obj.');
         }
         $objref = $obj[0].' '.$obj[1].' obj';
-        // ignore leading zeros
-        $offset += strspn($pdfData, "0\n", $offset);
+        // ignore newlines & leading zeros
+        $offset += strspn($pdfData, "0\r\n", $offset);
         if (strpos($pdfData, $objref, $offset) != $offset) {
             // an indirect reference to an undefined object shall be considered a reference to the null object
             return ['null', 'null', $offset];
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -455,7 +455,7 @@ class RawDataParser
      *
      * @param string $pdfData  PDF data
      * @param array  $xref
-     * @param string $obj_ref  Object number and generation number separated by underscore character
+     * @param string $objRef   Object number and generation number separated by underscore character
      * @param int    $offset   Object offset
      * @param bool   $decoding If true decode streams
      *
@@ -463,51 +463,59 @@ class RawDataParser
      *
      * @throws Exception if invalid object reference found
      */
-    protected function getIndirectObject($pdfData, $xref, $obj_ref, $offset = 0, $decoding = true)
+    protected function getIndirectObject($pdfData, $xref, $objRef, $offset = 0, $decoding = true)
     {
-        $obj = explode('_', $obj_ref);
-        if (2 != \count($obj)) {
+        // --- build indirect object header
+        // $objHeader = "[object number] [generation number] obj"
+        $objRefArr = explode('_', $objRef);
+        if (2 !== \count($objRefArr)) {
             throw new Exception('Invalid object reference for $obj.');
         }
-        $objref = $obj[0].' '.$obj[1].' obj';
-        // ignore newlines & leading zeros
-        $offset += strspn($pdfData, "0\r\n", $offset);
-        if (strpos($pdfData, $objref, $offset) != $offset) {
+        $objHeader = $objRefArr[0].' '.$objRefArr[1].' obj';
+
+        // --- check if we are in position
+        // ignore whitespace characters at offset
+        $offset += strspn($pdfData, "\0\t\n\f\r ", $offset);
+        // ignore leading zeros for object number
+        $offset += strspn($pdfData, '0', $offset);
+        if (substr($pdfData, $offset, \strlen($objHeader)) !== $objHeader) {
             // an indirect reference to an undefined object shall be considered a reference to the null object
             return ['null', 'null', $offset];
         }
+
+        // --- get content
         // starting position of object content
-        $offset += \strlen($objref);
-        // get array of object content
-        $objdata = [];
+        $offset += \strlen($objHeader);
+        $objContentArr = [];
         $i = 0; // object main index
         do {
-            $oldoffset = $offset;
+            $oldOffset = $offset;
             // get element
             $element = $this->getRawObject($pdfData, $offset);
             $offset = $element[2];
             // decode stream using stream's dictionary information
-            if ($decoding and ('stream' == $element[0]) and (isset($objdata[($i - 1)][0])) and ('<<' == $objdata[($i - 1)][0])) {
-                $element[3] = $this->decodeStream($pdfData, $xref, $objdata[($i - 1)][1], $element[1]);
+            if ($decoding && ('stream' === $element[0]) && (isset($objContentArr[($i - 1)][0])) && ('<<' === $objContentArr[($i - 1)][0])) {
+                $element[3] = $this->decodeStream($pdfData, $xref, $objContentArr[($i - 1)][1], $element[1]);
             }
-            $objdata[$i] = $element;
+            $objContentArr[$i] = $element;
             ++$i;
-        } while (('endobj' != $element[0]) and ($offset != $oldoffset));
-
+        } while (('endobj' !== $element[0]) && ($offset !== $oldOffset));
         // remove closing delimiter
-        array_pop($objdata);
+        array_pop($objContentArr);
 
-        // return raw object content
-        return $objdata;
+        // --- return raw object content
+        return $objContentArr;
     }
 
     /**
-     * Get the content of object, resolving indect object reference if necessary.
+     * Get the content of object, resolving indirect object reference if necessary.
      *
      * @param string $pdfData PDF data
      * @param array  $obj     Object value
      *
      * @return array containing object data
+     *
+     * @throws Exception
      */
     protected function getObjectVal($pdfData, $xref, $obj)
     {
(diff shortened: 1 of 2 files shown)


j0k3r:
@k00ni We really have an issue with Travis not triggered from time to time 😔

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -471,7 +471,7 @@ protected function getIndirectObject($pdfData, $xref, $obj_ref, $offset = 0, $de
         }
         $objref = $obj[0].' '.$obj[1].' obj';
         // ignore leading zeros
-        $offset += strspn($pdfData, '0', $offset);
+        $offset += strspn($pdfData, "0\n", $offset);
```
k00ni, line 474: Could you please describe how you come up with this fix?

tests/Integration/ParserTest.php
```diff
@@ -86,12 +86,10 @@ public function testParseFile()
      *      Missing catalog.
      *
      * @see https://github.com/smalot/pdfparser/issues/267
+     * @doesNotPerformAssertions
```
k00ni, line 89: The purpose of this test (`testIssue267`) was to keep track of a faulty functionality. After your fix this test should be altered and actually do something with the extracted data. I am open for suggestions.

src/Smalot/PdfParser/RawData/RawDataParser.php
PaulBehrendtVentoro: Yes, this is more like a dirty fix. The offset given by `RawDataParser::getXrefData` is off by a newline for the given PDF which does not fit assumptions in `RawDataParser::getIndirectObject`. But i am new to this package and not so familiar with PDF-standards, so i was afraid of changing regex in `getXrefData`. To me all seamed to work fine when adding newlines to the trimming in `getIndirectObject`. But i will review `RawDataParser::getXrefData` again, so maybe wait a bit if you intend to merge.

tests/Integration/ParserTest.php
PaulBehrendtVentoro: Ok i will add some test cases.

k00ni: Looking forward to it.

src/Smalot/PdfParser/RawData/RawDataParser.php
PaulBehrendtVentoro: Ok I did not find anything on cross-ref offsets and leading newlines. So i guess it is fine to throw it away. Assuming "Microsoft: Print To PDF" has correctly implemented PDF-1.7-standard & given PDF was not corrupted afterwards.

tests/Integration/ParserTest.php
```diff
@@ -66,7 +67,7 @@ public function testParseFile()
                     } catch (Exception $e) {
                         if (
                             'Secured pdf file are currently not supported.' !== $e->getMessage()
-                            && 0 != strpos($e->getMessage(), 'TCPDF_PARSER')
+                            && 0 !== strpos($e->getMessage(), 'TCPDF_PARSER')
```
PaulBehrendtVentoro, line 70: please double check this. looked like a bug to me, because the "Missing catalog"-Exception was not catched by this test.

k00ni: The `TCPPDF_PARSER` reference is obsolete there and could be removed. But it should be done separately as long as it doesn't interfere with your fix.

tests/Integration/ParserTest.php
```diff
@@ -77,26 +78,17 @@ public function testParseFile()
...
-        $this->expectExceptionMessage('Missing catalog.');
-
         $filename = $this->rootDir.'/samples/bugs/Issue267_array_access_on_int.pdf';
 
         $document = $this->fixture->parseFile($filename);
 
-        // triggers the exception
-        $document->getPages();
+        self::assertEquals(Image::class, \get_class($document->getObjectById('128_0')));
+        self::assertStringContainsString('4 von 4', $document->getText());
```
j0k3r, line 92: We usually use `$this->` instead of `self::`

```suggestion
        $this->assertEquals(Image::class, \get_class($document->getObjectById('128_0')));
        $this->assertStringContainsString('4 von 4', $document->getText());
```

src/Smalot/PdfParser/RawData/RawDataParser.php
PaulBehrendtVentoro: I think in the tc-lib-pdf-parser project this was solved by the `$offset++` there
See https://github.com/tecnickcom/tc-lib-pdf-parser/pull/6/files

tests/Integration/ParserTest.php
PaulBehrendtVentoro: So you mean i should remove it in another PR?

src/Smalot/PdfParser/RawData/RawDataParser.php
PaulBehrendtVentoro: Ok i tried this solution, works as well. Not sure which one is better.

tests/Integration/ParserTest.php
k00ni: As long as it doesn't interfere with your fix, leave it be. We can take care of it later on.

src/Smalot/PdfParser/RawData/RawDataParser.php
k00ni: ### About TCPDF
There are so many repositories from tecnickcom for PDF parser: 
* **https://github.com/tecnickcom/TCPDF**
* https://github.com/tecnickcom/tc-lib-pdf-parser
* https://github.com/tecnickcom/tc-lib-pdf

Previously used in this library was **TCPDF**. But i removed it and moved some of its code to us (Ref: #299).

---

### About this fix

First of all, its great that you support us here, so please don't get me wrong. But i am not sure, if you know **why** your changes fix the issue. You wrote it yourself before: `Yes, this is more like a dirty fix.`. 

You added `\r\n` to include line breaks, which is Windows specific. Others like Linux use `\n` for instance. That leads to the question if your fix also covers PDFs which were generated on a Mac for instance?

To be honest, i still lack detailed knowledge about PDF parsing. So I can't decide if your fix works or not. Furthermore I just started to get used to the code of this library too. **That said I really would like to hear a third opinion on that change.**

This library is used by >600 projects (said Github) and they relying on us to keep it stable. If you can point me to some specification (or something like that) which supports the change you made, that would help.

PaulBehrendtVentoro: Hey @k00ni , i totally understand that. I am not expecting this to be merged without further review.

To my understanding the cross reference table is parsed just fine in TCPDF. 

_The byte offset in the decoded stream shall be a 10-digit number, padded with leading zeros if necessary,
giving the number of bytes from the beginning of the file to the beginning of the object._ 
(https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf)

but i did not find good resources if an object can be padded at that offset? which microsoft seems to be doing?
so i think a solution in `RawDataParser::getIndirectObject` might be the right choice. ("this is a dirty fix" was my understanding yesterday)

i referenced the tcpdf commit because it seems to address a similar issue. this was just fyi. and it does basically the same.

I dont get your point on the line breaks? i am removing both (CR & LF) which should work for windows and unix, am i wrong?

PaulBehrendtVentoro: fyi: Setasign/FPDI seems to do this for \x20\x0A\x0C\x0D\x09\x00

see PdfParser/Tokenizer.php

PaulBehrendtVentoro: or https://github.com/mozilla/pdf.js/blob/v2.5.207/src/core/obj.js#L1504

PaulBehrendtVentoro: as i understand https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDFReference13.pdf:

_PDF treats any sequence of consecutive white-space characters as if there were just one_

_For the most part, EOL markers are treated the same as any other white-space characters._

and **3.2.9 Indirect Objects** defines no special meaning for EOL.

markus-lax:
Hi guys,

I guess I found a solution for the PDF-1.7 issue. For me it works perfect, so I hope it's useful and can be implemented. 

image

Regards,
Markus Lax

k00ni:
Hi, is there a way to post the whole "fix", because on the picture some text parts are cut off.

PaulBehrendtVentoro:
and it should be `if ( substr($pdfData, $offset, 1) === "\n" ) {`, which is basically the same i did, but without respecting carriage returns.

PaulBehrendtVentoro:
btw. @k00ni: what ist the status on https://github.com/smalot/pdfparser/pull/312/files/e8b815790a9b7d7e4f502693466bce7d02d7fc68#r444900023

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -470,8 +470,8 @@ protected function getIndirectObject($pdfData, $xref, $obj_ref, $offset = 0, $de
             throw new Exception('Invalid object reference for $obj.');
         }
         $objref = $obj[0].' '.$obj[1].' obj';
-        // ignore leading zeros
-        $offset += strspn($pdfData, '0', $offset);
+        // ignore newlines & leading zeros
+        $offset += strspn($pdfData, "0\r\n", $offset);
```
k00ni, line 474: In the link you posted they also check whitespaces. But here you only ignore newlines and leading zeros.

https://github.com/mozilla/pdf.js/blob/v2.5.207/src/core/obj.js#L1504

k00ni: And can you tell me, if there are any changes in future PDF versions related to this?

PaulBehrendtVentoro: added ignoring of whitespaces (see ISO 32000-1 > 7.2.2 Character Set > Table 1) which is revised version for PDF-1.7

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -455,59 +455,69 @@ protected function decodeXrefStream($pdfData, $startxref, $xref = [])
...
+     * @param string $objRef   Object number and generation number separated by underscore character
      * @param int    $offset   Object offset
      * @param bool   $decoding If true decode streams
      *
      * @return array containing object data
      *
      * @throws Exception if invalid object reference found
      */
-    protected function getIndirectObject($pdfData, $xref, $obj_ref, $offset = 0, $decoding = true)
+    protected function getIndirectObject($pdfData, $xref, $objRef, $offset = 0, $decoding = true)
```
PaulBehrendtVentoro, line 466: Pls double check my refactorings

k00ni: Looks good to me. I didn't found any `$obj_ref` anymore.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -455,59 +455,67 @@ protected function decodeXrefStream($pdfData, $startxref, $xref = [])
...
     {
-        $obj = explode('_', $obj_ref);
-        if (2 != \count($obj)) {
+        // --- build indirect object header
+        // $objHeader = "[object number] [generation number] obj"
```
k00ni, line 469: Please remove code which is commented out.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -455,59 +455,67 @@ protected function decodeXrefStream($pdfData, $startxref, $xref = [])
...
      * @return array containing object data
      *
      * @throws Exception if invalid object reference found
      */
-    protected function getIndirectObject($pdfData, $xref, $obj_ref, $offset = 0, $decoding = true)
+    protected function getIndirectObject($pdfData, $xref, $objRef, $offset = 0, $decoding = true)
     {
-        $obj = explode('_', $obj_ref);
-        if (2 != \count($obj)) {
+        // --- build indirect object header
```
k00ni, line 468: Just a small hint: I prefer separators like

```
/*
 * build indirect object header
 */
```
instead. The make it more clear that a new code section starts. 

But I wouldn't mind if you keep yours.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -455,59 +455,67 @@ protected function decodeXrefStream($pdfData, $startxref, $xref = [])
...
+        $objRefArr = explode('_', $objRef);
+        if (2 !== \count($objRefArr)) {
             throw new Exception('Invalid object reference for $obj.');
         }
-        $objref = $obj[0].' '.$obj[1].' obj';
-        // ignore leading zeros
+        $objHeader = $objRefArr[0].' '.$objRefArr[1].' obj';
+
+        // --- check if we are in position
+        // ignore whitespace characters at offset
```
k00ni, line 477: Please add all things you ignore there to the comment. As far as I can tell, you at least ignoring:

* tabs
* new lines
* whitespaces

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -455,59 +455,67 @@ protected function decodeXrefStream($pdfData, $startxref, $xref = [])
...
-            $oldoffset = $offset;
+            $oldOffset = $offset;
             // get element
             $element = $this->getRawObject($pdfData, $offset);
             $offset = $element[2];
             // decode stream using stream's dictionary information
-            if ($decoding and ('stream' == $element[0]) and (isset($objdata[($i - 1)][0])) and ('<<' == $objdata[($i - 1)][0])) {
-                $element[3] = $this->decodeStream($pdfData, $xref, $objdata[($i - 1)][1], $element[1]);
+            if ($decoding && ('stream' === $element[0]) && (isset($objContentArr[($i - 1)][0])) && ('<<' === $objContentArr[($i - 1)][0])) {
+                $element[3] = $this->decodeStream($pdfData, $xref, $objContentArr[($i - 1)][1], $element[1]);
```
k00ni, line 498: Can you be sure that the `1` key is always set in `$objContentArr[($i - 1)][1]`? You only check for the `0` in the if-clause before.

src/Smalot/PdfParser/RawData/RawDataParser.php
PaulBehrendtVentoro: this is more like pseudo code to know what is the result of the next few lines

src/Smalot/PdfParser/RawData/RawDataParser.php
PaulBehrendtVentoro: i just replaced stuff that safely can be changed to get stuff all little bit more readable to me. I don't wanted to introduce breaking changes.
Adding another check could change behavior.

k00ni:
Thank you.

Issue #201, opened 2018-05-25
getText() returns text breaking up words with spaces


kotsis:
I use as example the file http://www.pdf995.com/samples/pdf.pdf
getText() returns this:

"The p df9 95  s u it e  o f p ro ducts  -  P df9 95,  P dfE dit 9 95 , a n d S ig na tu re 99 5 -  is  a  c o m ple te  s o lu tio n fo r y o ur d ocu m ent p ublis h in g n ee ds. It 
p ro vid es e ase  o f u se , fle xib ilit y  in  fo rm at,  a nd  in d ustr y -s ta nd ard  s e cu rit y - a nd a ll  a t n o  c o st to  y o u. 
P df9 95	 m ake s it  e a sy a nd  a ffo rd able  to  c re ate  p ro fe ssio nal- q u alit y  d o cu m en ts  in  th e p o pula r P D F file  fo rm at. I ts  e asy-to -u se  in te rfa ce  
h elp s y o u to  c re a te  P D F file s b y s im ply  s e le ctin g t h e " p rin t"  c o m mand f r o m  a ny a pp lic a tio n,  c re atin g d ocu m ents  w hic h  c a n  b e v ie w ed 
o n a ny c o m pute r w it h  a  P D F v ie w er. P df9 95 s u pp orts  n etw ork  file  s a vin g, f a st u se r s w it c h in g  o n X P , C it r ix /T erm in al  S erv e r, c u sto m  
p age s iz e s a nd la rg e fo rm at  p rin tin g . P df9 95 is  a  p rin te r d riv e r th at w ork s w it h  a ny P osts crip t to  P D F c o nve rte r. T he p df9 95  p rin te r 
d riv e r a nd a  fr e e C onve rte r a re  a va ila b le  fo r e a sy d ow nlo ad . 	
P dfE dit 9 95	 o ffe rs  a  w ealt h  o f a ddit io na l  fu n ctio na lit y , s u ch  a s: c o m bin in g d o cu m en ts  in to  a  s in g le  P D F;  a uto m atic  lin k in se rtio n; 
h ie ra rc h ic a l  b ookm ark  in se rtio n ; P D F c o nve rs io n to  H TM L o r D O C ( te xt o nly ); in te gra tio n w it h  W ord  to olb ar w it h  a u to m atic  ta ble  o f 
c o nte nts  a nd lin k g enera tio n ; a u to atta ch  to  e m ail;  s ta tio ne ry  a n d s ta m pin g.  	
[...]

PR #318, opened 2020-07-08
respect space width when using "Move text position" stream operator

PaulBehrendtVentoro:
fixes at least #201 

maybe #72 too, but i cannot download the pdf anymore

closes https://github.com/smalot/pdfparser/pull/245
maybe closes https://github.com/smalot/pdfparser/pull/256


src/Smalot/PdfParser/PDFObject.php
```diff
@@ -274,14 +275,43 @@ public function getText(Page $page = null)
                         $args = preg_split('/\s/s', $command[self::COMMAND]);
                         $y = array_pop($args);
                         $x = array_pop($args);
+                        // get max char size and use as minimum for horizontal offset
+                        $fontMaxWidth = 0;
```
PaulBehrendtVentoro, line 279: should be done outside of foreach loop

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -274,14 +275,43 @@ public function getText(Page $page = null)
...
+                            if ('cidfonttype2' === strtolower($fontDictonary['Type'])) {
+                                if (isset($fontDictonary['DW']) && $fontDictonary['DW']) {
+                                    $fontMaxWidth = ((float) $fontDictonary['DW'] / 1000) * $current_font_size;
+                                } else {
+                                    // default
+                                    $fontMaxWidth = $current_font_size;
+                                }
+                            }
+                        }
+
```
PaulBehrendtVentoro, line 306: should be done outside of foreach loop

tests/Integration/ParserTest.php
```diff
@@ -99,4 +101,45 @@ public function testIssue267()
...
+        $i = 0;
+        foreach ($document->getObjects() as $object) {
+            if (Image::class === \get_class($object)) {
+                ++$i;
+            }
+        }
+        $this->assertEquals(1, $i, 'Asserting has exactly one image');
+    }
+
+    public function testIssue201()
```
k00ni, line 136: Can you leave a comment here which references the issue using the `\@see` tag?

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -274,14 +275,43 @@ public function getText(Page $page = null)
...
                         if (((float) $x <= 0) ||
                             (false !== $current_position_td['y'] && (float) $y < (float) ($current_position_td['y']))
                         ) {
                             // vertical offset
                             $text .= "\n";
-                        } elseif (false !== $current_position_td['x'] && (float) $x > (float) (
+                        } elseif (false !== $current_position_td['x'] && ((float) $x - (float) (
                                 $current_position_td['x']
-                            )
+                            )) > $fontMaxWidth
```
k00ni, line 314: That is complicated to read. Can you move some computation into a variable and use it in the `elseif` instead?

src/Smalot/PdfParser/PDFObject.php
k00ni, line 306: * [ ] For **documentation** purposes: Can you add a comment via

```
/*
 * Get max char size and use as minimum for horizontal offset.
 */
```

at the beginning explaining the content of this if-clause in a few sentences. It would help to understand your code. For instance, why do you have a case for `CIDFontType2` and use `/ 1000` there?

* [ ] Please avoid using `(float) $width` again and again. Instead cast the variable, save the result and use that later on. This will avoid the trap of forgetting the cast sometimes.

src/Smalot/PdfParser/Font.php
```diff
@@ -78,7 +78,15 @@ public function getDetails($deep = true)
 
         $details['Name'] = $this->getName();
         $details['Type'] = $this->getType();
-        $details['Encoding'] = ($this->has('Encoding') ? (string) $this->get('Encoding') : 'Ansi');
+        if ($this->has('Encoding')) {
+            if ($this->get('Encoding') instanceof Encoding) {
+                $details['Encoding'] = $this->get('Encoding')->getDetails()['BaseEncoding'];
```
k00ni, line 83: Can you be certain that the `BaseEncoding` key is always set?

k00ni (changes requested):
Thank you for your contribution!

I made a few comments, nothing critical.

But an important one is:
* [x] Did you only use PDFs which are free to use and with no obligations? (including images, logos...)

src/Smalot/PdfParser/PDFObject.php
k00ni: > should be done outside of foreach loop

OK and what do you mean by that? If you see fit, you can change it.

PaulBehrendtVentoro: yeah just a reminder to myself

k00ni: Ok, then please resolve this conversation if you are done.

k00ni:
What is the status here?

adamchrabaszcz:
any news on this pull request?

Connum:
Pinging @PaulBehrendtVentoro - this looks very promising as a fix for common issues and would be great to see merged into the master.

k00ni:
I will close this at the end of the month, if there is no activity. Feel free to reopen, if you wanna continue your work.

PaulBehrendtVentoro:
regarding license

Image used in PDF generated by me is licensed CC BY-SA 4.0

Did not find anything for issue201.pdf. Just took it from issue #201

PaulBehrendtVentoro:
Be aware this is somewhat breaky to users. I dont know your versioning scheme, but i would not declare this as a bugfix.

k00ni:
> Be aware this is somewhat breaky to users. I dont know your versioning scheme, but i would not declare this as a bugfix.

Thank you for coming back. Our versioning scheme is semantic versioning. When you say "breaky", what exactly do you mean by that? Do people have to change their code?

PaulBehrendtVentoro:
I dont know what people have build around the getText feature. I dont know what kinds of pdf files are out there. Converting a formatted document into string is not trivial. Its hard to say what someone expects in different cases. 

The new solution might **not** add a space, where a space used to be. So i feel they should have to manually decide to update, instead of just installing a bugfix version.

But i am new to open source stuff, so i have no strong opinion on this. Interested to hear other people.

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -274,18 +275,31 @@ public function getText(Page $page = null)
...
                         $x = array_pop($args);
-                        if (((float) $x <= 0) ||
-                            (false !== $current_position_td['y'] && (float) $y < (float) ($current_position_td['y']))
-                        ) {
+
+                        /*
+                         * When converting pdf into text some offsets between text should at least outputted as spaces.
+                         * But small move-text operations may be normal, see https://github.com/smalot/pdfparser/issues/201
+                         *
+                         * To find out what we are handling with, get the max char width for current font and check if offset is bigger than that
```
k00ni, line 283: Better would be: "what we are dealing with"

k00ni:
> The new solution might not add a space, where a space used to be. So i feel they should have to manually decide to update, instead of just installing a bugfix version.

If I understand you correctly, this PR changes the behavior of the `getText` function in way, that might break existing applications? In that case it would require a bump of the major version. Also, is that "fix" not possible without breaking current behavior (or was the behavior faulty in the first place)?

PaulBehrendtVentoro:
> If I understand you correctly, this PR changes the behavior of the `getText` function in a way that might break existing applications? In that case it would require a bump of the major version.

one can argue its a major, because might return different string
one can argue its a minor, because detecting if an offset is to small to be reduced to a space char is a new feature
one can argue its a bugfix, because docs are not precise on how getText should work
i just wanted to start this discussion. but as i said, i have no strong opinion on this.

> Also, is that "fix" not possible without breaking current behavior (or was the behavior faulty in the first place)?
if you ask issue #201 it is faulty behavior

src/Smalot/PdfParser/PDFObject.php
PaulBehrendtVentoro: thanks, you are right.

k00ni:
> one can argue its a major, because might return different string
one can argue its a minor, because detecting if an offset as too small to be reduced to a space char is a new feature
one can argue its a bugfix, because docs are not precise on how getText should work
i just wanted to start this discussion. but as i said, i have no strong opinion on this.

Thank you for the clarification. I have to be that precise, because I wanna determine whether that PR is possible to merge right now or if we need to think about branching out, because it requires a new major version. In my opinion, your first sentence strongly suggests that its the latter, unfortunately. 

Is there a way to make this change optional by "activate" it via parameter? Is that viable?

_About branching out:_ I mean we could create a branch called "1.x" which reflects all commits for a 1.x release. But then we have to maintain multiple branches (0.x and 1.x) and need to keep them in sync. For now I am against it, if I have to take care of it. If someone else wants to help out, I am fine.

I invite others from the community to comment. What do you think about this case?

CC @j0k3r @smalot @Connum @rubenvanerk

j0k3r (commented):
About if we should jump to a major version for that or not, I don't know.
Other tests doesn't seems to fail for now with that PR?

Also, in semantic versioning, some people argue that the v0.x can break whatever they want until the v1.0 is reached. Once the v1.0 is reached, anything that might break should require a major version bump.

k00ni:
> Also, in semantic versioning, some people argue that the v0.x can break whatever they want until the v1.0 is reached. Once the v1.0 is reached, anything that might break should require a major version bump.

Yes, that's right. But keep in mind that we also state that the current API (and its behavior) is considered stable now.

Connum:
As @PaulBehrendtVentoro said, it's difficult to say in what way the parsed text output is used... Of course there could be projects with code relying on the way the text is currently being parsed, that will now break or change its behaviour.

But then again, that is true for (m)any fixes or changes that we make. And looking at the issues that this PR closes, it seems quite obvious to me that the status quo with broken-up words etc. is definitely not a desirable result. So I would regard this as a bugfix.

Maybe a switch to enable or disable this feature would indeed be a good idea, but then I'd tend to switch it on by default. That way, text is parsed more in a way that you'd expect as a human being reading it, and upgrading the library while not breaking existing code would just require a small change to disable the feature.

Reqrefusion:
Actually I was thinking not to write to this pull request, but as I tried all the different pdf files I have and couldn't observe the issue #201  in them, I felt the need to write here.

The problem this pull request fixes is one of the most common problems encountered by the library, which I have experienced a lot with the pdf files I have. So the return is quite high. Other comments mentioned the possibility of a lump sum. First of all, I would like to point out that this is only a possibility. At least that's how I got it. If there is no error in the tests, I don't think it has a big drawback. If there are hesitations about this change, it is more functional to do more tests than to discuss.

I think the payoff in this case is more than the lump sum. Also, the problems that may occur are only hypothetical.

Reqrefusion:
First of all, thank you for giving such a quick and detailed answer to my comment, @k00ni . My opinion is to go over the facts we know, not the possibilities. I understand the concerns here but I don't see any evidence behind the concerns.

Is the problem of spaces in words one of the most common problems?
Yes. This problem has been observed in most of the documents I tested.

Has this pull request caused errors in the test material available?
No. I haven't tested it, but I think it will be mentioned in the comments if there is an error.

Could this pull request cause errors?
Yes. However, this is only a possibility. Although the tests are not comprehensive, they show us that the probability of not causing problems is higher than the probability of a problem.

What is the purpose of this pull request?
To solve the problem of having spaces in words.

Is this pull request a major change?
No. This part is where I find it most desperate to explain. As I said at the beginning, it is necessary to talk about facts, not possibilities. @PaulBehrendtVentoro expressed his concerns and said that this change may be seen differently depending on his point of view. But the purpose of this change and what it does is to fix a bug. How something is done does not change what it is. Although it depends on the perspective of what the change is, it seems to be an bug fix when the basis of the change is taken. The fact that this fix goes to the basics of the code does not change the fact that it is a bug fix.

When I read what I wrote, I realized that no matter how clear I tried to be, I started to get a little confusing towards the end. I hope my opinion is understandable. According to the evolution of the discussion, I can open this idea further.

k00ni:
OK, after latest feedback from @Reqrefusion I gave this another thought and wanna propose the following:

* [ ] @PaulBehrendtVentoro: Is this PR done or do you need time to complete it?
* [ ] @PaulBehrendtVentoro: Please fix conflicting files.
* [ ] A positive review from me **and** someone else (can be anyone, not limited to the collaborators).

Also, we will keep 0.x for now and consider it a bug fix.

To be transparent here a short summary about what lead to my decision:
* Faulty behavior in the first place
* No failing tests
* Multiple positive feedback from @Connum. I am not informed enough to decide what all these changes do and if it's aligned with the specification. But I trust people we proofed multiple times that they know their stuff.

@PaulBehrendtVentoro, I hope you are still with us and would like to see this PR gets merged. Thank you.

k00ni:
Please reopen, if you wanna continue to work on it. Thanks for the effort until now.

PR #324, opened 2020-08-21
Fix error when Font aren't available

j0k3r:
This can happen (like in many places of `Page.php`) when the retrieved element is missing (and the instance if `ElementMissing`. In that case, we just return an empty array (the default value).

Fix https://github.com/smalot/pdfparser/issues/321


k00ni (changes requested):
Looks good. Can you add a simple test which touches/triggers these new lines?

j0k3r:
I'll try but none of the `ElementMissing` case in that file are tested actually 😞

k00ni (approved):
Thank you for taking the time to add a test. :+1:

PR #328, opened 2020-08-31
Handle corrupted PDF

j0k3r:
Also:
- remove some stuff from export (in `.gitattributes`)
- move PHPStan parameters into the config file
- move `TestCase` in `tests` folder (keeping the code base clean)
- refined composer.json (use https for URL)


k00ni (approved):
Changes make sense.

Although I would kept the `Test` namespace. The reason for it was to separate actual tests from test related code, like `TestCase` or test files like PDFs. But this way it might also work. :+1:

PR #337, opened 2020-09-16
replace ext-mbstring requirement with symfony/polyfill-mbstring to support PHP 8.0.0+

### LINKED ISSUE #335: PHP 8:  Call to undefined function mb_convert_encoding()
I'm currently testing PHP 8.0.0 beta3, and using this library. I encountered the following error:

> Call to undefined function Smalot\PdfParser\mb_convert_encoding() in ***\vendor\smalot\pdfparser\src\Smalot\PdfParser\Font.php:500´

running the following code:
```php
    $pdf    = $parser->parseFile('a_valid_pdf_file.pdf');
    $text = $pdf->getText();
```

Connum:
This should fix my own issue #335 - tested and confirmed to be working with PHP 8.0.0 beta 3.

Fixes #335

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "homepage": "https://www.pdfparser.org",
     "require": {
         "php": ">=5.6",
-        "ext-mbstring": "*",
+        "symfony/polyfill-mbstring": "^1.18.1",
         "ext-zlib": "*"
     },
     "require-dev": {


Connum:
side note: it would probably be a good idea to add PHP 8 to the CI (given that it is available at Travis CI yet?)

composer.json
```diff
@@ -16,7 +16,7 @@
     "homepage": "https://www.pdfparser.org",
     "require": {
         "php": ">=5.6",
-        "ext-mbstring": "*",
```
k00ni, line 19: Why removing this requirement? IMHO this should be kept to outline that this library depends on the extension (even with a polyfill). Or doesn't it make sense when using PHP 8?

k00ni (changes requested):
Thank you for this quick PR and a bug fix!

k00ni:
> side note: it would probably be a good idea to add PHP 8 to the CI (given that it is available at Travis CI yet?)

Can you adapt the .travis.yml file? That is what we currently have: https://github.com/smalot/pdfparser/blob/master/.travis.yml#L18

composer.json
Connum: I think as we added the polyfill, the mbstring extension is no longer a requirement, strictly speaking. Sure, it will be used if it's there, but the environment is not required to have it installed or activated in order to be able to run pdfparser.

I don't know if there are any environments doing this, but I could imagine a scenario where a script could check the composer.json for (extension) requirements and refuse the installation of pdfparser (or a depending package) when running on PHP 8 or missing the mbstring extension for some other reason.

k00ni (approved):

k00ni:
OK, I leave this open for some days so that others have a chance to comment. If there are no objections I will merge it.

Connum:
> 
> 
> > side note: it would probably be a good idea to add PHP 8 to the CI (given that it is available at Travis CI yet?)
> 
> Can you adapt the .travis.yml file? That is what we currently have: https://github.com/smalot/pdfparser/blob/master/.travis.yml#L18

I'll have a go at it in another PR and we'll see whether the CI build will go through!

j0k3r (approved):
:+1:

Issue #336, opened 2020-09-16
getDataTm() seems to be broken completely (wrong coordinates and wrongly decoded text content)


Connum:
I created a very simple PDF for testing, because I had issues getting correct x and y coordinates when using `getDataTm()` (unfortunately, I can't provide any other PDFs for privacy reasons...)

=> [xy.pdf](https://github.com/smalot/pdfparser/files/5231236/xy.pdf)

```php
    $parser = new \Smalot\PdfParser\Parser();
    $pdf    = $parser->parseFile('xy.pdf');
    
    $pages = $pdf->getPages();

    foreach ($pages as $index => $page) {
        if (null === $page) {
            continue;
        }
        
        var_dump($page->getDataTm());
    }
```

For this PDF, the Text is not read correctly when using `getDataTm()` (but interestingly, it works when using `getText()` instead). Instead, where the text should be, there's what looks like HEX-encoded strings:

```php
array (size=13)
  0 => 
    array (size=2)
      0 => 
        array (size=6)
          0 => string '1' (length=1)
          1 => string '0' (length=1)
          2 => string '0' (length=1)
          3 => string '1' (length=1)
          4 => string '33.173' (length=6)
          5 => string '665.721' (length=7)
      1 => string '00270050005000430042005300010001' (length=32)
  1 => 
    array (size=2)
      0 => 
        array (size=6)
          0 => string '1' (length=1)
          1 => string '0' (length=1)
          2 => string '0' (length=1)
          3 => string '1' (length=1)
          4 => string '213.173' (length=7)
          5 => string '665.721' (length=7)
      1 => string '00230042005B004300420053' (length=24)
  2 => [...]
```

k00ni:
Hi @izabala, because the functionality was provided by you, can you help us here?

izabala:
Ok. Let me have a look on this!
I let you know.
Ivan Zabala Cell: +58414-929-8558 / +58412-2102012 e-mail: izabala@yahoo.com 

    On Wednesday, 16 September 2020, 05:38:54 am GMT-4, Konrad Abicht <notifications@github.com> wrote:  
 
 


Hi @izabala, because the functionality was provided by you, can you help us here?

izabala:
I will have a look as soon as I can.

Connum:
#340 should fix a part of this issue, the text decoding. I'll try to look into the coordinates myself as well, but I'm not sure if I'm able to wrap my head about the whole thing, not being experienced with the PDF specifications at all.

[...]

PR #338, opened 2020-09-16
update .travis.yml to add a PHP 8 build

Connum:
Let's hope that the Travis CI build runs through... But it doesn't seem to be that trivial according to
* https://github.com/sebastianbergmann/phpunit/issues/4325
* https://github.com/sebastianbergmann/phpunit/issues/3526

Seems like we'd also have to use PHPUnit 9 to increase our chances, but as far as I understood, even with that not all the dependencies support PHP8 yet. Well, we'll see, fingers crossed...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ php:
   - 5.6
   - 7.0
   - 7.1
-  - 8.0
 
 jobs:
   include:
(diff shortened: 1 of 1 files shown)


Connum:
I'm running out of time to follow up on this right now, but it seems that the unit tests will have to be adapted in order to be able to use PHPUnit 9.3. But as an info to anyone willing to look into this further: The "nightly" on Travis CI is currently 8.0.0-dev, so there's no need (and it's not possible I think) to add 8.0 as a version yet. In the last commit, I there fore removed the allowed failure of the nightly build.

k00ni:
> I'm running out of time to follow up on this right now, ... But as an info to anyone willing to look into this further

No problem. Can you tell me when you wanna work on this further, approximately?

Connum:
As this might require a complete rewrite of the way the tests are written, while PHP 8 is still in beta and some APIs might change, I'd prefer to use invest time in more pressing issues, like looking into #336 myself.

I would be fine with closing this and adding it to the backlog until someone else finds the time or PHP 8 ist stable and there's official support for it in PHPUnit and on Travis CI.

k00ni:
> I'd prefer to use invest time in more pressing issues,...

I support that decision. So lets close it.

>  like looking into #336 myself.

That would be great! THanks.

PR #340, opened 2020-09-17
rely on getTextArray() in getDataTm() to extract the texts

Connum:
Instead of parsing the text content manually, falling prone to overseeing things like hexadecimal text encoding etc., I changed the code to rely on the result of `getTextArray()`.

I also added a test case assertion for the exemplary dummy file from #336. The tests for the other files still pass, so there seem to be no side-effects.

Looking forward to others' input, especially from original author @izabala

PR #341, opened 2020-09-18
fix missing BT command before each section (could result in wrong crdinates) and its resetting of Tm

Connum:
See also the discussion in #336.

Some information about what I did and why I did it:
* as the RegEx in `getSectionsText()` assumes all sections starting with BT and ending with ET, we can also safely add the BT command before each section in `$extractedData`
* (we do not, hovever, need to explicitly add ET afterwards, as the reset will happen with the following BT anyway; actually, adding ET explicitly resulted in several tests failing because it messes up the text decoding due to different array sizes when having #340 implemented, because the text has already been added to the `$extractedData` when ET is being encountered!)
* `$Tm` was erroneously reset to `$defaultTl` instead of `$defaultTm` for both BT and ET in the switch/case statement of `getDataTm()`
* I had to adjust some tests, because they would obviously assume a wrong count of commands now with the added BT. Apart from that, the tests run through successfully.
* I also added a test for the BT command in the extracted data

I have also already written a test for the coordinates of the two "Lorem" texts in the sample PDF to be the same, but I'd like to add that after the PR is merged as well as #340, so I can test the decoding and coordinates in one go.

### DIFF
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -301,6 +301,8 @@ class Page extends PDFObject
         } else {
             $sectionsText = $content->getSectionsText($content->getContent());
             foreach ($sectionsText as $sectionText) {
+                $extractedData[] = ['t' => '', 'o' => 'BT', 'c' => ''];
+
                 $commandsText = $content->getCommandsText($sectionText);
                 foreach ($commandsText as $command) {
                     $extractedData[] = $command;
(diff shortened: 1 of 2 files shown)


izabala:
Ups.. the $Tm reset to $defaultTl instead of $defualtTm should be a typo error!!! Thanks for correcting that!

k00ni (commented):
Thank you again for the effort. 

Looks good to me. Is it ready to be merged?

Connum:
From my side, we're good to go! If you merge #340 first, I could add the additional tests with this PR instead of having to post a new one. Whatever you prefer!

k00ni:
OK, will keep it open for a week to allow others to comment.

And we will do it in this order:
* #340
* this PR
* and a new PR for tests

Thanks!

k00ni:
Thank you @Connum!

Issue #334, opened 2020-09-16
Characters are translated correctly in Demo but not in Local


CharmaineLohSiYing:
Hi, 

I have used this file for testing both on the demo page and on my local server (using PHP version 7.4.9).  
[tested-file.pdf](https://github.com/smalot/pdfparser/files/5229114/tested-file.pdf)

This is the output on the demo page:

![demo output](https://user-images.githubusercontent.com/44079990/93282727-a1b6f800-f801-11ea-8a4e-adab89cf9b3f.png)

and this is the output on my local server:

![local output](https://user-images.githubusercontent.com/44079990/93283636-81883880-f803-11ea-9274-6c580e162529.png)

To view the decimal values of the unmatched characters, I replaced "return $use_default ? self::MISSING : $char;" with "return $use_default ? self::MISSING : $dec." ";" in the translatechar() function in Font.php:

![local decimal](https://user-images.githubusercontent.com/44079990/93283843-ea6fb080-f803-11ea-8ef8-b627ee04355c.png)

From this, I managed to match these (abnormally large) decimal values with the characters that they are 'supposed' to be matched with:

- decimal 0 --> 'y'
- decimal 18432 --> 'e'
- decimal 768 --> ' ' (space)
- decimal 22272 --> 't'
- decimal 20992 --> 'o' 
...etc.

1. I noticed that these large decimal values usually appear when my pdf file is generated by converting from a webpage (using the print --> convert to pdf function). Could it be that files which are generated this way have different encoding rules?

2. May I know what changes I can make to attain the same result as on the demo page?

Thank you!

Connum:
Reading "works in the demo but not on my dev system" is ringing my alarm bells now! 😅 I was able to confirm yet again that this is the same regression introduced by 4f4fd10 as a fix for #260, as in #322 and #319! The more I see issues caused by that "fix" for alleged bug that remains yet to be reproduced, in some mysterious PDF file that was never provided... the more I tend to recommending to revert that fix.

Issue #322, opened 2020-08-05
Changes in commit 4f4fd10 lead to more slashes getting stripped, than neccessary


Karmel0x:
Sample pdf: [cv3.pdf](https://github.com/smalot/pdfparser/files/5026842/cv3.pdf)

Where is problem:
![img3](https://user-images.githubusercontent.com/25367564/89385411-d95d5b00-d6ff-11ea-8d8b-e7d638b2e5a4.PNG)

Latest pdfparser commit output:
![img1](https://user-images.githubusercontent.com/25367564/89385078-5cca7c80-d6ff-11ea-9e96-ec080c0d3e3b.PNG)

Output with reversed changes from commit 4f4fd10:
![img2](https://user-images.githubusercontent.com/25367564/89385083-5f2cd680-d6ff-11ea-80ce-5c57f8cf5f73.PNG)

k00ni:
Hi @Karmel0x, so if I understand it correctly commit 4f4fd10 made things worse? Based on the code more slashes get stripped than before. Can someone else approve this observation?

/CC @machekhin

Karmel0x:
It seems so, at least in my case.

j0k3r:
Should we revert the commit and hope to find a better solution?

k00ni:
Can anyone write a simple test which proofs this observation (take the posted PDF from first post and check output)? I am :+1 to revert, if it made things worse.

Connum:
It would be great if the author of #260, @OneSoftIO, could provide a dummy PDF that replicates the issue from before the fix in 4f4fd10. Because when we revert the commit, we'll re-introduce the other bug again. An optimal test case would test both PDFs for correct parsing and help find a solution that fixes both issues at the same time.

Connum:
Here's a test for `ParserTest.php` (saving the PDF from the opening post as `samples/bugs/Issue322.pdf`:

```php
    /**
     * Test that issue related pdf can now be parsed
     *
     * @see https://github.com/smalot/pdfparser/issues/322
     */
    public function testIssue322()
    {
        $filename = $this->rootDir.'/samples/bugs/Issue322.pdf';

        $document = $this->fixture->parseFile($filename);

        $this->assertStringContainsString('this text isn’t working properly, I’ve edited it in Google Documents', $document->getText());
    }
```

This will fail in the current master state, but pass with the changes from `4f4fd10` reverted.
However, as I said, we'd need a sample for #260 as well...

Issue #319, opened 2020-07-09
Fix encoding in demo


romanmatyus:
Please, I have PDF 
[VysledokAukcie-127691.pdf](https://github.com/smalot/pdfparser/files/4899897/VysledokAukcie-127691.pdf)

Output on demo page it's: 

```
100,00


1


EUR


1589,00


0,00


1589,00


HS technology s.r.o., IČO: 52081061, Sídlo: Fraňa Mojtu 22, 94901 Nitra,
Slovenská republika


99,10
...
```

My best encoded output it's:

```
100,00


1


EUR


1589,00


0,00


1589,00


HS technologVUR,26tGOR)UDD0RMWX1LWUD,
 Slovenská republika


99,10
```

How you encode and write text to html?

Thanks

Connum:
I was able to confirm that this is the same regression introduced by 4f4fd10 as a fix for #260, as in #322!
Prior to that change, the text was decoded properly!

PR #342, opened 2020-09-22
revert 4f4fd10 preserving fix for #260, fixing #319, #322 and #334

Connum:
This reverts the change from 4f4fd10 that caused at least issues #319, #322 and #334 (maybe even more...).
With a tiny change to the original code, it was confirmed in #260 that it still fixes that issue.

I also wrote a test for #319 locally, but wasn't sure whether we could use the sample PDF or not:
```php
/**
 * Test that issue related pdf can now be parsed
 *
 * @see https://github.com/smalot/pdfparser/issues/319
 */
public function testIssue319()
{
    $filename = $this->rootDir.'/samples/bugs/Issue319.pdf';

    $document = $this->fixture->parseFile($filename);

    $this->assertStringContainsString('HS technology s.r.o., IČO: 52081061, Sídlo: Fraňa Mojtu 22, 94901 Nitra,\n
    Slovenská republika\n
    \n
    \n
    99,10', $document->getText());
}
```


k00ni (changes requested):
Thank you for your contribution! Your patch seems reasonable.

> I also wrote a test for #319 locally, but wasn't sure whether we could use the sample PDF or not:
* [ ] ~~What does that mean? Is it available now?~~

* [ ] ~~I saw that you asked around in the issues you referenced. Can you give me a short summary about what you found out? That would be helpful.~~

* [ ] ~~Please address my in-code comments.~~

**EDIT:** Solved in next review.

tests/Integration/ParserTest.php
```diff
@@ -91,4 +91,34 @@ public function testIssue267()
...
+    /**
+     * Test that issue related pdf can now be parsed
+     * License of the content taken from https://stackoverflow.com in the sample PDF:
+     * CC BY-SA 2.5 https://creativecommons.org/licenses/by-sa/2.5/
+     *
+     * @see https://github.com/smalot/pdfparser/issues/334
+     */
+    public function testIssue334()
+    {
+        $filename = $this->rootDir.'/samples/bugs/Issue334.pdf';
```
k00ni, line 118: That file contains screenshots from questions and answers out of Stackoverflow. I am not sure if you can include them in that state. Is there a way to make the PDF as small as possible and/or remove usernames and other copyright-related entities (user images). I am not an attorney, but I am afraid that might infringe copyright law.

k00ni:
@romanmatyus (#319), @Karmel0x (#322) and @CharmaineLohSiYing (#334): It would be great if you could give us a short feedback, if this PR fixes your problems. Thank you in advance!

tests/Integration/ParserTest.php
```diff
@@ -91,4 +91,34 @@ public function testIssue267()
         $this->assertEquals(Image::class, \get_class($document->getObjectById('128_0')));
         $this->assertStringContainsString('4 von 4', $document->getText());
     }
+
+    /**
+     * Test that issue related pdf can now be parsed
```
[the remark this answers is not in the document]
Connum: Sure, I just copied that from another issue-specific test. I'll add a short description!

tests/Integration/ParserTest.php
Connum: You're probably right! The CC license only covers the content of the questions and answers. I'll try to reduce the sample PDF to the text.

Connum:
> > I also wrote a test for #319 locally, but wasn't sure whether we could use the sample PDF or not:
> 
>     * [X]  What does that mean? Is it available now?
I can't include that test without a sample file to test against. It means I was able to verify that my changes fix the parsing of the file provided in #319, but we can't simply use that file without consent. I can try to reduce the file, but I fear changing and re-saving it could result in the original issue not occuring in the first place...

>     * [X]  I saw that you asked around in the issues you referenced. Can you give me a short summary about what you found out? That would be helpful.

I basically just confirmed that the issues were caused by the fix in 4f4fd10. In #260, the issue related to that fix, the author of the fix confirmed that with my changes their (not publically available) PDF still worked fine, affirming that we can have all three new issues fixed at the same time without reintroducing the issue raised in #260.

>     * [X]  Please address my in-code comments.
Done, I will address those in an additional commit.

tests/Integration/ParserTest.php
Connum: Oh, by the way...
> remove usernames

that would explicitly go against the attribution part of the CC license... 😉 But I'll try to strip off any images without losing the parts containing the issue (fingers crossed)

CharmaineLohSiYing:
> @romanmatyus (#319), @Karmel0x (#322) and @CharmaineLohSiYing (#334): It would be great if you could give us a short feedback if this PR fixes your problems. Thank you in advance!

Hello! This PR fixes my problems. All the characters in my pdf file are now output properly just like in the demo. Thank you!

Connum:
@k00ni: The sample PDF from #319 is signed, so it cannot be edited directly. Once I print it to a new PDF or import and export it with Illustrator, it does seem to introduce some new issues, but not necessarily the one we need to reproduce...

tests/Integration/ParserTest.php
k00ni: I was not talking about the PDF of #319, but #334!

Connum: Maybe I got you wrong, but well, we talked about both of them really, but inside this review conversation, it's all about #334. And the CC license applies to that PDF.

_(Sorry for resolving this, I'm not too familiar with all the processes of OS collaboration, and I thought that regarding #334, everything was now ok, as I reduced the sample via the additional commit)_

k00ni: Based on the following text from Stackoverflow (Soure: https://stackoverflow.com/help/licensing):

```
As noted in the Stack Exchange Terms of Service and in the footer of every page, all publicly 
accessible user contributions are licensed under Creative Commons Attribution-ShareAlike 
license as follows:

* Content contributed before 2011-04-08 (UTC) is distributed under the terms of CC BY-SA 2.5.

* Content contributed from 2011-04-08 up to but not including 2018-05-02 (UTC) is distributed under 
the terms of CC BY-SA 3.0.

* Content contributed on or after 2018-05-02 (UTC) is distributed under the terms of CC BY-SA 4.0.
The license applicable for each Question and Answer revision is available on the post timeline. 
See this post for more information.

Please read the terms of service and the full legal text of the license carefully for more details on 
how your content can be used and for how you can use publicly accessible content contributed to 
the site by other users.
```
The content of that file we talked about falls under the CC [BY-SA 2.5](https://creativecommons.org/licenses/by-sa/2.5/) (its older than 2011-04-08).

CC BY-SA allows us to "remix, transform, and build upon the material for any purpose, even commercially.", therefore removals of usernames are allowed. Anyway, I did some research and found that usernames are not "creative work", therefore not subject for copyright law.

So file is OK and subject solved. Thank you for helping out here.

k00ni (approved):
> The sample PDF from #319 is signed, so it cannot be edited directly. Once I print it to a new PDF or import and export it with Illustrator, it does seem to introduce some new issues, but not necessarily the one we need to reproduce...

We can't use it, because it contains a logo, but its optional for this PR so we can ignore it.

* [x] I saw that you asked around in the issues you referenced. Can you give me a short summary about what you found out? That would be helpful.

Solved :+1:

k00ni:
I will keep that open for a ~week to allow others to comment. If there are no objections, I will merge.

MrMadClown:
Hi,
might be related to [4f4fd10](https://github.com/smalot/pdfparser/commit/4f4fd108270c32c0e81ea3dceaac2249162f27dc).
The text retrieved from [This PDF](https://www.btg-bestellservice.de/pdf/10070000.pdf) is pretty scrambled in the current version (0.16.2). I reverted to 0.14.0 (pre 4f4fd10) no more scrambling.
I'm still seeing way more spaces in the parsed Text than I can see in the PDF but those might actually be zero-width-spaces not sure.

k00ni:
Can you please check if you have to same behavior with the version in this branch (https://github.com/Connum/pdfparser/tree/fix-260-319-322-334)?

Connum:
I can confirm that the parsing of the PDF provided by @MrMadClown is also fixed with this PR! One example:

Before
```
Gestaltung und Sat4GIGNKPFKU9GUVRJCN)TC�M&GUKIn / Norbert Lauterbach
```

After
```
Gestaltung und Satz: Regelindis Westphal Grafik-Design / Norbert Lauterbach
```

All other instances of scrambled text are fixed as well, no more "strange" characters!

k00ni:
Thank you @Connum and the others for your effort.

PR #343, opened 2020-09-25
fix expected test values after #341 was merged

Connum:
See https://github.com/smalot/pdfparser/issues/336#issuecomment-698830397


k00ni:
Thank you. Its good to know, that only outdated test data were used. As always, I will keep it open for a few days and will merge it, if there are no complains.

k00ni (approved):

k00ni:
Because other PRs may depend on it, I will merge it now.

PR #345, opened 2020-09-26
allow for line breaks when splitting xrefs for id and position

### LINKED ISSUE #19: Undefined Offset in Parser.php causing FollowUp Fatal Error
```
PHP Notice:  Undefined offset: 1 in smalot/pdfparser/src/Smalot/PdfParser/Parser.php on line 178
PHP Warning:  substr() expects parameter 2 to be long, string given in smalot/pdfparser/src/Smalot/PdfParser/Parser.php on line 190
PHP Fatal error:  Call to undefined method Smalot\PdfParser\Object::decodeText() in smalot/pdfparser/src/Smalot/PdfParser/Object.php on line 306
```

http://stackoverflow.com/questions/7116995/is-it-possible-in-php-to-prevent-fatal-error-call-to-undefined-function

Connum:
When looking at #19 and testing with the last PDF file provided there, I found out that the issue came from the raw data having a line break between the xref id and position instead of a space. That resulted in an empty string instead of a number for the position, resulting in an invalid argument for the `substr()` function call.

This fix splits for any regex `\s` character(s) instead of just a space, which resolves the issue.

_Please note that the CI check will expectedly fail for test case `testGetDataTmIssue336` while PR #343 is pending._

edit: fixes #19


k00ni (changes requested):
Thank you! 

* [x] Can you provide a simple test that outlines the problem and demonstrates the fix? That would be great.

Connum:
I tried before making the PR, but it's not that trivial, unfortunately... In order to unit test `parseObject()`, we need to provide `$structure` and `$document` as well, so we'd need to add a test file and do the complete parsing process on it.
I tried manipulating one of the existing sample files into having a line break in the xref position, but it seems those are not in the raw data that is human-readable, but somehow encoded... I don't know how else we could test this for now, other than parsing the provided file (https://github.com/smalot/pdfparser/issues/19#issuecomment-44669205) manually before and after the fix (which I did of course).

k00ni:
You could create a sub class in the test file, which calls the protected method for you, like:

```php
// Parser.php
class Parser {
  protected function parseObject() {
    // secret stuff, you want to test
  }
}

// in the test file
class ParserSub extends Parser {
  public function exposedParseObject(...) {
    return $this->parseObject(...);
  }
}

class ParserTest ... {
   public function testParseObject()
   {
       $fixture = new ParserSub();
       $fixture->exposedParseObject(...);
   }
}
```
Source: https://stackoverflow.com/a/249776/5301527

This way you could inject a prepared `Document` instance, so the new code is being used. What do you think?

Connum:
> This way you could inject a prepared `Document` instance, so the new code is being used. What do you think?

Amazing, thanks for the input! That way, I managed to create a test case that reproduces the error message before the fix, and passes after the fix. It was a bit tricky, and I also extended the `TestCase` base class to allow catching errors that are normally not catchable via a `try/catch` statement. This might come in handy for future tests as well!

tests/TestCase.php
```diff
@@ -46,13 +46,77 @@ abstract class TestCase extends PHPTestCase
...
     public function setUp()
     {
         parent::setUp();
 
         $this->rootDir = __DIR__.'/..';
     }
 
+    public function tearDown() {
+        // if we changed the error handler using catchAllErrors(), reset it now
+        $currentErrorHandler = set_error_handler('var_dump');
```
k00ni, line 77: Using var_dump as error handler seems odd here.

tests/TestCase.php
```diff
@@ -46,13 +46,77 @@ abstract class TestCase extends PHPTestCase
...
+            64 => 'E_COMPILE_ERROR',
+            128 => 'E_COMPILE_WARNING',
+            256 => 'E_USER_ERROR',
+            512 => 'E_USER_WARNING',
+            1024 => 'E_USER_NOTICE',
+            2048 => 'E_STRICT',
+            4096 => 'E_RECOVERABLE_ERROR',
+            8192 => 'E_DEPRECATED',
+            16384 => 'E_USER_DEPRECATED',
+            32767 => 'E_ALL'
```
k00ni, line 115: I think that is a little bit out of scope. Is it really necessary to create a list of PHP error types? Why not just print the error message and available meta data.

k00ni (changes requested):
Thank you for effort to add a test!

I made some in-code comments.

tests/TestCase.php
```diff
@@ -46,13 +46,77 @@ abstract class TestCase extends PHPTestCase
...
+            $this->fail(
+                sprintf('%s: "%s" in %s:%d',
+                    $this->getErrorType($typeNumber),
+                    $message,
+                    $file,
+                    $lineNumber
+                )
+            );
+        };
+    }
```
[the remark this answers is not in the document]
k00ni: You can set settings like display_errors in the PHPUnit config as described in https://stackoverflow.com/a/31279986/5301527 for instance.

tests/TestCase.php
Connum: This is just a workaround because PHP has no function to get the current error handler. (see https://stackoverflow.com/questions/12378644/how-can-i-retrieve-the-current-error-handler) `var_dump` just acts as a dummy here and is immediately reset in the following line via `restore_error_handler()`. PHPUnit itself uses this in its `DeprecationErrorHandler.php`!

tests/TestCase.php
Connum: Yeah, it might be a bit over the top... I thought it could be helpful to see the type of error before the error message in a human-readable way when writing the tests. But if you think it's too much, I can remove it! Please let me know again.

tests/TestCase.php
k00ni: OK, will accept it. But can you please paste a link to the code you refer in 

> PHPUnit itself uses this in its DeprecationErrorHandler.php

tests/TestCase.php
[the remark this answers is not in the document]
Connum: Whoops! The `ini_set('display_errors', 1);` was a leftover from my testing, it's not needed for the test at all! Didn't mean to commit this, sorry about that!

Ok, so this functionality is needed to write a useful test for this issue... Should I revert those changes from this branch and make a new PR for it, and we meanwhile put this PR on-hold until it is merged? I can't think of any other way to really test that this works, other than just letting the test case fail with an error instead of an assertion, which doesn't sound like a sound way to do this to me either.

tests/TestCase.php
k00ni: Please remove it here. But it could be worth looking into it again in another PR.

tests/TestCase.php
Connum: > can you please paste a link into the `tearDown` function to the code you refer in

Sure! To also have this documented here, here it is:
https://github.com/symfony/phpunit-bridge/blob/master/DeprecationErrorHandler.php#L183-L184

tests/TestCase.php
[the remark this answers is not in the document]
k00ni: I am not sure if I missed something. Do you mean that you wanna manually provoke an error to see if this thing works?

In that case PHPUnit may help you with functions like `expectException`: https://phpunit.readthedocs.io/en/8.5/writing-tests-for-phpunit.html#testing-php-errors-warnings-and-notices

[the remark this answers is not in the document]
Connum: Not quite... The original bug will cause an E_NOTICE with the message `Undefined offset: 1 in smalot/pdfparser/src/Smalot/PdfParser/Parser.php on line 178`. When that happens (e.g. someone might inadvertently reverse the fix, or add some code that will lead to the issue egain, which is what we want to catch with the test), the test case will not fail regularly (with an `F` in the command line and counting towards `Failures:`), but will result in an error (marked with an `E` and resulting in an additional `Error:` count). As far as I can see, following tests are still run, but it doesn't seem to me that this is how an error occuring in a test that we're explicitly trying to prevent should be handled.

By catching all types of errors, the test script doesn't result in an error, but we can mark the test as failed instead if such an error occurs. There's no `expectNotice`, unfortunately, or I would have used that and spared the time! 😄 But I might be thinking overly complicated, and maybe having a test result in an error is just fine!

tests/TestCase.php
k00ni: Thank you.

tests/TestCase.php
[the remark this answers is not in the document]
k00ni: Maybe there is a difference in the **test approach**:

I would start with a test that provokes that `E_NOTICE`. Then you deploy your fix. If you run the same code again, the E_NOTICE should be gone. If it stays, the fix was not complete. Your tests should focus on the things you expect. In this case we expect the E_NOTICE to be gone and all tests run without an `F`. Its hard to test unexpected things. 

Another point, we have these options in the PHPUnit config file: https://github.com/smalot/pdfparser/blob/master/phpunit.xml#L10-L12

```
convertErrorsToExceptions="true"
convertNoticesToExceptions="true"
convertWarningsToExceptions="true"
```
That means that PHPUnit will convert everything fishy to an exception, regardless if its a notice or error. This way you don't have to care about E_NOTICE, because it will be thrown as an exception by PHPUnit all the time.

tests/TestCase.php
```diff
@@ -46,13 +46,59 @@ abstract class TestCase extends PHPTestCase
...
+        // if we changed the error handler using catchAllErrors(), reset it now
+        // this is a workaround for PHP providing a way to directly get the current error handler,
+        // by temporarily setting it to var_dump and immediately resetting it.
+        // This is also used in PHPUnit's phpunit-bridge, see
+        // https://github.com/symfony/phpunit-bridge/blob/b6c713f26fd7ec6d0c77934e812f8c3d181e2fd2/DeprecationErrorHandler.php#L183-L184
+        $currentErrorHandler = set_error_handler('var_dump');
+        restore_error_handler();
+        if($currentErrorHandler === [$this, 'catchAllErrorHandler']) {
+            restore_error_handler();
+        }
```
k00ni, line 85: Why call `restore_error_handler` a second time?

tests/TestCase.php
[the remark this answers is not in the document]
Connum: Ok, that's sounds like we're going the `E` route then. I'll remove the complicated testing mechanism and aim for a simpler test. Should a notice pop up in the future, the test will fail as an error. Thanks for your input!

tests/TestCase.php
Connum: The first one just resets the error handler of the workaround that sets it to `var_dump`, the second reset is only executed if the current error handler (which we got via the workaround) is the error handler set via the class itself. This was made to make sure that we do not reset the PHPUnit error handler by accident.

Anayway, I have removed that code from the branch now!

Connum:
I think I found an easier test mechanism now, simply checking for the parsed object in the test data. I also merged the master branch so we can finally see the CI build passing.

tests/TestCase.php
k00ni: > The first one just resets the error handler of the workaround that sets it to var_dump, the second reset is only executed if the current error handler (which we got via the workaround) is the error handler set via the class itself. This was made to make sure that we do not reset the PHPUnit error handler by accident.

Ah OK. I just wanted to make sure that only valid code remained. Your explanation makes sense.

> Anayway, I have removed that code from the branch now!

OK, if the code still works as intended.

k00ni (approved):
Good job! It looks so much cleaner now. 

Is your work done here? If so, I will keep this open for a few days to allow our community to comment.

Connum:
Yes, all done, thanks for your support!

Issue #250, opened 2019-06-26
issue in page->getText() function


satheshk385:
Warning: mb_convert_encoding(): Illegal character encoding specified in /home/ubuntu/workspace/vendor/smalot/pdfparser/src/Smalot/PdfParser/Font.php on line 514 Call Stack: 0.0004 234288 1. {main}() /home/ubuntu/workspace/index.php:0 0.0005 234736 2. require('/home/ubuntu/workspace/wp-blog-header.php') /home/ubuntu/workspace/index.php:17 0.0005 235672 3. require_once('/home/ubuntu/workspace/wp-load.php') /home/ubuntu/workspace/wp-blog-header.php:13 0.0006 236432 4.

GiacomoK:
same here, someone can help pls?

k00ni:
Can you try again with our latest release `0.16.1`?

Connum:
This was fixed in cb1299bc78909f729356b584e59f93da2db45323, which is part of the currently pending PR #349.

PR #349, opened 2020-09-30
revived #257: Properly decode ANSI encodings

Connum:
This is a revival of PR #257, which was closed due to inactivity and its author no longer using pdfparser.

The original PR listed several issues that it supposedly fixed. However, most of them were already fixed due to other code changes or didn't provide any sample PDF files and were subsequently closed.

However, I can now confirm that this definitely fixes issue #202, without breaking any other files or unit tests for me, after I made some additional changes to the original code. I also added a unit test specific to this decoding issue.

It also fixes missing decoded text in a file provided in #19 (depending on PR #345 for another problem).


Connum:
Interesting... The CI build fails due to a test that passes on my environment.

Connum:
Got it! The code checking for UTF-8 strings was removed via the original PR, and its author forgot to check if it's already an UTF-8 encoded string before converting the resulting string to UTF-8 in ANSI environment (this can seemingly happen when ANSI and UTF-8 strings are mixed, which is exactly what the failing test was catching!)

Connum:
PHPstan issues resolved and ready for review!

src/Smalot/PdfParser/Font.php
```diff
@@ -109,7 +115,9 @@ public function translateChar($char, $use_default = true)
      */
     public static function uchr($code)
     {
-        return html_entity_decode('&#'.((int) $code).';', ENT_NOQUOTES, 'UTF-8');
+        // html_entity_decode() will not work with UTF-16 or UTF-32 char entities,
+        // therefore, we use mb_convert_encoding() instead
+        return  mb_convert_encoding('&#'.((int) $code).';', 'UTF-8', 'HTML-ENTITIES');
```
k00ni, line 120: Super tiny thing: can you remove that second white space after the `return`?

src/Smalot/PdfParser/Font.php
```diff
@@ -391,20 +394,18 @@ public function decodeText($commands)
...
     }
 
     /**
      * @param string $text
-     * @param bool   $unicode
      *
      * @return string
      */
-    public function decodeContent($text, &$unicode)
+    public function decodeContent($text)
```
k00ni, line 408: You changed the function head of a public function which is part of the API. It requires us to bump major version. Can we avoid that?

src/Smalot/PdfParser/Font.php
```diff
@@ -448,61 +449,42 @@ public function decodeContent($text, &$unicode)
...
+                    $result .= self::uchr($dec_ap);
+                }
             }
+            $text = $result;
+        } elseif ($this->get('Encoding') instanceof Element &&
+                  $this->get('Encoding')->equals('MacRomanEncoding')) {
+            // mb_convert_encoding does not support MacRoman/macintosh,
+            // so we use iconv() here
+            $text = iconv('macintosh', 'UTF-8', $text);
+        } elseif (!mb_check_encoding($text, 'UTF-8')) { // don't double-encode strings already in UTF-8
```
k00ni, line 486: Please move that comment on the next line.

src/Smalot/PdfParser/Header.php
```diff
@@ -60,6 +60,13 @@ public function __construct($elements = [], Document $document = null)
         $this->document = $document;
     }
 
+    public function init()
+    {
+        foreach ($this->elements as $name => $element) {
+            $element->init();
+        }
```
k00ni, line 67: Can we do that in the constructor instead?

src/Smalot/PdfParser/Page.php
```diff
@@ -107,10 +107,19 @@ public function getFont($id)
             return $fonts[$id];
         }
 
-        $id = preg_replace('/[^0-9\.\-_]/', '', $id);
+        // $id = preg_replace('/[^0-9\.\-_]/', '', $id);
```
k00ni, line 110: Please remove that line

tests/Integration/ParserTest.php
```diff
@@ -77,6 +77,46 @@ public function testParseFile()
...
+        $testString_armenian = "լոռեմ իպսում դոլոռ սիթ ամեթ վիս ին իմպեդիթ ադմոդում ծու ապպառեաթ սծռիպսեռիթ մել մել եթ \nդոմինգ ծոնսեքուունթուռ ծիվիբուս վիվենդում պռոդեսսեթ ադ մեի թիբիքուե ապպառեաթ սիմիլիքուե թե \nվիմ վիխ ծասե սեմպեռ դոլոռեմ եխ եամ եա սթեթ մեդիոծռեմ ծոնսեթեթուռ ռաթիոնիբուս ինթելլեգամ \nմել թե";
+        $testString_georgean = "ლორემ იფსუმ დოლორ სით ამეთ ესთ ეთ სონეთ ზრილ მელიუს ელიგენდი თორყუათოს \nელოყუენთიამ ესთ ეხ უსუ ფალლი ალთერა ცეთეროს ინ ეთ ომითთამ თრაცთათოს ჰის ეუ ველ \nალთერუმ ვოლუფთათუმ მაზიმ ფერთინახ ჰენდრერით ინ ფრი ნეც ინ თემფორ ფეთენთიუმ ვერო \nფოსთულანთ ელოყუენთიამ უსუ ნე ან ყუი ლიბერ ეფიცური ასსუევერით იდ ნიბჰ ყუას ჰაბემუს სეა";
+        $testString_korean = "그 임기는 4년으로 하며. 이 경우 그 명령에 의하여 개정 또는 폐지되었던 법률은 그 명령이 승인을 얻지 못한 때부터 당연히 효력을 \n회복한다. 가부동수인 때에는 부결된 것으로 본다. 법률과 적법한 절차에 의하지 아니하고는 처벌·보안처분 또는 강제노역을 받지 \n아니한다.";
+        $testString_western = 'ÄÖÜöäüßẞ Ññ¡¿ øÅå';
+        // @todo the other languages in the test document need work because of issues with UTF-16 decoding (Chinese, Japanese) and missing right-to-left language support
```
k00ni, line 95: Please move that TODO to the function header.

k00ni (changes requested):
Thank you for the effort of reviving #257 and fixing it.

**I have a few things:**
* [x] I added #19 and #202 as being closed when this PR gets merged. Was that correct?
* [x] Please move `src/Smalot/PdfParser/Encoding/PostScriptGlyphs.json` to another place, because `src` should only contain PHP files. Maybe `encoding/` is a good place? And isn't it more consistent to use PHP instead of JSON as file format?
* [x] When using PHP instead of JSON for `PostScriptGlyphs.json` the file `src/Smalot/PdfParser/Encoding/PostScriptGlyphs.php` can be removed, because no JSON decoding is required as well as `getCodePoint` can be done at the place, when its needed.
* [ ] See my comment about `src/Smalot/PdfParser/Font.php`::decodeContent

src/Smalot/PdfParser/Font.php
Connum: I could leave the parameter in for now, so it just wouldn't be used anymore. What would be the best way to mark it as deprecated?

src/Smalot/PdfParser/Font.php
Connum: Will be fixed with the next commit. Thos looks like something the cs-fixer should catch, doesn't it?

Connum:
>     * [ ]  I added #19 and #202 as being closed when this PR gets merged. Was that correct?
Ths is correct!

>     * [ ]  Please move `src/Smalot/PdfParser/Encoding/PostScriptGlyphs.json` to another place, because `src` should only contain PHP files. Maybe `encoding/` is a good place? And isn't it more consistent to use PHP instead of JSON as file format?
>     * [ ]  When using PHP instead of JSON for `PostScriptGlyphs.json` the file `src/Smalot/PdfParser/Encoding/PostScriptGlyphs.php` can be removed, because no JSON decoding is required as well as `getCodePoint` can be done at the place, when its needed.

What about keeping PostScriptGlyphs.php, but instead of having `getGlyphs()` parse the data from the JSON file, have it return the data directly as an array, like the other Encoding classes do with `getTranslations()`?

>     * [ ]  See my comment about `src/Smalot/PdfParser/Font.php`::decodeContent

See my comment in the conversation above! :)

src/Smalot/PdfParser/Font.php
k00ni: I suggest we use the follow message here: `@deprecated Usage of second parameter $unicode is deprecated. It might be removed in a future release.`

I also noticed that without `$unicode` being used inside, the behavior of this function changes. Is that right? That might also be a blocker and requires the bump of the major version.

src/Smalot/PdfParser/Font.php
k00ni: It should, but I don't know which rule to enforce here.

k00ni:
> What about keeping PostScriptGlyphs.php, but instead of having getGlyphs() parse the data from the JSON file, have it return the data directly as an array, like the other Encoding classes do with getTranslations()?

That even better. Handle the licence information as you see fit.

src/Smalot/PdfParser/Font.php
Connum: With the changed code, the `$unicode` variable is no longer used from the arguments. There still is a $unicode variable in the function body, but it's always set with a value returned by `mb_check_encoding()`.

Connum: The `@deprecated` marks the whole method as deprecated (at least in VS Code with Inteliphense).
grafik

Maybe it would be better to add a description to the parameter itself? This looks fine:
grafik

src/Smalot/PdfParser/Header.php
Connum: As far as I see, the elements are never set by anything other than the constructor. That could be an argument for moving it to the constructor, right after the elements are set.

However, this could hypothetically have a performance impact. Right now, this init method is explicitly called from the `Document`'s method of the same name, which is called from its method `setObjects()`. Imagine a scenario where you bulk-process hundrets or thousands of PDF files, where many of them will expectedly fail (secured pdfs, other parsing errors, anything that can happen before `setObjects()` is called within `parseContent()` of the `Parser`...). The `init` of every `Encoding` and `Font` that is encountered will be called as soon as the header is built. And that looks quite expensive, with several RegExes used especially for `Font`s, where `loadTranslateTable()` is called from within init.

That case might be quite constructed and improbable, but I wouldn't want to risk it just for saving a few lines of code.
Also, I think when Sebastien implemented it that way for `$object->init()` in `Document->init()`, which could probably also be moved to the object's constructor, he did that on purpose, so I'd tend to keep it that way for the header as well!

src/Smalot/PdfParser/Font.php
k00ni: You are right, good point.

src/Smalot/PdfParser/Header.php
k00ni: Good arguments, so we keep it as it is.

src/Smalot/PdfParser/Font.php
Connum: Done!

Connum:
All points are done from my side, thanks again for your input!

k00ni (approved):
Thank you again for your effort @Connum! Looks good.

Connum:
Merged in the current master, resolving a conflict after #345 was merged.
Also found out this fixes #250.

Issue #331, opened 2020-09-14
Does not parse pages other than the first page


Reqrefusion:
It converts only the first page of a 3 page pdf file. It does not do anything related to other pages. You can also observe this on the demo page.
Related pdf file:
https://www.resmigazete.gov.tr/ilanlar/eskiilanlar/2020/09/20200914-4-4.pdf

Reqrefusion:
Acrobat PDFMaker 10.1 for Word is used for this file, but similarly, files converted to PDF have the same error.

Connum:
Interesting... You can use `$pdf->getPages();` to get all three pages, but `->getText()` returns an empty string on the last two of them. Furthermore, the third page is of type `Smalot\PdfParser\PDFObject` instead of `Smalot\PdfParser\Page`. That should be handled in `Pages.php` inside the if/else construct in `getPages()`. In my opinion this should never return anything else than `Page`s!

Connum:
I invested several hours now, trying to figure out what's wrong here...
I think one thing we should do at least is to change the `} else {` in `Pages.php` to
```php
} elseif ($kid instanceof Page) {
```

`getPages()` should only ever return array elements of the type `Page`, but evidently, `'Kids'` can hold other types too that we wouldn't want to pollute the resulting array with!

Other than that, I'm at a bit of a loss here...

Reqrefusion:
@Connum Thanks for trying but I didn't see any improvement. Maybe in the future this problem will be resolved somehow. Thank you again.

Connum:
Yeah, that wasn't a fix for the issue itself, just something I noticed we'd want to change while investigating it. For the issue itself, maybe I'll find some more time and motivation to investigate it, otherwise hopefully someone else will.

hpvd:
ahh, found the same problem when trying the demo for several documents https://github.com/smalot/pdfparser/issues/416
-> are the any multipage files working fine?

hpvd:
Another example file can be found here: https://github.com/smalot/pdfparser/issues/416#issuecomment-829056645

PR #350, opened 2020-10-03
getPages() should only ever return elements of type 'Page'

Connum:
While investigeting #331, I found that `Pages->getPages()` will return anything inside the `Kids` of a `Pages` object, regardless of type. However, one would expect it to only return actual `Page` objects, from which text can be parsed and which count towards the page count in the `Document`'s details.

In case another `Pages` object is encountered, its `Pages` are already being merged into the resulting array. Other types might need to be processed in a specific way to get `Page` objects out of them, which is probably the cause for #331. This does _not_ fix the issue, but is a move in the right direction


Reqrefusion:
I tried the change. It doesn't fix #331  but it doesn't create it in another problem.

Connum:
By the way, this _does_ fix a very real situation. When looping over the results of `getPages()` with the document provided in #331 and calling `getDataTm()`

```php
$pdf    = $parser->parseFile($file);
$pages  = $pdf->getPages();

foreach ($pages as $index => $page) {
    var_dump($page->getDataTm());
}
```

That will throw
> **Fatal error:** Uncaught Error: Call to undefined method Smalot\PdfParser\PDFObject::getDataTm() in *** Stack trace: #0 ***: parsePDF('test/Issue331.p...') #<span>1</span> {main} thrown in *** on line 95
As end user, I shouldn't have to implement type checks here, but should reliably receive `Page` elements.

@Reqrefusion could you please provide a modified version of the pdf file with a all the probably copyrighted content removed (or replaced with some dummy text), keeping the three pages otherwise intact? That way, we could use that file to create a unit test!

Reqrefusion:
@Connum No need to worry about copyright. Related pdf file is a part of Turkey Official Gazette. This means the following. Official Gazette exist for; laws, regulations, communiqués and similar publications public disclosure. Due to its content, the Official Gazette is not subject to copyright. There are many court orders and regulations about this.

The official newspaper is published daily to serve the public. I am also adding the current issue of the Official Gazette.
https://www.resmigazete.gov.tr/eskiler/2020/10/20201004.pdf

Connum:
Added the sample and a test case. Regarding my comment concerning the page count, it might be a bit unconventional to assert something that we actually know to be incorrect. However, I spent several hours already trying to get to the bottom of this, and I think the cause is rooted quite deep in the parsing process... So I think we don't want to miss out in case this is "accidentally" fixed by some other code changes. In case that happens, the failing test will act as a reminder to resolve the issue, and it will prevent that a fix as a by-effect of some other changes will then be reverted without anyone noticing.

k00ni (approved):
Good work on hardening the library.

PR #351, opened 2020-10-06
Added a few tests to check current behavior of RawData/FilterHelper.php

k00ni:
## About

This PR extends test coverage of `RawData/FilterHelper.php` by adding a few tests which check current behavior (ref: https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/RawData/FilterHelper.php).

I had problems testing the following filters, because I couldn't get my original data in the end: 
* LZWDecode
* RunLengthDecode

### LZWDecode

Because LZW is using binary data an online encoder was not usable. I tried the following code to generate a string during the tests but had no success:
* https://webdevwonders.com/lzw-compression-and-decompression-with-javascript-and-php/
* https://www.programmersought.com/article/87352186122/
* https://github.com/vrana/php-lzw/blob/master/lzw.inc.php#L13

### RunLengthDecode

For "RunLengthDecode" I am not sure if our decoder or the online tools I tried are faulty (or I lack basic understanding of it). AFAIK the string "Compressed string" results to "C1O1M1P1R1E1S2E1D1S1T1R1I1N1G1" using the online encoder https://www.dcode.fr/rle-compression. But the test fails when trying to decode it.

---

Any feedback on these two is appreciated.

### DIFF
--- a/tests/Integration/RawData/FilterHelperTest.php
+++ b/tests/Integration/RawData/FilterHelperTest.php
@@ -45,6 +45,34 @@ class FilterHelperTest extends TestCase
         $this->fixture = new FilterHelper();
     }
 
+    /*
+     * Tests for filter ASCII85Decode
+     */
+
+    public function testDecodeFilterASCII85Decode()
+    {
+        $compressed = '6Z6g\Eb0<5ARlp)FE2)5B)'; // = Compressed string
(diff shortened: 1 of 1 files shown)


j0k3r (approved):
Good job in finding these strings to decode 👍

k00ni:
OK, we will take care of LZW and RunLengthDecode some time in the future.

PR #369, opened 2020-11-18
ElementName.php: Improve regex performance by removing group naming

mardc21:
I removed group naming from the regular expression to decrease memory usage and improve performance, in very large files I was experiencing memory limit problems.


k00ni:
So it basically does the same as before but with less overhead?

Can we test that change in some way?

k00ni:
Ping @mardc21

Issue #305, opened 2020-06-05
Idea: use configs/profiles to work with PDFs


k00ni:
***Disclaimer:** This issue is just a starting point for some ideas i got after working with PDFParser in the last couple of days. So for now, please don't expect anything from it besides some discussion. If someone wants to create PR's based on this, you are welcome.*

---

**Issue:** There are scenarios in which you need a certain "config" when parsing a PDF. E.g. the PDF requires certain settings or has features, which PDFParser can't handle but you wanna get the most out of it.

**Here some use cases which might fit:**

* https://github.com/smalot/pdfparser/issues/259: @nickano asks about making PDFObject's `$font_space` value configurable from outside.
* https://github.com/smalot/pdfparser/issues/294: define your own threshold for a variable to keep `\n` in parsed text.
* https://github.com/smalot/pdfparser/pull/87#issuecomment-638621328, its about how to separate table cell content after it was parsed. In some cases it won't matter, in some it does.

**Related areas:**
* parsing process
  * certain thresholds, variable values
  * don't run certain functions (e.g. maybe they crash for a certain file cause features are not supported)
  * ...
* data enrichment afterwards
  * format table content
  * handle attachments or pictures
  * ...

**Implementation sketch**: Use a config object which gets passed to the parser. Its values will be passed to certain functions along the parsing process to override default values or do activate/deactivate PDFParser features/methods.

Would that be helpful? What do you think?

nickano:
Defining a config to pass into the object sounds like a much cleaner way of doing it

Nick Allan
Computer Systems Manager
Building Validation Solutions Ltd
01204 328 560
n.allan@bv-solutions.co.uk
www.bv-solutions.co.uk
Waters Meeting House, 1 Waters Meeting Road, Bolton, BL1 8HQ
 
​
​CONFIDENTIALITY: The information in this e-mail and any attachments is confidential. It is intended only for the named recipient(s). It may contain privileged and confidential information that is exempt from disclosure under English Law and if you are not an intended recipient, you must not copy, distribute or take any action in reliance on it. Copyright in documents created by or on behalf of Building Validation Solutions Ltd remains vested in Building Validation Solutions Ltd. If you are not the named recipient please notify the sender immediately. Whilst every possible precaution has been taken the contents of this e-mail or attachment may contain software viruses that could damage your own computer system. You should therefore carry out your own virus checks before opening the attachment. 
​
[...]

PR #375, opened 2020-12-15
First step towards a config object (e.g. adapt font space limit)

### LINKED ISSUE #259: getFontSpaceLimit - class variable?
Is there any chance you could make this a configurable class variable?  Dirtyish quick way would be to set it statically on the Font object before you create a Parser instance and away we go?

At the moment we have to manually change the file every time we update the package in composer and we can't really extend the classes efficiently to fix it that way.

k00ni:
This PR is based on #305 and adds a config object to the library. It shall contain scalar values only (e.g. font space limit) which are changeable from the outside. That is helpful if PDFParser fails to parse a particular PDF and one wanna try different config values to fix it.

There are only a few code changes and it is fully backward compatible, although it mostly supports the happy path: 
* use `Parser` class, 
* parse a file and 
* work with the result. 

In case one goes deeper and does more manually, one has to pass config object himself eventually at some point.

PHP-CS-Fixer fixed many files which pollute the diff view. To see my implementation only, please use https://github.com/smalot/pdfparser/pull/375/commits/72696efb4b4d7b45769a268e38472421a308a9b0.

I consider this implementation very basic (and not final) and wanna hear the community opinions. Outside tests are appreciated.

It replaces #358 (same functionality but uses a singleton).

Fixes #259 

CC @j0k3r @Connum @nickano @butschster


Connum:
Well done, I really like the completely backwards-compatible approach. I don't have the time to thoroughly test it at the moment, but after a quick code review, it looks great! Looking forward to more configuration options in the future, this adds some real value to the library!

k00ni:
Linter case fails because of `Use of undefined constant XDEBUG_CC_UNUSED`. It seems fixed upstream (https://github.com/xdebug/xdebug/pull/699), but has anyone an idea how to fix it here?

src/Smalot/PdfParser/Parser.php
```diff
@@ -46,16 +46,30 @@
...
      */
     protected $objects = [];
 
     protected $rawDataParser;
 
-    public function __construct($cfg = [])
+    public function __construct($cfg = [], Config $config = null)
     {
         $this->rawDataParser = new RawDataParser($cfg);
+        $this->config = null == $config ? new Config() : $config;
```
j0k3r, line 64:
```suggestion
        $this->config = $config ?: new Config();
```

src/Smalot/PdfParser/Parser.php
```diff
@@ -46,16 +46,30 @@
  */
 class Parser
 {
+    /**
+     * @var Config
+     */
+    protected $config;
```
j0k3r, line 52: Why not define it as `private`? (same question apply on other files)

```suggestion
    private $config;
```

j0k3r (commented):
Looks ok to me.
About the xdebug issue, I think you have to force that env var `XDEBUG_MODE=coverage`

src/Smalot/PdfParser/Parser.php
k00ni: I kept the same visibility as other properties, which is `protected` and not `private`. But we can change it, especially because there is a getConfig function.

j0k3r:
You should have rebased your PR instead of merging master inside.

k00ni:
Oh, I thought it doesn't matter if we squash commits in the end.

j0k3r:
Oh yeah of course, then don't forget to squash :)

PR #376, opened 2020-12-15
Ditch Travis and use GitHub Actions

j0k3r:
Travis is less and less available for open source project. Plus we need to migrate the project from .org to .com.
Anyway, GitHub Actions is now _de facto_ the way to go.


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,186 @@
...
+        php:
+          - "7.2"
+          - "7.3"
+          - "7.4"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 2
```
k00ni, line 31: Just wondering why `2` and not `1` here?

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "7.4"
+          tools: cs2pr, composer:v1
```
k00ni, line 26: What is `cs2pr` for?

k00ni (approved):
Very nice! I was wondering myself about Travis, runs were delayed and required days before they start.

These new CI cases are helpful too.

.github/workflows/continuous-integration.yml
j0k3r: Good question, I might have picked it from an other repo as example. I'll remove it as the default value is `1`

.github/workflows/coding-standards.yml
j0k3r: Instead of dumping error inside the build, cs2pr will inline comment in the _files changed_ tab.
See that example 
image

k00ni: Super cool!

k00ni (approved):

k00ni:
We should merge this before #375, because of extended CI.

Issue #372, opened 2020-11-23
PDF with xref table corrupted : memory leak


alkaan:
**Context** : a pdf with corrupted xref table (and with password protection change) do an infinite loop.
**file** : RawData/RawDataParser.php
**method** : getRawObject()
When an object is corrupted there are cases where the $offset don't change, 
These are the case in loops "do { } while loop. " in case conditions '[' or ']' and  '<' or '>' .
```php
case '[':   // \x5B LEFT SQUARE BRACKET
            case ']':  // \x5D RIGHT SQUARE BRACKET
                    // array object
                    $objtype = $char;
                    ++$offset;
                    if ('[' == $char) {
                        // get array content
                        $objval = [];
                        do {
                            // get element
                            $element = $this->getRawObject($pdfData, $offset);
                            $offset = $element[2];
                            $objval[] = $element;
                        } while (']' != $element[0]);
                        // remove closing delimiter
                        array_pop($objval);
                    }
                    break;

            case '<':  // \x3C LESS-THAN SIGN
            case '>':  // \x3E GREATER-THAN SIGN
                    if (isset($pdfData[($offset + 1)]) and ($pdfData[($offset + 1)] == $char)) {
                        // dictionary object
                        $objtype = $char.$char;
                        $offset += 2;
                        if ('<' == $char) {
                            // get array content
                            $objval = [];
                            do {
                                // get element
                                $element = $this->getRawObject($pdfData, $offset);
                                $offset = $element[2];
                                $objval[] = $element;
                            } while ('>>' != $element[0]);
                            // remove closing delimiter
                            array_pop($objval);
```
this is similar of the loop "do{} while" in method getIndirectObject() where the $offset case that don't change  is treated with memorising the last offset and comparing the old with the new. If they are same, we exit of the loop.
```php
protected function getIndirectObject(...) { 
....
do {
            $oldOffset = $offset;
            // get element
            $element = $this->getRawObject($pdfData, $offset);
            $offset = $element[2];
            // decode stream using stream's dictionary information
            if ($decoding and ('stream' == $element[0]) and (isset($objdata[($i - 1)][0])) and ('<<' == $objdata[($i - 1)][0])) {
                $element[3] = $this->decodeStream($pdfData, $xref, $objdata[($i - 1)][1], $element[1]);
            }
            $objdata[$i] = $element;
            ++$i;
        } while (('endobj' != $element[0]) and ($offset != $oldOffset));
....
```
[...]

PR #377, opened 2020-12-21
Fix infite loop when xref table corrupted

partulaj:
Hello,

I open a PR with the modification suggested for issue #372 if you can take a look

Thank you

### DIFF
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -631,58 +631,60 @@ class RawDataParser
 
             case '[':   // \x5B LEFT SQUARE BRACKET
             case ']':  // \x5D RIGHT SQUARE BRACKET
-                    // array object
-                    $objtype = $char;
-                    ++$offset;
-                    if ('[' == $char) {
+                // array object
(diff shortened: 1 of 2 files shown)


k00ni:
@partulaj thank you for your contribution!

Can you provide a few tests please, which show problems you solved with your fix?

partulaj:
Unfortunately I can't provide the file on which we had the issue since it contain some confidential informations, but I will try to take a closer look to see if I can manage to find a way to make this testable

k00ni:
If we can figure out a minimal example which causes the infinite loop, I can help you with tests (see my last comment on #372).

k00ni:
Simple test added, thanks for your help @partulaj. Good to go.

aurelianzaha:
@k00ni are you planning to do a new release after this fix?

k00ni:
I didn't plan a new release, but I can make a new one soon, no problem.

aurelianzaha:
> I didn't plan a new release, but I can make a new one soon, no problem.

That would be appreciated, so I can properly include the new version in the composer file.

k00ni:
New release published: https://github.com/smalot/pdfparser/releases/tag/v0.18.1

Issue #104, opened 2016-05-18
Memory Leak


harinderbachhal:
Hi i am using this to get pdf text works well, but there is memory leak in this, i am using it as under

```
 $parser = new \Smalot\PdfParser\Parser();
$pdf = $parser->parseContent($getdata[0]);
 $pages = $pdf->getPages();
 foreach ($pages as $page) {
                 echo $page->getText();
 }
unset($parser);
unset($pdf);
unset($page);
unset($pages);
echo '<h1>Memory =', round(memory_get_usage() / 1000), ' KB</h1><br>';
```

Memory =63226 KB

How do i fix this, how can i release memory which is used.

smalot:
It's really hard to fix a such issue.
There is many circular references between objects which block memory garbage.
Usually we use "__destruct" to break such behavior by settings properies to "null" or unset them to help garbage collector

harinderbachhal:
where do i use __destruct, in this library? Parser.php this file or have to read full library ? and make changes .

citionzeno:
I am having the same problem. Trying to parse multiple PDF files within the same script ended up with a huge memory leak. With PHP 7 (or PHP >= 5.3), you can use `gc_collect_cycles();` to call the garbage collector in order to delete objects with circular references. The memory usage goes back to normal for me after this call.

vishaldevrepublic:
Having same issue, Any solution for this. 

```
$parser = new \Smalot\PdfParser\Parser();
$pdf = $parser->parseFile(asset($pdf_file)); // [ Stuck with "Allowed memory size of 1073741824 bytes exhausted (tried to allocate 26811 bytes)" message here. ]
$pages = $pdf->getPages(); 
```

yapsr:
Thank you @citionzeno, your suggestion works for me.

citionzeno:
Glad it helped. It works if you are trying to parse multiple small files. For one single large file however, like @vishaldevrepublic, I don't know how to do. Some deep work into the library might be needed.

amineharoun:
> I am having the same problem. Trying to parse multiple PDF files within the same script ended up with a huge memory leak. With PHP 7 (or PHP >= 5.3), you can use `gc_collect_cycles();` to call the garbage collector in order to delete objects with circular references. The memory usage goes back to normal for me after this call.

Where do i have to do that garbage collection in the code please ?

fm89:
[...]

PR #383, opened 2021-01-01
Introduced PHP 8 support; removed support for PHP 5.6 + 7.0

k00ni:
**Updated 2021-03-10**:

* Introduced **PHP 8** support (mostly test related changes)
* ❗ **Removed support for PHP 5.6 + 7.0**
* Extended **ElementXRef::equals** to allow equality checks in PHP 8+, otherwise tests would fail because `5_0` != `5` (see TODOs below)
* **Separated dev-tooling** (e.g. PHPUnit) from the library by using a dedicated folder named `dev-tools`, which contains a separate `composer.json` file. A detailed explanation why I did that can be found in https://github.com/smalot/pdfparser/pull/383#issuecomment-795322159

**TODOs:**
* [ ] Feedback on the following addition by @smalot: https://github.com/smalot/pdfparser/pull/383/files#diff-d419ba202a4a50b99b1a7004d91d330081176213a0e92f696ec881f3cfefa169R59-R76
* [ ] Create new major version after merging this


k00ni:
Note to myself:
* [ ] merge in master

j0k3r (approved):
> I am not sure about the following addition

If that fix the issue, I'm ok with it.

k00ni:
> If that fix the issue, I'm ok with it.

It fixes the test for specific parameters, but I am not sure if it covers it 100%. Can you give me a hint what range of parameter values `equals` has to cover?

j0k3r:
> Can you give me a hint what range of parameter values `equals` has to cover?

I've no idea...

k00ni:
Maybe @Connum?

k00ni:
We should wait with a merge until further information from #104. Its about a memory leak, but some still use PHP 5.6. I would like to sort it out before we cut off PHP 5.6 + 7.0 support.

src/Smalot/PdfParser/Element/ElementXRef.php
```diff
@@ -56,6 +56,24 @@ public function getObject()
...
+         * would fail (= 5_0 and 5 are not equal in PHP 8.0+).
+         */
+        if (
+            true === is_numeric($value)
+            && true === \is_string($this->getContent())
+            && 1 === preg_match('/[0-9]+\_[0-9]+/', $this->getContent(), $matches)
+        ) {
+            return (float) ($this->getContent()) == $value;
+        }
+
```
k00ni, line 76: @smalot Do you see any issues with that addition?

DEVELOPER.md
```diff
@@ -0,0 +1,40 @@
...
+## Development Tools and Tests
+
+Our test related files are located in `tests` folder.
+Tests are written using PHPUnit.
+
+To install (and update) development tools like PHPUnit or PHP-CS-Fixer run:
+
+> make install-dev-tools
+
+Development toosl are getting installed in `dev-tools/vendor`.
```
j0k3r, line 16:
```suggestion
Development tools are getting installed in `dev-tools/vendor`.
```

j0k3r (commented):
What the benefit of moving dev tools deps in a dedicated folder?

DEVELOPER.md
k00ni: Good catch

k00ni:
Semantic Versioning is not clear what happened if a supported environment (in this case PHP 5.6 + 7.0) is dropped. I found a few issues discussing this, like https://github.com/semver/semver/issues/526#issuecomment-519616743.

I would follow the argument and create a new major version (= 1.0.0). 0.18.2 would be the last `0.x` version.

smalot:
Hi @k00ni 
I agree with you to create a new major release "1.x" and to keep old release under the "0.x".
It seems to be the way to reduce any risk of BC Break for current users.

src/Smalot/PdfParser/Element/ElementXRef.php
smalot: Can you remind me when does this method is called and why we try to validate equality between "5" and "5_0" ?

.github/workflows/continuous-integration.yml
```diff
@@ -107,16 +100,16 @@ jobs:
       - name: "Install dependencies with Composer"
         uses: "ramsey/composer-install@v1"
 
-      - name: "Run PHPUnit (with coverage)"
-        run: "php vendor/bin/simple-phpunit -v --coverage-clover coverage/clover.xml"
+      - name: "Install dev tools"
+        run: "make install-dev-tools"
+
+      - name: "Run PHPUnit"
+        run: "php dev-tools/vendor/bin/phpunit -v --coverage-clover coverage/clover.xml"
```
smalot, line 107: Should it be better to use `make run-phpunit ....` instead to hide complexity ?

Makefile
```diff
@@ -0,0 +1,11 @@
...
+	composer update --working-dir=dev-tools
+
+run-php-cs-fixer:
+	dev-tools/vendor/bin/php-cs-fixer fix
+
+run-phpstan:
+	dev-tools/vendor/bin/phpstan/phpstan/phpstan analyse
+
+run-phpunit:
+	dev-tools/vendor/bin/phpunit
```
smalot, line 11: may be allow parameters to be added at the end of command line

smalot (commented):
I need to understand why ElementXRef contains this fix

src/Smalot/PdfParser/Element/ElementXRef.php
k00ni: If you remove this code, [assertion](https://github.com/smalot/pdfparser/blob/php8/tests/Integration/Element/ElementXRefTest.php#L115) in `testEquals` of ElementXRefTest.php will fail.

Makefile
k00ni: Done using optional `$(ARGS)` parameter, see https://github.com/smalot/pdfparser/pull/383/commits/a65d972f0a30ed7fe5e95fdf7e307893099c3fef

If you have a better idea how to do that, please tell me. I am not a Makefile geek.

.github/workflows/continuous-integration.yml
k00ni: Done using optional `$(ARGS)` parameter, see https://github.com/smalot/pdfparser/pull/383/commits/a65d972f0a30ed7fe5e95fdf7e307893099c3fef

If you have a better idea how to do that, please tell me. I am not a Makefile geek.

smalot (approved):

smalot:
Thanks @k00ni, everything seems to be good for me

k00ni:
@smalot Just to be sure we didn't forget something. What about https://github.com/smalot/pdfparser/pull/383/files/f595ac26308b167c3a02c3a6a77abd653c47b61e#r592171760 (my addition to `ElementXRef.php`)?

k00ni:
We may merge #401 and #402 before this so that older installations can used latest feature.

k00ni:
Because there is no activity/progress in referenced issues I am inclined to merge this PR and prepare a major release.

Issue #397, opened 2021-02-26
Fatal error: Uncaught exception 'Exception' with message 'Object list not found. Possible secured file.'


sergey833:
```
Fatal error: Uncaught exception 'Exception' with message 'Object list not found. Possible secured file.' in ...\src\Smalot\PdfParser\Parser.php:118 Stack trace: 
#0 ...\src\Smalot\PdfParser\Parser.php(97): Smalot\PdfParser\Parser->parseContent('%PDF-1.5\n%\xE2\xE3\xCF\xD3\n...') 
#1 ...\index.php(770): Smalot\PdfParser\Parser->parseFile('......') 
#2 {main} thrown in ...\src\Smalot\PdfParser\Parser.php on line 118
```

PDF loaded from https://elib.vsmu.by/bitstream/123/2275/1/Kabanova-SA_Osnovy%20cheliustno-litsevoj%20khirurgii_T-1_2011.pdf

ppritcha:
We are experiencing this problem also.  It appears to be the same as #392, #373, #356, and #330.  #330 was reported on August 30 of last year.

@k00ni, Is there anything we can do to help get this resolved?

k00ni:
Unfortunately I don't have the time to dive into that, but I will assist any pull request which aims to solve this issue. For what I saw until now, I assume that this library is not able (yet) to parse this kind of files.

ppritcha:
I see now that this issue was first reported in 2014, in the earliest issue that is still open: #30   Also, #209, #192, #185 .  It's been reported about 9 times over the course of 7 years.   It appears to be a problem with the library not handling PDFs that are have a version > 1.4.

k00ni:
Could you please give us feedback if #400 solves this problem?

ppritcha:
That solves the problem for us.

k00ni:
Thank you for your quick response.

Issue #392, opened 2021-02-22
Fatal error: Uncaught Exception: Object list not found. Possible secured file. in


BeycanDeveloper:
https://prnt.sc/10477by

I am getting this error when I try to read a pdf file.

k00ni:
Could you please give us feedback if #400 solves this problem?

BeycanDeveloper:
Hello then, I could not solve my problem, I wrote here but I did not get an early return, so I took care of my work with apache tika and I deleted all the files, so the file that does not work in the pdf parser is not available now, so I cannot test it.

So I'd like to help but I guess I won't be. I also don't know how to use github, sorry I'm just learning.

Issue #373, opened 2020-12-07
"File May Be Secure" Error


jlomeo:
Hi All,
I use PDF Parser to parse court opinions in PDF from http://www.cafc.uscourts.gov/opinions-orders 
Recently (seemingly due to a change in PDF formatting from cafc), I started getting the error that the PDF cannot be parsed because the file may be secure.  After speaking with @smalot, I was told these files should not be getting this error.  One of the files is: http://www.cafc.uscourts.gov/sites/default/files/opinions-orders/19-2293.OPINION.12-4-2020_1696173.pdf

The document is not in fact secure and hopefully we can get it working with PDF parser.  Thanks all for your hard work on this great software.

Jim

roalvesrj:
Hi @jlomeo ,
I share your problem, however, mine is with bank files, until a few months ago the files were generated in version 1.4, however, now they use version 1.5 and apparently PDFParse has some problem. Looking at other problems here, I saw that people use qPDF to "solve" the file problem, where after going through the program it is read normally. Well, I made contact with the people of qPDF, because I wanted to put it in the routine of my script to "correct" and then follow the processing, and I had some feedbacks from the people there:

> Regarding PDFParse not accepting 1.5 files, I can only guess what the issue might be. Perhaps it doesn't support 256-bit encryption, or perhaps it doesn't support object streams, or maybe it just doesn't like the version number.

@k00ni I don't know if the comment from the qPDF staff can help with anything

Issue #356, opened 2020-10-12
Problems with PDF Version: 1.5 (Object list not found. Possible secured file. Parser.php 102)


Sandbird:
I know there are several posts with this problem, but I didn't want to hijack their post, plus i thought i could provide some more data on the problem.
**PDF File**: https://eody.gov.gr/wp-content/uploads/2020/10/covid-gr-daily-report-20201011.pdf **(free to use, [ref](https://github.com/smalot/pdfparser/issues/356#issuecomment-724618850))**
**Error**: Object list not found. Possible secured file. (src\Smalot\PdfParser\Parser.php102)
**Fonts Used**: http://prntscr.com/uxrjhi
**Security details**: http://prntscr.com/uxrjyt
**qpdf check**:
```
C:\qpdf --check daily.pdf
checking daily.pdf
PDF Version: 1.5
File is not encrypted
File is not linearized
No syntax or stream encoding errors found; the file may still contain errors that qpdf cannot detect
```
Parsing the file first with **qpdf** does solves the problem, but that's not really a solution (at least for me since i cant install qpdf on my server, and i have to use the windows version first and then manually upload the file on my linux server).

k00ni:
Thank you for reporting it. Can you reference these posts you mentioned, please?

clicksistema:
It happens for me too and the version of PDF is 1.7
[file](https://www.riodasostras.rj.gov.br/wp-content/uploads/2020/07/encarte-ldo-ed-1197.pdf)

k00ni:
@Sandbird and @clicksistema: Can we use your PDF files in our test environment? They must be free to use and without any obligations.

Sandbird:
Yup.

clicksistema:
yes

PR #400, opened 2021-03-08
Add support for PDF 1.5 Xref stream

### LINKED ISSUE #30: Object list not found. Possible secured file.
When trying to encode my pdf i get the following error 

```
Object list not found. Possible secured file.
```

What does this mean?

If I use pdftotext from the command line, I'm able to output the text of my test-file just fine.

smalot:
should fix #30, #192, #209, #330, #356, #373, #392, #397

see: https://www.yumpu.com/en/document/read/15319213/document-management-portable-document-format-adobe/7
chapter 7.5.8

### DIFF
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -239,6 +239,7 @@ class RawDataParser
         }
         $valid_crs = false;
         $columns = 0;
+        $predictor = null;
         $sarr = $xrefcrs[0][1];
         if (!\is_array($sarr)) {
             $sarr = [];
@@ -310,7 +311,7 @@ class RawDataParser
 
         // decode data
         if ($valid_crs && isset($xrefcrs[1][3][0])) {
-            if (isset($predictor)) {
+            if (!is_null($predictor)) {
                 // number of bytes in a row
                 $rowlen = ($columns + 1);
                 // convert the stream into an array of integers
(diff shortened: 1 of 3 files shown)


src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -310,83 +310,95 @@ protected function decodeXrefStream($pdfData, $startxref, $xref = [])
...
-                        $row_upleft = 0;
-                    } else {
-                        $row_left = $row[($i - 1)];
-                        $row_upleft = $prev_row[($j - 1)];
-                    }
-                    switch ($predictor) {
-                        case 10:  // PNG prediction (on encoding, PNG None on all rows)
-                            $ddata[$k][$j] = $row[$i];
-                            break;
+            if (isset($predictor)) {
```
k00ni, line 313: Is `isset` the right choice here? `$predictor`, when set, is of type `integer` based on:

https://github.com/smalot/pdfparser/blob/6db6fc934575070bcbf2821a656cac71c054deea/src/Smalot/PdfParser/RawData/RawDataParser.php#L291

For better code readability, I propose something like the following instead:

```
if (0 <= $predictor) { 
```

Maybe set `$predictor = null;` at the beginning of function too.

k00ni (commented):
Welcome back @smalot! Good to see you here again.

src/Smalot/PdfParser/RawData/RawDataParser.php
smalot: Fixed

k00ni (approved):

Issue #85, opened 2015-10-23
Missing encoding data for: "MacExpertEncoding"


trapero:
Hello,

I´m trying to parse the metadata from different files, the library works great but in some documents give me this error:

Missing encoding data for: "MacExpertEncoding"

You can access one of the document in:

http://www.statistik.rlp.de/fileadmin/dokumente/berichte/C1013_201500_1j_L.pdf

I get this error in the demo too.

Do you know a solution?

Thank you

garraeth:
I'm getting this same message. 

A couple resources I found, but I'm no PDF dev so it's foreign to me:

https://github.com/rohitpaulk/pdfparser/blob/master/gems/pdf-reader-master/lib/pdf/reader/encodings/mac_expert.txt

https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf
	page: 665

This would be a great addition.

Thanks! Love the project!

garraeth:
Just on a whim, I created a MacExpertEncoding class (using the MacRomanEncoding.php file as a template) and copied the $encoding data from MacRomanEncoding into my new MacExpertEncoding class...and it worked!

Granted, I might have just gotten lucky.

smalot:
Could you create a merge request to contribute ?

adrianbj:
Sorry if I should open a new issue, but I am actually getting this error:

Missing encoding data for: ""

I have worked around it for now by specifying `StandardEncoding` like this:
```
$className = '\\Smalot\\PdfParser\\Encoding\\StandardEncoding';
```

The problem is that `$this->get('BaseEncoding')` returns `false`

Any thoughts on how to fix this properly?

Thanks.

k00ni:
Can you provide us the PDF which is causing this problem? It must be free of charge and without any obligations.

adrianbj:
Hi @k00ni  - thanks for responding so quickly. The issue can be seen with this file.
[healthy-chesapeake-waterways.pdf](https://github.com/smalot/pdfparser/files/5880915/healthy-chesapeake-waterways.pdf)

k00ni:
Is `healthy-chesapeake-waterways.pdf` free of charge and without any obligations?

adrianbj:
Yes - it's from here: https://ian.umces.edu/press/newsletters/publication/1/healthy_chesapeake_waterways_2002-05-01/

[...]

PR #407, opened 2021-04-06
Encoding::__toString complies with PHP specification from now on

k00ni:
Besides an if-else I also added more details in code and text.

Hint came from @igor-krein in https://github.com/smalot/pdfparser/issues/85#issuecomment-808087939:

> Currently, in the Encoding.php, there is a "magic" function __toString() implementation, which calls a getEncodingClass() function, which, in its turn, may throw an exception (line 150).

> **From php.net:** *Warning It was not possible to throw an exception from within a __toString() method prior to PHP 7.4.0. Doing so will result in a fatal error.*

> I think, if there is a need to keep a PHP 7.4 support, better change the line 150 somehow (for example, simply return an empty string). Note that __toString() method must return a string.

### DIFF
--- a/src/Smalot/PdfParser/Encoding.php
+++ b/src/Smalot/PdfParser/Encoding.php
@@ -128,9 +128,9 @@ class Encoding extends PDFObject
     /**
      * Returns the name of the encoding class, if available.
      *
-     * @return string
+     * @return string Returns encoding class name if available or empty string (only prior PHP 7.4).
      *
-     * @throws \Exception On PHP 7.4 an exception is thrown if encoding class doesn't exist.
+     * @throws \Exception On PHP 7.4+ an exception is thrown if encoding class doesn't exist.
      */
     public function __toString()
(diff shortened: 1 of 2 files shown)


src/Smalot/PdfParser/Encoding.php
```diff
@@ -126,13 +126,24 @@ public function translateChar($dec)
...
-        return $this->getEncodingClass();
+        try {
+            return $this->getEncodingClass();
+        } catch (Exception $e) {
+            // prior to PHP 7.4 toString has to return an empty string.
+            if (version_compare(\PHP_VERSION, '7.4.0', '<')) {
+                return '';
+            } else {
+                throw $e;
+            }
```
igor-krein, line 145: Optional tiny simplification (but it is actually a coding style matter):

```suggestion
            }
            throw $e;
```

j0k3r (approved):
👍

PR #421, opened 2021-05-14
Fix decode octal regex

gdiasb12:


k00ni (commented):
Thank you for your contribution!

What is the matter with #422? Is this one you main PR?

tests/Integration/FontTest.php
```diff
@@ -277,6 +277,7 @@ public function testDecodeOctal()
     {
         $this->assertEquals('AB C', Font::decodeOctal('\\101\\102\\040\\103'));
         $this->assertEquals('AB CD', Font::decodeOctal('\\101\\102\\040\\103D'));
+        $this->assertEquals('AB \199', Font::decodeOctal('\\101\\102\\040\\199'));
```
k00ni, line 280: Can you explain your test extension here a bit please?

gdiasb12: This test will guarantee that non-octal numbers like **199**  will not be converted. That's because after PHP 7.4, if you try to pass a non-octal number  to the function [octdec](https://www.php.net/manual/en/function.octdec.php) it's  the  generate a deprecation notice. So, basically this line guarantee that the octal numbers _\\101\\102\\040_ will be converted properly and the last one _\\199_ will not.

gdiasb12:
> Thank you for your contribution!
> 
> What is the matter with #422? Is this one (#421) your main pull request?

Sorry about the mess. lol

Yes, the (#421) is the main one.

tests/Integration/FontTest.php
k00ni: Thank you.

k00ni (approved):
Good to go. If there aren't any objections I will merge it soon.

PR #424, opened 2021-05-18
Fixed remaining parameter lists which miss Config parameter

### LINKED ISSUE #423: (Another) Call to a member function getFontSpaceLimit() on null
@StudPro I opened a new issue so we can check and handle the error you mentioned [here](https://github.com/smalot/pdfparser/pull/403#issuecomment-842102409).

Error: 

> Call to a member function getFontSpaceLimit() on null

PDF which causes the problem: https://github.com/smalot/pdfparser/files/6492106/Martucci.et.al.pdf

Because I introduced this faulty code (Config class) I will prepare a PR with a fix.

k00ni:
This PR should fix the following error message once and for all:

> Call to a member function getFontSpaceLimit() on null

Closes #423

I tested it with https://github.com/smalot/pdfparser/files/6492106/Martucci.et.al.pdf successfully. Mentioned PDF seems to be a copyright protected paper, therefore it was not added to our tests.

### DIFF
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -211,7 +211,7 @@ class Page extends PDFObject
                     }
 
                     $header = new Header([], $this->document);
-                    $contents = new PDFObject($this->document, $header, $new_content);
+                    $contents = new PDFObject($this->document, $header, $new_content, $this->config);
                 }
             } elseif ($contents instanceof ElementArray) {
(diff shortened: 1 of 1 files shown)


k00ni:
@StudPro does this PR fix your problems?

j0k3r (approved):
Looks ok

StudPro:
> 
> 
> @StudPro does this PR fix your problems?

Yes, thank's a lot

PR #433, opened 2021-06-08
Don't throw an exception if Encoding doesn't have BaseEncoding

### LINKED ISSUE #432: "BaseEncoding" should be optional
In `Encoding->getEncodingClass()`, if the field `BaseEncoding` doesn't exist, it throws an Exception but, accordingly PDF 1.5 Reference Table 5.11 (attached), the field `BaseEncoding` is optional, so, if any PDF builder don't put this field, pdfparser will crash.

image

LucianoHanna:
Fixes #432 

`$this->get('Encoding')->__toString()` throws an exception when Encoding don't have BaseEncoding attribute, so I put it on a try catch because if Encoding don't have BaseEncoding, of course `WinAnsiEncoding::class === $this->get('Encoding')->__toString()` will return false

I didn't make any unit test because I only have a confidential PDF file which cause this bug

### DIFF
--- a/src/Smalot/PdfParser/EncodingNotFoundException.php
+++ b/src/Smalot/PdfParser/EncodingNotFoundException.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Smalot\PdfParser;
+
+class EncodingNotFoundException extends \Exception
+{
+}
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -113,7 +113,7 @@ class Font extends PDFObject
                     $fallbackDecoded = self::uchr($dec);
                 }
             } catch (EncodingNotFoundException $e) {
-                // Encoding->getEncodingClass() throws an exception when BaseEncoding doesn't exists
+                // Encoding->getEncodingClass() throws EncodingNotFoundException when BaseEncoding doesn't exists
                 // See table 5.11 on PDF 1.5 specs for more info
             }
         }
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -107,9 +107,15 @@ class Font extends PDFObject
             \strlen($char) < 2
             && $this->has('Encoding')
             && $this->get('Encoding') instanceof Encoding
-            && WinAnsiEncoding::class === $this->get('Encoding')->__toString()
         ) {
-            $fallbackDecoded = self::uchr($dec);
+            try {
+                if (WinAnsiEncoding::class === $this->get('Encoding')->__toString()) {
+                    $fallbackDecoded = self::uchr($dec);
+                }
+            } catch (\Exception $e) {
+                // Encoding->getEncodingClass() throws an exception when BaseEncoding doesn't exists
+                // See table 5.11 on PDF 1.5 specs for more info
+            }
         }
 
         return $use_default ? self::MISSING : $fallbackDecoded;
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -112,7 +112,7 @@ class Font extends PDFObject
                 if (WinAnsiEncoding::class === $this->get('Encoding')->__toString()) {
                     $fallbackDecoded = self::uchr($dec);
                 }
-            } catch (\Exception $e) {
+            } catch (EncodingNotFoundException $e) {
                 // Encoding->getEncodingClass() throws an exception when BaseEncoding doesn't exists
                 // See table 5.11 on PDF 1.5 specs for more info
             }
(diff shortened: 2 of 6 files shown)


src/Smalot/PdfParser/Font.php
```diff
@@ -107,9 +107,15 @@ public function translateChar($char, $use_default = true)
...
         ) {
-            $fallbackDecoded = self::uchr($dec);
+            try {
+                if (WinAnsiEncoding::class === $this->get('Encoding')->__toString()) {
+                    $fallbackDecoded = self::uchr($dec);
+                }
+            } catch (\Exception $e) {
+                // Encoding->getEncodingClass() throws an exception when BaseEncoding doesn't exists
+                // See table 5.11 on PDF 1.5 specs for more info
+            }
```
k00ni, line 118: I get what you want to achieve here but a "catch-all" at this point might hide other viable exceptions. 

What do you think about adapting `getEncodingClass` to throw a specific exception like `EncodingClassNotFoundException`. We could catch that and ignore the rest at this point.

k00ni (changes requested):
@LucianoHanna thank you for the PR!

src/Smalot/PdfParser/Font.php
LucianoHanna: I agree, a "catch-all" can create a new bug if any other Exception was raised.

LucianoHanna:
Hi,

I tried to create a pdf without sensitive information but I did not make it...
So I created a test only barely reproducing the state that would cause this error.

Test with a real pdf would be a lot better... I appreciate a lot if someone could help with it!

src/Smalot/PdfParser/Font.php
```diff
@@ -107,9 +107,15 @@ public function translateChar($char, $use_default = true)
...
             && $this->get('Encoding') instanceof Encoding
-            && WinAnsiEncoding::class === $this->get('Encoding')->__toString()
         ) {
-            $fallbackDecoded = self::uchr($dec);
+            try {
+                if (WinAnsiEncoding::class === $this->get('Encoding')->__toString()) {
+                    $fallbackDecoded = self::uchr($dec);
+                }
+            } catch (EncodingNotFoundException $e) {
+                // Encoding->getEncodingClass() throws an exception when BaseEncoding doesn't exists
```
k00ni, line 116:
```suggestion
                // Encoding->getEncodingClass() throws EncodingNotFoundException when BaseEncoding doesn't exists
```

src/Smalot/PdfParser/EncodingNotFoundException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Smalot\PdfParser;
+
+class EncodingNotFoundException extends \Exception
+{
+}
```
k00ni, line 7: Please move to `src/Smalot/PdfParser/Exception`.

tests/Integration/FontTest.php
```diff
@@ -379,4 +379,17 @@ public function testXmlContent()
 
         $this->assertEquals('Example PDF', $text);
     }
+
+    /**
+     * Create a header containing a Encoding that doesn't have a BaseEncoding
+     * so I test if the Font won't raise a exception because Encoding don't have BaseEncoding
```
k00ni, line 385:
```suggestion
     * Create an instance of Header containing an instance of Encoding that doesn't have a BaseEncoding.
     * Test if the Font won't raise a exception because Encoding don't have BaseEncoding.
```

k00ni (changes requested):
Nicely done!

About tests: There are different types, like integration and functional tests. Using a PDF is not the best way, because it involves a lot of code. Though, if we have PDFs available to exploit existing problems, they should be part of the tests. At this point if prefer integration tests like you did with `testEncodingWithoutBaseEncoding`. 👍🏿

LucianoHanna:
> Nicely done!
> 
> About tests: There are different types, like integration and functional tests. Using a PDF is not the best way, because it involves a lot of code. Though, if we have PDFs available to exploit existing problems, they should be part of the test suite. At this point I prefer integration tests like you did with `testEncodingWithoutBaseEncoding`. 👍🏿

Thank you for your feedback!

k00ni (approved):
Looks good to me. Will leave this open for a few days.

@j0k3r @izabala any remarks?

j0k3r (approved):
Looks good!

LucianoHanna:
Any news about this merge and new release?

k00ni:
> Any news about this merge and new release?

For me its important to give the community a time window to comment/test new pull requests. Therefore I wrote that I will leave this open for a few days. 

Do you need it asap?

LucianoHanna:
> > Any news about this merge and new release?
> 
> For me its important to give the community a time window to comment/test new pull requests. Therefore I wrote that I will leave this open for a few days.
> 
> Do you need it asap?

Sorry, I thought you would leave this open for a few days until have another approval.

It's better to give the community a time window.

Thank you

Issue #434, opened 2021-06-13
Document's object dictionary not used when asking for default font


jee7:
From v0.17.0 a `PDFObject::getDefaultFont()` was asked in the beginning of `PDFObject::getText()` method. The `getDefaultFont()` asks for all the fonts inside a document and the first one is picked. The `Document::getFonts()` method uses `Document::getObjectsByType()`, which is uncached. Meaning it does not use the dictionary built inside the `Document` class, but rather loops through *all* the objects again, checking their type to find the fonts.

For PDF files with a lot of objects, this causes *significant* overhead. For example with the [PDF file from this page](https://comserv.cs.ut.ee/ati_thesis/datasheet.php?id=70198&year=2020&language=en). I apologize for not having a custom-made PDF here. With that PDF there is a drastic difference when parsing pages 77 and 78. With the v0.16.2 version both pages take < 0.1 seconds to parse. With versions starting from 0.17.0 (including the current latest version) they take ~300 seconds together.

![image](https://user-images.githubusercontent.com/3455815/121791624-a7a50280-cbf4-11eb-8b29-9d9ea7ccf639.png)

Test code:
```
<?php

include('vendor/autoload.php');

set_time_limit(720);

$parser = new \Smalot\PdfParser\Parser();

// load PDF file content
$data = file_get_contents('https://comserv.cs.ut.ee/home/files/Shoush_ComputerScience_2020.pdf?study=ATILoputoo&reference=76F6FAFD4C9E6981D9A434D32D2E7EE2AE9C49E7');


// give PDF content to function and parse it
$pdf = $parser->parseContent($data); 

$pages = $pdf->getPages();

foreach ($pages as $i => $page) { /** @var $page Page */
	if ($i < 77) continue;
	if ($i > 78) continue;

	$startTime = microtime(true);
	$pageText = $page->getText();
	$endTime = microtime(true);
	
	echo '<b>Page ' . $i . ' (took ' . ($endTime - $startTime) . ' seconds, ' . round(memory_get_usage() / (1000 * 1000), 2) . ' MB RAM)</b><br>';
	var_dump($pageText);
	echo '<br><br>';
}

?>
```

When getting the text for all the pages (not just 77, 78), then the script doesn't seem to complete (Apache crashed at one point). The main issues here seems to be that the dictionary cache is not used and thus this huge number of calls to the `Header::get()` and other functions create an overhead. This is even more unnecessary as only the first font is actually required.

PR coming up.

[...]

PR #435, opened 2021-06-13
Fix for #434. Reworked the Document's object cache dictionary. 

jee7:
The getObjectsByType() method now uses it correctly. The dictionary also should support subtype searches. Only one font is asked for and returned to get the default font.

This solution parses the 2 pages of the big PDF document linked in #434 in about the same time as v0.16.2 did (<0.1 sec per page). On my machine, it's 0.005 s for page 77 and 0.113 s for page 78 (comparable numbers to v0.16.2 speed).


src/Smalot/PdfParser/Document.php
```diff
@@ -91,10 +91,26 @@ protected function buildDictionary()
         $this->dictionary = [];
 
         foreach ($this->objects as $id => $object) {
+            // Cache objects by type and subtype
             $type = $object->getHeader()->get('Type')->getContent();
 
-            if (!empty($type)) {
-                $this->dictionary[$type][$id] = $id;
+            if (null != $type) {
```
k00ni, line 97: Why changing from `!empty` to `null !=` here?

src/Smalot/PdfParser/Document.php
```diff
@@ -169,6 +185,11 @@ public function getObjectById($id)
         return null;
     }
 
+    public function hasObjectsByType($type, $subtype = null)
```
k00ni, line 188: Since its a new function, please make it more explicit about what it expects and returns. Please check my suggestion and adapt accordingly.

```suggestion
    public function hasObjectsByType(string $type, ?string $subtype = null): bool
```

src/Smalot/PdfParser/Document.php
```diff
@@ -198,25 +221,36 @@ public function getFonts()
         return $this->getObjectsByType('Font');
     }
 
+    /**
+     * @return PDFObject
+     */
+    public function getFirstFont()
```
k00ni, line 227: What if there are no fonts? Also add return value type.

```suggestion
    public function getFirstFont(): PDFObject
```

src/Smalot/PdfParser/Document.php
```diff
@@ -198,25 +221,36 @@ public function getFonts()
...
-        if (isset($this->dictionary['Catalog'])) {
+        if ($this->hasObjectsByType('Catalog')) {
             // Search for catalog to list pages.
-            $id = reset($this->dictionary['Catalog']);
+            $catalogues = $this->getObjectsByType('Catalog');
+            $catalogue = reset($catalogues);
 
             /** @var Pages $object */
-            $object = $this->objects[$id]->get('Pages');
+            $object = $catalogue->get('Pages');
```
k00ni, line 247: `getObjectsByType` can return an empty array, therefore `$catalogue` is null and `$catalogue->get('Pages');` leads to an error, doesn't it? If so, please add a check up front.

k00ni (changes requested):
@jee7 thanks for your pull request! 👍🏿

Besides my comments below, is there a way to add at least 1 test which check changes or outlines performance boost? Maybe based on your test code from [here](https://github.com/smalot/pdfparser/issues/434#issue-919686188).

k00ni:
Hey @jee7 how is it going?

jee7:
> Hey @jee7 how is it going?

Apologies, I did not get any notifications from your comments before.

* The type declarations are added as per recommendations. Although, if this will be the new convention for this project, then it should be done everywhere, right? Currently, these two methods are now different from the rest of the code because of the PHP type declarations.
* I cannot come up with a way to implement the performance test in a good manner. I'm new to unit tests and as far as I understood they should not be used for measuring performance. Your suggestion?

src/Smalot/PdfParser/Document.php
jee7: That is what the
```if ($this->hasObjectsByType('Catalog')) {```
checks beforehand, yes.

If we have objects by type `'Catalog'`, then we proceed to get these objects and do not worry if there aren't any (because we know there are). So there should always be at least one `$catalogue` there.

File: src/Smalot/PdfParser/Document.php (hunk 1 above)
Comment (jee7) [reply src/Smalot/PdfParser/Document.php:97]: Because the variable is declared here, thus `empty()` will not give anything more here than a null check. I suspect even a typed null check might work (or ideally the code should be made such). 

Using methods that check for more than necessary makes the code more confusing for others to read. Because it makes the reader think that all the functionality is necessary (the variable could be undeclared or the value equal to false/null -- and both checks are necessary for some reason when in reality they aren't).

**Comment by k00ni** [comment]
> * The type declarations are added as per recommendations. Although, if this will be the new convention for this project, then it should be done everywhere, right? Currently, these two methods are now different from the rest of the code because of the PHP type declarations.

No, in my opinion we should be as explicit about new things as possible (e.g. add return values etc.), but that doesn't mean we have to adapt the old code too. It can be updated later on during another PR or some refactoring. This way we have better code quality and understanding and don't need to rely on PHP doc.

> * I cannot come up with a way to implement the performance test in a good manner. I'm new to unit tests and as far as I understood they should not be used for measuring performance. Your suggestion?

Another Github Action could be used which only contains performance tests. You can copy `.github/workflows/continuous-integration.yml` to `performance-tests.yml` (or something better than that) and write 1, 2 simple tests. To get this thing going it would be very cool if you could replicate the thing you talked about initialy:

> This solution parses the 2 pages of the big PDF document linked in #434 in about the same time as v0.16.2 did (<0.1 sec per page). On my machine, it's 0.005 s for page 77 and 0.113 s for page 78 (comparable numbers to v0.16.2 speed).

So you performance test should load a bunch of PDFs and fails if they need longer than x seconds to parse, for instance.

**Comment by jee7** [comment]
Okay, how about now?

File: tests/Performance/runPerformanceTests.php [hunk 5]
```diff
@@ -0,0 +1,26 @@
+<?php
+
+require __DIR__.'/../../alt_autoload.php-dist';
+
+// FIXME: Autoload these.
+require_once __DIR__.'/Exception/PerformanceFailException.php';
+require_once __DIR__.'/Test/AbstractPerformanceTest.php';
+require_once __DIR__.'/Test/DocumentDictionaryCacheTest.php';
```
k00ni, line 8: Autoloading for test related classes is already possible (see [this](https://github.com/smalot/pdfparser/blob/master/composer.json#L27-L31)). Please use `vendor/autoload.php` instead.

```suggestion
require __DIR__.'/../../vendor/autoload.php';
```

.github/workflows/performance.yml
```diff
@@ -0,0 +1,27 @@
...
+    strategy:
+      matrix:
+        php:
+          - "7.4"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+
+      - name: "Test dictionary cache"
```
k00ni, line 26: Please change name, because related file may run multiple tests in the future. `Run performance tests` or something like that might be a good replacement.

.github/workflows/performance.yml
```diff
@@ -0,0 +1,27 @@
+name: "Performance"
```
k00ni, line 1: Rename workflow file to `performance-tests.yml`, so its more clear what it is about.

```suggestion
name: "Performance Tests"
```

.github/workflows/performance.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  performance-tests:
+    name: "Tests for the performance testing the PDF parsing"
+    runs-on: "ubuntu-20.04"
+
+    strategy:
+      matrix:
+        php:
+          - "7.4"
```
k00ni, line 20: In your opinion, is it worth testing performance with all supported PHP versions? If so, please add 7.2, 7..3 and 8.0 here.

src/Smalot/PdfParser/Document.php
```diff
@@ -169,6 +185,14 @@ public function getObjectById($id)
         return null;
     }
 
+    /**
+     * @param ?string $subtype
```
k00ni, line 189: Please remove, because its a duplicate to the PHP code.

tests/Performance/Test/AbstractPerformanceTest.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Tests\Smalot\PdfParser\Performance\Test;
+
+abstract class AbstractPerformanceTest
+{
+    /**
+     * Initializes the test (eg, fetches the files etc).
+     *
+     * @return void
+     */
+    abstract public function init();
```
k00ni, line 12:
```suggestion
     */
    abstract public function init(): void;
```

tests/Performance/Test/AbstractPerformanceTest.php
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Executes the test.
+     *
+     * @return void
+     */
+    abstract public function run();
```
k00ni, line 19:
```suggestion
     */
    abstract public function run(): void;
```

tests/Performance/Test/AbstractPerformanceTest.php
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Returns the time over which the test is considered a fail.
+     *
+     * @return int
+     */
+    abstract public function getMaxEstimatedTime();
```
k00ni, line 26:
```suggestion
     */
    abstract public function getMaxEstimatedTime(): int;
```

tests/Performance/Test/DocumentDictionaryCacheTest.php
```diff
@@ -0,0 +1,82 @@
...
+{
+    /**
+     * @var Parser
+     */
+    protected $parser;
+    protected $data;
+
+    public function init()
+    {
+        $this->parser = new \Smalot\PdfParser\Parser();
```
k00ni, line 53: `Parser` class already included via use above, therefore no need to get full path.

```suggestion
        $this->parser = new Parser();
```

tests/Performance/Test/DocumentDictionaryCacheTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+        // load PDF file content
+        $this->data = file_get_contents('https://comserv.cs.ut.ee/home/files/Shoush_ComputerScience_2020.pdf?study=ATILoputoo&reference=76F6FAFD4C9E6981D9A434D32D2E7EE2AE9C49E7');
```
k00ni, line 56: I don't like dependencies from other websites, because if it goes offline, our tests break. Please copy the file to `samples` instead.

Furthermore, file must be free of charge and without any obligations. Are you `Mahmoud Kamel Shoush`?

tests/Performance/runPerformanceTests.php
```diff
@@ -0,0 +1,26 @@
...
+    $test->init();
+
+    $startTime = microtime(true);
+    $test->run();
+    $endTime = microtime(true);
+
+    $time = $endTime - $startTime;
+
+    if ($test->getMaxEstimatedTime() < $time) {
+        throw new \Tests\Smalot\PdfParser\Performance\Exception\PerformanceFailException(sprintf('Performance failed on test "%s". Time taken was %.2f seconds, expected less than %d seconds.', get_class($test), $time, $test->getMaxEstimatedTime()));
```
k00ni, line 24: Just a minor thing, why not `<=`?

k00ni (changes requested):
Thank you for your changes. Looks good, we are getting there!

tests/Performance/Test/DocumentDictionaryCacheTest.php
jee7: I am not. I'm an employee of the university this thesis document is from (I'm using the pdfparser to parse all the thesis documents). The PDF includes a license for the university to distribute this file. However, this does not extend to other third parties. So I don't think we can put this to the repo.

.github/workflows/performance.yml
jee7: As the issue was currently with the code, not with the PHP version, then I don't think running it with different PHP versions helps a lot right now (rather it will make the testing unnecessarily slower).

jee7:
The Type Declaration should be now refactored to be better throughout. Other smaller fixes were done as well. Autoloading does not work. I asked the author of the PDF file if we can also host it. I'll let you know what he responds.

tests/Performance/runPerformanceTests.php
k00ni: You need to run `composer update` before you start your tests.

.github/workflows/performance.yml
k00ni: Fair point.

tests/Performance/Test/DocumentDictionaryCacheTest.php
k00ni: In this case we need to find another solution, because I am inclined to not accept PDF via links (as described before). Is there a way to create an artificial PDF which also works here?

k00ni (changes requested):
Thank you for your adaptions.

But why did you refactored all the other files too? It bloats your pull request. I checked my statement:

 > in my opinion we should be as explicit about new things as possible (e.g. add return values etc.), **but that doesn't mean we have to adapt the old code too.** It can be updated later on during another PR or some refactoring.

Please revert these changes so your pull request only contains optimizations towards a better object cache dictionary.

jee7:
Autoloading with `composer update` ran before works, thanks. The author of the thesis e-mailed me back with the approval of using his thesis PDF in this repo. Added it to the `samples` folder. You told me to remove the PHPDoc in favor of Type Declarations. It only makes sense to then unify the entire project to use the same convention. I find it bothersome to work with code, which mixes different conventions together. If we want new code to use Type Declarations instead of PHPDoc, then the rest of the project must follow suit before this.

k00ni:
> I find it bothersome to work with code, which mixes different conventions together. If we want new code to use Type Declarations instead of PHPDoc, then the rest of the project must follow suit before this.

I understand your point. My suggestion to solve this issue would be that you move these changes into a separate branch and make another pull request. OK for you?

jee7:
Okay. PR #440 now includes the PHPDoc / Type Declaration changes.

k00ni:
Is this PR now dependent on #440?

jee7:
> Is this PR now dependent on #440?

That would seem efficient, yes. When that gets approved, I can pull and merge it back into this PR. This will hopefully save a few hours of trying to put all the PHPDocs back here the way they were.

k00ni:
@jee7 can you check if #441 affects your performance test?

jee7:
> @jee7 can you check if #441 affects your performance test?

Seems not to affect it to me.

k00ni:
@jee7 your PR with refactored code was merged. Can you solve merge conflicts so we can continue here?

jee7:
> @jee7 your PR with refactored code was merged. Can you solve merge conflicts so we can continue here?

Should be good now.

k00ni (approved):
👍🏿 Thank you, looks good!

I will keep this open for around week then I will merge.

PR #440, opened 2021-07-14
Code Refactoring

jee7:
Removed PHPDoc where possible in favor of Type Declarations.

Refactored the loading of Encoding classes for better typing.

Other refactorings like removing unused or immediately overwritten variables etc.


src/Smalot/PdfParser/Element/ElementDate.php
```diff
@@ -59,29 +59,25 @@ class ElementDate extends ElementString
...
-     * @param \DateTime $value
+     * @var \DateTime
      */
+    protected $value;
+
     public function __construct($value)
     {
         if (!($value instanceof \DateTime)) {
-            throw new \Exception('DateTime required.');
+            throw new \Exception('DateTime required.'); // FIXME: Sometimes strings are passed to this function
```
k00ni, line 69: Can you explain what you mean with the `FIXME` here?

src/Smalot/PdfParser/Parser.php
```diff
@@ -58,28 +58,18 @@ class Parser
...
     {
         return $this->config;
     }
 
-    /**
-     * @param string $filename
-     *
-     * @return Document
-     *
-     * @throws \Exception
```
k00ni, line 80: Please keep this line.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -779,16 +781,15 @@ protected function getRawObject($pdfData, $offset = 0)
     /**
      * Get Cross-Reference (xref) table and trailer data from PDF document data.
      *
-     * @param string $pdfData
-     * @param int    $offset  xref offset (if know)
-     * @param array  $xref    previous xref array (if any)
+     * @param int   $offset xref offset (if know)
```
k00ni, line 784:
```suggestion
     * @param int   $offset xref offset (if known)
```

k00ni (changes requested):
Thank you for your work! I took a quick look and found only a few things here and there.

Making function parameters and return values explicit shouldn't break API compatibility IMHO. Any opinions on that? As always, non-collaborator feedback is welcomed too!

CC @j0k3r @smalot

src/Smalot/PdfParser/Element.php
```diff
@@ -98,24 +89,12 @@ public function getContent()
...
     }
 
-    /**
-     * @param string   $content
-     * @param Document $document
-     * @param int      $position
-     *
-     * @return array
-     *
-     * @throws \Exception
```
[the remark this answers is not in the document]
jee7: The reason I removed it was that I did not see any Exceptions thrown in the code. Where is that exception coming from?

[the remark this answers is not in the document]
k00ni: Because it may get overridden by a subclass, but you are right, overriding function only requires the `@throws` PHP doc.

src/Smalot/PdfParser/Element/ElementDate.php
jee7: Yes. I could not type the input parameter `$value` there to be of type `DateTime` as would be logical, because it seemed sometimes strings were passed in there. I did not dig too much into it, but for example, in `ElementDateTest` there is a test method `testGetContentInvalidParameter`, which does exactly that. When the argument is typed, there is an error instead of an exception in such cases.
Seems that this logic should be refactored somehow. That we should always pass a `DateTime` object into the constructor. If strings are needed, there can be some `createFromString` static method available. An exception for checking the input parameter type does not seem good, but I don't know enough about how it is actually used to change it right now.

src/Smalot/PdfParser/Parser.php
k00ni: Same as before, only keep if function itself can throw an exception.

src/Smalot/PdfParser/Element/ElementArray.php
```diff
@@ -58,20 +54,12 @@ public function getContent()
         return parent::getContent();
     }
 
-    /**
-     * @return array
-     */
     public function getRawContent()
```
j0k3r, line 57: Did you try?

```suggestion
    public function getRawContent(): array
```

src/Smalot/PdfParser/Element/ElementString.php
```diff
@@ -39,30 +39,20 @@
  */
 class ElementString extends Element
 {
-    /**
-     * @param string $value
-     */
     public function __construct($value)
```
j0k3r, line 42:
```suggestion
    public function __construct(string $value)
```

src/Smalot/PdfParser/Font.php
```diff
@@ -417,13 +364,7 @@ public function decodeText($commands)
         return implode(' ', $words);
     }
 
-    /**
-     * @param string $text
-     * @param bool   $unicode This parameter is deprecated and might be removed in a future release
```
j0k3r, line 422: Maybe we should keep that comment?

src/Smalot/PdfParser/Parser.php
jee7: I changed this. The called `parseContent` method threw an exception and there was a big comment there (I missed it before), which said that users of the library want the `@throws \Exception` PHPDoc on this particular function.

src/Smalot/PdfParser/Header.php
```diff
@@ -176,14 +161,7 @@ protected function resolveXRef($name)
         return $this->elements[$name];
     }
 
-    /**
-     * @param string   $content  The content to parse
-     * @param Document $document The document
-     * @param int      $position The new position of the cursor after parsing
```
j0k3r, line 182: Shouldn't this comment be kept?

src/Smalot/PdfParser/Element/ElementString.php
jee7: This one does not work. Sometimes objects are passed in there. The `ElementDate` extends `ElementString` and in the constructor calls the parent constructor with the same argument as it gets. Thus `ElementString` sometimes gets a `DateTime` object (and not string) as the `$value` parameter in the constructor.

src/Smalot/PdfParser/Element/ElementArray.php
jee7: Seemed to work, yes.

src/Smalot/PdfParser/Header.php
jee7: Must have been too eager in removing these. Re-added them.

src/Smalot/PdfParser/Header.php
```diff
@@ -126,22 +119,16 @@ public function getDetails($deep = true)
 
     /**
      * Indicate if an element name is available in header.
```
k00ni, line 121:
```suggestion
    /**
     * Indicate if an element name is available in header.
     *
     * @param string $name The name of the element.
```

src/Smalot/PdfParser/Pages.php
```diff
@@ -36,14 +36,12 @@
 class Pages extends PDFObject
 {
     /**
-     * @param bool $deep
-     *
      * @todo Objects other than Pages or Page might need to be treated specifically in order to get Page objects out of them,
      * see https://github.com/smalot/pdfparser/issues/331
```
k00ni, line 40:
```suggestion
     * @see https://github.com/smalot/pdfparser/issues/331
```

src/Smalot/PdfParser/Pages.php
```diff
@@ -36,14 +36,12 @@
...
      *
-     * @return array
+     * @return array|string
```
k00ni, line 42: Why did you add `string` here?

src/Smalot/PdfParser/Parser.php
```diff
@@ -96,14 +89,10 @@ public function parseFile($filename)
     }
 
     /**
-     * @param string $content PDF content to parse
-     *
-     * @return Document
-     *
      * @throws \Exception if secured PDF file was detected
```
k00ni, line 92:
```suggestion
     * @param string $content PDF content to parse
     *
     * @throws \Exception if secured PDF file was detected
```

src/Smalot/PdfParser/Element.php
```diff
@@ -98,24 +89,12 @@ public function getContent()
...
-     */
-    public static function parse($content, Document $document = null, &$position = 0)
+    public static function parse(string $content, ?Document $document = null, int &$position = 0)
```
k00ni, line 97: To be consistent with other return-type changes.

```suggestion
    public static function parse(string $content, ?Document $document = null, int &$position = 0): array
```

k00ni (changes requested):
Almost done, just a few things.

src/Smalot/PdfParser/Element.php
jee7: The return type `array` cannot be added as the extended class `ElementString` does not return an array from that function.

k00ni:
After updating your code please resolve related open conversation.

src/Smalot/PdfParser/Pages.php
jee7: I misunderstood what object is being fetched with the `->get` call. Added PHPDoc to make it more clear and refactored the return type.

jee7:
> After updating your code please resolve related open conversation.

Okay. I thought you'd want to do it yourself to check if all the things were fixed accordingly.

k00ni:
We had problems with Scrutinizer lately, but #443 fixes them. After it is merged, I can take care of this PR next.

k00ni:
Can you merge latest master branch in to trigger CI? I wanna see if scrutinizer fails again.

k00ni (approved):
Thanks for your work and patience @jee7! Looks good, will leave it open for a few days so everyone has a chance to comment.

jee7:
> Thanks for your work and patience @jee7! Looks good, will leave it open for a few days so everyone has a chance to comment.

Thank you for maintaining the repo and for the code review feedback. In the future, the logic in the lib should be also refactored at certain places if we want to go with type declared inputs and outputs. Especially at the places that currently get differently typed inputs or produce differently typed outputs.

Also, something to think about is that if we use PHPDoc for describing some (but not all) of the input parameters or output, then the PHPStorm IDE for example tells me that the PHPDoc is incomplete.

I think some guidelines for how to format the code could also be written in DEVELOPER.md. So that next PR-s know to use type declaration instead of PHPDoc, or when exactly and how PHPDoc should be used in this project etc.

k00ni:
> Thank you for maintaining the repo and for the code review feedback. In the future, the logic in the lib should be also refactored at certain places if we want to go with type declared inputs and outputs. Especially at the places that currently get differently typed inputs or produce differently typed outputs.

As one of the maintainers I personally focus on managing issues and pull requests, only some coding here and there. I want to enable people to contribute to the library, but it is always a trade off between refactoring and good, readable code vs. API compatibility. Going the slow, but steady way seems to be a good middle ground.

> Also, something to think about is that if we use PHPDoc for describing some (but not all) of the input parameters or output, then the PHPStorm IDE for example tells me that the PHPDoc is incomplete.

If I remember correctly its because of incomplete paths to the classes. In my opinion, its sufficient for now to just outline the class name, so developers can see and look for themselves. But it should be made more reliable in the future.

> I think some guidelines for how to format the code could also be written in DEVELOPER.md. So that next PR-s know to use type declaration instead of PHPDoc, or when exactly and how PHPDoc should be used in this project etc.

Good remark. Feel free to make a suggestion.

src/Smalot/PdfParser/Font.php
```diff
@@ -122,22 +110,14 @@ public function translateChar($char, $use_default = true)
         return $use_default ? self::MISSING : $fallbackDecoded;
     }
 
-    /**
-     * @param int $code
-     *
-     * @return string
-     */
-    public static function uchr($code)
+    public static function uchr(int $code): string
```
jesse-greathouse, line 113: This change seems to have caused a problem of when the variable `$code` is not an int but rather a float.

```bash
 Smalot\PdfParser\Font::uchr(): Argument #1 ($code) must be of type int, float given, called in /home/jessegreathouse/dcol/src/vendor/smalot/pdfparser/src/Smalot/PdfParser/Font.php on line 223

  at vendor/smalot/pdfparser/src/Smalot/PdfParser/Font.php:138
    134▕
    135▕     /**
    136▕      * Convert unicode character code to "utf-8" encoded string.
    137▕      */
  ➜ 138▕     public static function uchr(int $code): string
    139▕     {
    140▕         if (!isset(self::$uchrCache[$code])) {
    141▕             // html_entity_decode() will not work with UTF-16 or UTF-32 char entities,
    142▕             // therefore, we use mb_convert_encoding() instead
```

This was most likely an unforeseen consequence of an innocent change.

Here is the pdf document for your convenience:
[15-1039-1.pdf](https://github.com/smalot/pdfparser/files/12210039/15-1039-1.pdf)

```php
<?php
use Smalot\PdfParser\Parser;
$parser = new Parser();
$pdf = $this->parser->parseFile('15-1039-1.pdf');
?>
```

PR #441, opened 2021-07-21
implemented config setting $retainImageContent

Connum:
Setting this to false should drastically reduce memory usage, allowing to parse documents containing graphics that would otherwise cause the memory limit to be reached, resulting in a fatal error.

This addresses part of the problems discussed in #104, but does not fix the actual memory leak, so that issue has to stay open.


Connum:
It seems that `gc_collect_cycles()` doesn't do its job reliably with PHP < 7.3 ... At least that's my only explanation for why the memory usage test fails for those two versions and works for all others... Howe shall we handle this?

k00ni:
We can just skip the test on PHP version lower 7.3 (using `markTestSkipped`). The functionality is there and works, but we can't compensate for a faulty PHP function.

k00ni (approved):
👍🏿

I will leave this PR open for a few days to give the community the chance to react and comment.

k00ni:
@Connum: latest Scrutinizer run crashed, can you please have a look?

https://scrutinizer-ci.com/g/smalot/pdfparser/inspections/2db87eb0-536f-41e8-afb1-e65f0bdd07b4

Connum:
> @Connum: latest Scrutinizer run crashed, can you please have a look?
> 
> https://scrutinizer-ci.com/g/smalot/pdfparser/inspections/2db87eb0-536f-41e8-afb1-e65f0bdd07b4/log

That's strange! It fails in `tests/Integration/ParserTest.php:61`, so before the memory consumption test that I added. Two ideas that come to mind:
1. Let's call `gc_collect_cycles()` in `setUp()` before doing anything else, to clean any leftovers from previous tests?
2. In `testRetainImageContentImpact()`, it should be enough to test with one exectution instead of a loop of currently 2 iterations, to see the impact of the setting (we'd have to adapt the value of currently `200000000` bytes, of course, to whatever the expected memory with and without the setting roughly is)

I'd suggest to test 1. first, then 2. if that doesn't help

k00ni:
It seems that `testParseFile` loads all available PDFs from `samples/bugs` and tries to parse them ([ref](https://github.com/smalot/pdfparser/blob/master/tests/Integration/ParserTest.php#L51-L80)). But it run out of memory.

Calling `gc_collect_cycles()` before each PDF might help here. I don't think we have to call it in `setUp()`. Thanks for the tip!

Now we need to check if thats the problem.

k00ni:
I made some short tests in a separate branch: https://github.com/smalot/pdfparser/commits/fix/out-of-memory-error-parsertest-scrutinizer

Adding `gc_collect_cycles()` just seem to move the problem to another file. I assume it doesn't fail in Github Action because it provides more memory. Any idea how to solve this?

Connum:
> Adding `gc_collect_cycles()` seem to only move the problem to another file.

Do you mean another one of the demo files that are being looped over, or another test file?

If it were crashing on the recently added memory test, I'd suggest we could maybe use `ini_get('memory_limit')` in conjunction with `memory_get_usage()` to decide whether to skip the test, just like we are skipping it for PHP <= 7.3 already. But as it seems to be the sheer number and size of test pdfs being parsed... Here's a thought: Do we really need to parse all the test files in a dedicated test? Most of the files should be related to a specific issue and have an additional test that parses the file in order to get some content and see if that was parsed correctly. If there are some that don't have a specific additional test, we should parse just those explicitly instead.

k00ni:
> Do you mean another one of the demo files that are being looped over, or another test file?

I meant another PHP (test) file.

> If it were crashing on the recently added memory test, I'd suggest we could maybe use ini_get('memory_limit') in conjunction with memory_get_usage() to decide whether to skip the test, just like we are skipping it for PHP <= 7.3 already. But as it seems to be the sheer number and size of test pdfs being parsed... Here's a thought: Do we really need to parse all the test files in a dedicated test? Most of the files should be related to a specific issue and have an additional test that parses the file in order to get some content and see if that was parsed correctly. If there are some that don't have a specific additional test, we should parse just those explicitly instead.

Good idea, let me think about it.

PR #443, opened 2021-07-29
Fixes Out-of-memory errors during Scrutinizer run

k00ni:
Our tests run fine in Github Actions. But when running in Scrutinizer PHP runs out of memory at some point. Because I couldn't fix these tests, I marked them as `memory-heavy` (see [PHPUnit doc](https://phpunit.de/manual/6.5/en/appendixes.annotations.html#appendixes.annotations.group)) and ignore them when running Scrutinizer. We know that PDFParser might not have the best memory management, but at least this way we can make sure it works and keep using advantages of Scrutinizer (like code analysis etc.).

PR also contains a few improvements on new unit-test code of #441.

Thanks @Connum for your input in #441.

Any remarks?

CC @j0k3r

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -9,7 +9,7 @@ build:
                         # to not enough memory.
                         # @https://github.com/smalot/pdfparser/issues/410
                         # @https://github.com/smalot/pdfparser/pull/412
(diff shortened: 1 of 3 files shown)


j0k3r (commented):
That's a good idea, maybe you can squash all your commits

k00ni:
> That's a good idea, maybe you can squash all your commits

Of course :)

Issue #450, opened 2021-08-17
Error thrown on getDataTm() - Call to a member function decodeText() on null


eddturtle:
Hello, I'm trying to find the X, Y coords for a specific piece of text inside a PDF. I'm trying to use `getDataTm()` (correct me if that's the wrong method to use).

This works for many pdfs, but throws an error for this one example pdf.

[myfile.pdf](https://github.com/smalot/pdfparser/files/6998873/myfile.pdf)

Example code:

```php
$parser = new \Smalot\PdfParser\Parser();
$pdf = $parser->parseFile('myfile.pdf');

$pages = $pdf->getPages();
$page = $pages[0];
$dataTm = $page->getDataTm();
```

Error thrown:

```
EXCEPTION (Error): vendor/smalot/pdfparser/src/Smalot/PdfParser/PDFObject.php line 484 user 1234 -- Call to a member function decodeText() on null

array(3) {
  [0]=>
  array(5) {
    ["file"]=>
    string(62) "/vagrant/vendor/smalot/pdfparser/src/Smalot/PdfParser/Page.php"
    ["line"]=>
    int(257)
    ["function"]=>
    string(12) "getTextArray"
    ["class"]=>
    string(26) "Smalot\PdfParser\PDFObject"
    ["type"]=>
    string(2) "->"
  }
  [1]=>
  array(5) {
    ["file"]=>
    string(62) "/vagrant/vendor/smalot/pdfparser/src/Smalot/PdfParser/Page.php"
    ["line"]=>
    int(561)
    ["function"]=>
    string(12) "getTextArray"
    ["class"]=>
    string(21) "Smalot\PdfParser\Page"
    ["type"]=>
    string(2) "->"
  },
  // removed
}
```

I've tried this on php 7.4 and php 8.0 (running through apache2) on ubuntu 18.04.

Any ideas on how to get this pdf to process?

k00ni:
Maybe @Connum or @izabala can help out here?

izabala:
@K00ni I will have a look. I have a long time since I helped on this.  May you (or someone  else ) help me, remembering the things I have to do prior to commit the changes (I remmember that i will make the Unit Test changes in Atoum, but I dont remmember if i have to use any style tool or anything else... So if you have any help on things to do for changing the files, It will help me to do it quickly).

@eddturtle I have a look at the file, it just have 1 line... with: {signature:signer505906:Please+Sign+Here} on it.  Is that ok?

izabala:
[...]

PR #452, opened 2021-08-19
workaround for the Issue #450

izabala:
The file makes that 2 of the Page methods fails.

1. The Page->extractDecodedRawData was not returning the correct string.  This was corrected.

2. The Page->getTextArray breaks when the Page->get(´Contents´) returns a PDFObject, but this object makes that the PDFObject->getTextArray($this) throw an Error. But if you detected it and instead call PDFObject->getTextArray() , it returns the correct data.  This is a workaround, because, what is exactly the difference in the format of this PDF and why it fails, needs to have a more deep investigation. I run all the PageTests and they work.

This happends because the sample Pdf file is not format as we usually see in other files. Actually, I have a similar (not exactly the same) case for a file created with FPDI, that also broke the getTextArray and getDataTm methods, but I am doing a research to see what is actually happends before I open an Issue for that.  As soon as I know what is happening in that case, I will opened the Issue, hopefully with the workaround or fix already done.


izabala:
Please delete this.. I will make another one, hopefully with the CS Fixer run!

k00ni:
> Please delete this.. I will make another one, hopefully with the CS Fixer run!

Pull requests can't be deleted, only closed.

Btw., if you create another one can you please use a dedicated branch for it and not `master`? The reason is, that we can't push into your master in case additional help is needed.

PR #453, opened 2021-08-19
workaround for the Issue #450

izabala:
The file makes that 2 of the Page methods fails.

The Page->extractDecodedRawData was not returning the correct string. This was corrected.

The Page->getTextArray breaks when the Page->get(´Contents´) returns a PDFObject, but this object makes that the PDFObject->getTextArray($this) throw an Error. But if you detected it and instead call PDFObject->getTextArray() , it returns the correct data. This is a workaround, because, what is exactly the difference in the format of this PDF and why it fails, needs to have a more deep investigation. I run all the PageTests and they work.

This happends because the sample Pdf file is not format as we usually see in other files. Actually, I have a similar (not exactly the same) case for a file created with FPDI, that also broke the getTextArray and getDataTm methods, but I am doing a research to see what is actually happends before I open an Issue for that. As soon as I know what is happening in that case, I will opened the Issue, hopefully with the workaround or fix already done.


izabala:
Guys, I already run CS Fixer, but I still get the errors.   Please some advice here.

k00ni:
Thank you for your work @izabala!

> Guys, I already run CS Fixer, but I still get the errors. Please some advice here.

I am busy this week, but I will try to assist you next week.

eddturtle:
I grabbed a copy of your repo @izabala + phpcs was failing (like above). I opened it in my editor and saved and it auto changed quite a few things, now it passes php cs. **Here's the updated code:**
```php
    public function testExtractDecodedRawDataIssue450()
    {
        $filename = $this->rootDir.'/samples/bugs/Issue450.pdf';
        $parser = $this->getParserInstance();
        $document = $parser->parseFile($filename);
        $page = $pages[0];
        $extractedDecodedRawData = $page->extractDecodedRawData();
        $this->assertIsArray($extractedDecodedRawData);
        $this->assertGreaterThan(3, \count($extractedDecodedRawData));
        $this->assertIsArray($extractedDecodedRawData[3]);
        $this->assertEquals('TJ', $extractedDecodedRawData[3]['o']);
        $this->assertIsArray($extractedDecodedRawData[3]['c']);
        $this->assertIsArray($extractedDecodedRawData[3]['c'][0]);
        $this->assertEquals(3, \count($extractedDecodedRawData[3]['c'][0]));
        $this->assertEquals('{signature:signer505906:Please+Sign+Here}', $extractedDecodedRawData[3]['c'][0]['c']);
    }

    public function testGetDataTmIssue450()
    {
        $filename = $this->rootDir.'/samples/bugs/Issue450.pdf';
        $parser = $this->getParserInstance();
        $document = $parser->parseFile($filename);
        $page = $pages[0];
        $dataTm = $page->getDataTm();
        $this->assertIsArray($dataTm);
        $this->assertEquals(1, \count($dataTm));
        $this->assertIsArray($dataTm[0]);
        $this->assertEquals(2, \count($dataTm[0]));
        $this->assertIsArray($dataTm[0][0]);
        $this->assertEquals(6, \count($dataTm[0][0]));
        $this->assertEquals(1, $dataTm[0][0][0]);
        $this->assertEquals(0, $dataTm[0][0][1]);
        $this->assertEquals(0, $dataTm[0][0][2]);
        $this->assertEquals(1, $dataTm[0][0][3]);
        $this->assertEquals(67.5, $dataTm[0][0][4]);
        $this->assertEquals(756.25, $dataTm[0][0][5]);
        $this->assertEquals('{signature:signer505906:Please+Sign+Here}', $dataTm[0][1]);
    }
```

Hope this helps!

As mentioned in the issue, I think the actual code change is looking good to me.

izabala:
Thanks @k00ni.  I wait, I just want to be sure what I am doing wrong!

src/Smalot/PdfParser/Page.php
```diff
@@ -240,6 +240,12 @@ public function getTextArray(self $page = null): array
 
                     $header = new Header([], $this->document);
                     $contents = new PDFObject($this->document, $header, $new_content, $this->config);
+                } else {
+                    try {
+                        $contents->getTextArray($this);
+                    } catch (\Error $e) {
```
k00ni, line 246: Why not `Exception` here?

k00ni (changes requested):
@izabala thank you for the PR.

I took the liberty to directly adapt a few files in your repository (see latest commits). All tests are good now.

In ParserTest I had to refine a few lines to fix failing test. It was not related to your changes.

I have one question (see below). After we solved it we are good to go.

src/Smalot/PdfParser/Page.php
```diff
@@ -345,7 +351,8 @@ public function extractDecodedRawData(array $extractedRawData = null): array
                     $decodedText = '';
                     if (isset($currentFont)) {
                         $decodedText = $currentFont->decodeOctal($tmpText);
-                        //$tmpText = $currentFont->decodeHexadecimal($tmpText, false);
+                    } else {
+                        $decodedText = $tmpText;
```
j0k3r, line 355: Maybe this can be simplified:

```php
$decodedText = isset($currentFont) ? $currentFont->decodeOctal($tmpText) : $tmpText;
```

src/Smalot/PdfParser/Page.php
izabala: Because it raises (throw) and Error and not an Exception. (I don't know why it throw Error instead of an Exception). So catching an Exception will not help because Error class does not inherit from Exception:
`class Error implements Throwable` (This is from the php online docs in https://www.php.net/manual/en/class.error.php)

Actually in the php online documentations: https://www.php.net/manual/en/language.errors.php7.php
It says: 
_**As the Error hierarchy does not inherit from Exception, code that uses catch (Exception $e) { ... } blocks to handle uncaught exceptions in PHP 5 will find that these Errors are not caught by these blocks. Either a catch (Error $e) { ... } block or a set_exception_handler() handler is required.**_
And even though it says that the problem is in PHP 5, I tested it... I am using PHP 8 and it does not catch it, unless you catch Error class.

src/Smalot/PdfParser/Page.php
izabala: Done!  Thanks for the suggestion!

src/Smalot/PdfParser/Page.php
izabala: Let me know if my answer is OK.

src/Smalot/PdfParser/Page.php
```diff
@@ -343,10 +349,7 @@ public function extractDecodedRawData(array $extractedRawData = null): array
                     }
                     $tmpText = $data[$i]['c'];
                     $decodedText = '';
```
j0k3r, line 345: You can then remove that line

izabala: Done. Thanks.

src/Smalot/PdfParser/Page.php
k00ni: I would catch `Throwable` then. In both cases the semantics behind is `Something went wrong`, isn't it?

izabala: Good! Let me do that... and test!!!

izabala: Good advise @k00ni! Done!

k00ni (approved):
👍🏿 Thanks.

As always, I will keep this open for a few days so others may comment.

Issue #454, opened 2021-08-24
extractRawData, extractDecodedRawData, getDataTm and getDataXY do not work with a Pdf file produced by FPDI/FPDF


izabala:
I am using FPDI/FPDF (setasign\Fpdi\Fpdi) to split and merge some PDF files.  When I try to parsed them, all the mentioned methods failed. They return nothing or almost nothing (empty arrays).  
After that, I create a PDF file (also done with FPDI/FPDF) merging some of the sample files for this library. And tried again.  They failed. Here the sample file:
[FpdiFpdf.pdf](https://github.com/smalot/pdfparser/files/7041458/FpdiFpdf.pdf)

I made my investigation: 
`Page->getText()` works fine.  But all the info that make posible to get the Text Matrix done in the methods `extractRawData` and the rest, are not keep it as in normal Pdf files.  The information for Pdf commands and Text Matrix are saved inside the `xObjects `in a Form class, not in the pages. 

I already made a fix/workaround for that, that involved changes in:  g`etTextArray`, `extractRawData`, `extractDecodedRawData`, but it involve to know the Producer of the file (It asked if the file was produced by FPDF) in order to know what to do, and this is because I couldn't find any other way to work this out.

I will wait, till the pull request #453 (for the Issue #450) is finally merged into the master branch, because it touches the same methods,  to make a pull request using my fix/workaround for this new problem.

izabala:
I got the same CS Fixer and PHPStan errors.

PR #455, opened 2021-08-28
Fix for Issue #454

izabala:
This is the fix/workaround I have been working this to fix that Issue.


izabala:
Hi, @k00ni it seems that I still have the problem (when I create a new branch) with the CS Fixer.  May you help me with that?

src/Smalot/PdfParser/Page.php
```diff
@@ -216,8 +216,56 @@ public function getText(self $page = null): string
         return '';
     }
 
+    /**
+     * Return True if the current page is a (setasign\Fpdi\Fpdi) FPDI/FPDF document
```
k00ni, line 220:
```suggestion
     * Return true if the current page is a (setasign\Fpdi\Fpdi) FPDI/FPDF document
```

src/Smalot/PdfParser/Page.php
```diff
@@ -216,8 +216,56 @@ public function getText(self $page = null): string
...
+     * 
+     * @return bool true is the current page is a FPDI/FPDF document
+     */
+    public function isFpdf(): Bool
```
k00ni, line 224:
```suggestion
    public function isFpdf(): bool
```

src/Smalot/PdfParser/Page.php
```diff
@@ -216,8 +216,56 @@ public function getText(self $page = null): string
...
+    {
+        $pages = $this->document->getPages();
+        $numOfPages = count($pages);
+        for ($pageNum = 0; $pageNum < $numOfPages; $pageNum++) {
+            if ($pages[$pageNum] === $this) {
+                break;
+            }
+        }
+        return $pageNum;
+    }
```
k00ni, line 249: What does it mean if `$pageNum` is `0`?

src/Smalot/PdfParser/Page.php
```diff
@@ -292,6 +340,16 @@ public function extractRawData(): array
...
         } else {
+            if ($this->isFpdf()) {
+                /** 
+                 * This code is for the (setasign\Fpdi\Fpdi) FPDI-FPDF documents. 
+                 * The page number is important for getting the PDF Commands and Text Matrix 
+                 */
+                    $pageNum = $this->getPageNumber();
+                    $xObjects = $this->getXObjects();
+                    /** The correct page info is in $xObject[$pageNum] */
+                    $content = $xObjects[$pageNum];
```
k00ni, line 351:
```suggestion
                /*
                 * This code is for the (setasign\Fpdi\Fpdi) FPDI-FPDF documents. 
                 * The page number is important for getting the PDF Commands and Text Matrix 
                 */
                $pageNum = $this->getPageNumber();
                $xObjects = $this->getXObjects();
                // The correct page info is in $xObject[$pageNum]
                $content = $xObjects[$pageNum];
```

src/Smalot/PdfParser/Page.php
```diff
@@ -321,6 +379,16 @@ public function extractDecodedRawData(array $extractedRawData = null): array
         }
         $currentFont = null; /** @var Font $currentFont */
         $clippedFont = null;
+        $xObject = null;
+        if ($this->isFpdf()) {
+            /** This code is for the (setasign\Fpdi\Fpdi) FPDI-FPDF documents. 
+             * The page number is important for getting the PDF Commands and Text Matrix 
+             */
```
k00ni, line 386:
```suggestion
            /* 
             * This code is for the (setasign\Fpdi\Fpdi) FPDI-FPDF documents. 
             * The page number is important for getting the PDF Commands and Text Matrix 
             */
```

tests/Integration/PageTest.php
```diff
@@ -611,4 +611,93 @@ public function testGetDataTmIssue450()
         $this->assertEquals(756.25, $dataTm[0][0][5]);
         $this->assertEquals('{signature:signer505906:Please+Sign+Here}', $dataTm[0][1]);
     }
+    public function testIsFpdf()
```
k00ni, line 614:
```suggestion
    }

    public function testIsFpdf()
```

k00ni (changes requested):
I don't have much time these days, but I hope it helps you a bit for now.

Also, did you run php-cs-fixer in `dev-tools` folder?

> make run-php-cs-fixer

izabala:
Ok... I am just travelling for 2 weeks.   As soon as I get back, I will make all changes and tests.

izabala:
I am getting back to this.  I will start looking all the recommendations.  Just one thing:

I have to look how to use the:
make run-php-cs-fixer

This is because I develop in a windows machine where I can run make, I am looking if I can install a windows tools that let me use make.

I let you know.

src/Smalot/PdfParser/Page.php
izabala: It means that it is the first page.  This is a work around for this bug.  The reason, is because the whole information of the Text Matrix (the content), is not save in the page, but in the xObject array inside the page, but is in array of pages, so you need the page number... but in the Page object, there isn't an attribute that tells you which page of the document the page is.

This is just for the fpdf/fpdi documents.

izabala:
I already installed a make tool for windows (the GNU one). But when you run the:
`make run-php-cs-fixer`
but it tries to run: 
`dev-tools/vendor/bin/php-cs-fixer fix` 
that in windows is not a command (the / instead of \\).
So I have to run it manually ... it makes 64 changes in files that I havent touched. I will push that changes. Let me know if I have to re-do everything again.

izabala:
Ok. Let me know if with the changes, this is Ok.

src/Smalot/PdfParser/Page.php
```diff
@@ -363,7 +442,8 @@ public function extractDecodedRawData(array $extractedRawData = null): array
                 }
             } elseif ('Tf' == $command['o'] || 'TF' == $command['o']) {
                 $fontId = explode(' ', $command['c'])[0];
-                $currentFont = $this->getFont($fontId);
+                // If document is a FPDI/FPDF the $page has the correct font
+                $currentFont = (isset($page)) ? $page->getFont($fontId) : $this->getFont($fontId);
```
k00ni, line 446:
```suggestion
                $currentFont = isset($page) ? $page->getFont($fontId) : $this->getFont($fontId);
```

.vscode/launch.json
```diff
@@ -0,0 +1,48 @@
+{
```
k00ni, line 1: This file is new. Please remove.

k00ni (changes requested):

k00ni:
I try to get back to you as soon as possible. Looks good so far.

.vscode/launch.json
izabala: Yes., sorry, that was for debugging! I forgot to deleted it.

izabala (commented):
I took it out, that was a file for debugging that I forget to take it out.

k00ni (commented):
First of all, than you for another important pull request.

Your implementation looks fine, but I would like to discuss the places of these if-clauses which check for isFpdf. Using these here and there seems to bloat the code to an extent and makes it also a bit harder to read.

In your opinion, does it makes sense to use special functions, like `getTextAreaForFPdf` etc. to handle it? Or add further functions, because some code seems like (almost) the same multiple times.

What do you think?

src/Smalot/PdfParser/Page.php
```diff
@@ -216,8 +216,59 @@ public function getText(self $page = null): string
...
 
+    /**
+     * Return true if the current page is a (setasign\Fpdi\Fpdi) FPDI/FPDF document
+     *
+     * @return bool true is the current page is a FPDI/FPDF document
+     */
+    public function isFpdf(): bool
+    {
+        if (\array_key_exists('Producer', $this->document->getDetails()) &&
+            \is_string($this->document->getDetails()['Producer']) &&
```
k00ni, line 227: When is `Producer` not a string?

izabala: Well, I don't remmember if that could be null.  But I may just taking care that str_starts_with is apply to a String.  Do you want me to take it out?

izabala:
@k00ni About your comment:

```
In your opinion, does it makes sense to use special functions, like getTextAreaForFPdf etc. to handle it? Or add further functions, because some code seems like (almost) the same multiple times. I would prefer the latter.

What do you think?
```
Actually I just start doing a workaround/fix over the problem as soon as possible.  Let me re-think the fix, and see if there is a better solution, now that we Know what the problem is, and how to solve it.

Give me a couple of days to work on this.

izabala:
@k00ni , May you have a look on this approach? Let me know if this is ok or not.

Thanks for all your recommendations

src/Smalot/PdfParser/Page.php
k00ni: > Do you want me to take it out?

I would like to discuss implications for this if-clause first. We should document background information (maybe with a link to the spec?) so that in the future developer know what it means. For instance, are there other information we can use to determine whether something is produced by FPdf or not? I know that is additional work, anything in this direction is appreciated!

src/Smalot/PdfParser/Page.php
```diff
@@ -216,8 +216,91 @@ public function getText(self $page = null): string
...
+
+        return $pageNum;
+    }
+
+    /**
+     * Return the xObject if the document is from fpdf
+     *
+     * @return object The xObject for the page
+     */
+    public function getXObjectForFpdf()
```
k00ni, line 258: Can you add the return type here please? Is it `object` or does it differ from time to time?

src/Smalot/PdfParser/Page.php
```diff
@@ -216,8 +216,91 @@ public function getText(self $page = null): string
...
+
+        return $xObjects[$pageNum];
+    }
+
+    /**
+     * Return a PDFObject if document is from fpdf
+     *
+     * @return object The xObject for the page
+     */
+    public function getPDFObjectForFpdf()
```
k00ni, line 271: Add return type please, see comment above.

src/Smalot/PdfParser/Page.php
```diff
@@ -216,8 +216,91 @@ public function getText(self $page = null): string
...
+
+        return new PDFObject($xObject->document, $header, $new_content, $config);
+    }
+
+    /**
+     * Return page if document is from fpdf
+     *
+     * @return object The page
+     */
+    public function getPageForFpdf()
```
k00ni, line 286: Add return type please, see comment above.

tests/Integration/PageTest.php
```diff
@@ -611,4 +611,94 @@ public function testGetDataTmIssue450()
...
+        $page = $pages[0];
+        $this->assertFalse($page->isFpdf());
+        $filename = $this->rootDir.'/samples/bugs/Issue454.pdf';
+        $document = $parser->parseFile($filename);
+        $pages = $document->getPages();
+        $page = $pages[0];
+        $this->assertTrue($page->isFpdf());
+    }
+
+    public function testGetPageNumber()
```
k00ni, line 630:
```suggestion
    public function testGetPageNumber(): void
```

tests/Integration/PageTest.php
```diff
@@ -611,4 +611,94 @@ public function testGetDataTmIssue450()
         $this->assertEquals(756.25, $dataTm[0][0][5]);
         $this->assertEquals('{signature:signer505906:Please+Sign+Here}', $dataTm[0][1]);
     }
+
+    public function testIsFpdf()
```
k00ni, line 615:
```suggestion
    public function testIsFpdf(): void
```

tests/Integration/PageTest.php
```diff
@@ -611,4 +611,94 @@ public function testGetDataTmIssue450()
...
+        $this->assertEquals(0, $page->getPageNumber());
+        $page = $pages[1];
+        $this->assertEquals(1, $page->getPageNumber());
+        $page = $pages[2];
+        $this->assertEquals(2, $page->getPageNumber());
+        $page = $pages[3];
+        $this->assertEquals(3, $page->getPageNumber());
+    }
+
+    public function testIssue454()
```
k00ni, line 671:
```suggestion
    public function testIssue454(): void
```

k00ni:
It looks better, thank you. I try to find the time to give you a more in depth feedback soon, hopefully next week, maybe soon.

src/Smalot/PdfParser/Page.php
izabala: Understood.  I use two sources of the information:
### 1. The Pdf 1.7 Specification.   [PDF 32000-1:2008](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf)
**Chapter 14: Document Interchange, Section 14.3: Metada, subSection: 14.3.3 Document Information Dictionary** 
**Page: 550**
It defines: 
> **Key:** Producer **Type:** text string **Value:** (Optional) If the document was converted to PDF from another format, the name of the conforming product that converted it to PDF.

It also has an example of values in the same page.

### 2. The FPDF source code. The file **fpdf.php**
> **FPDF Version:** 1.83 **Date:** 20021-04-18 **Author:** Olivier Plathey

In the line: 1834 it defines the function _putinfo():

```
protected function _putinfo()
{
	$this->metadata['Producer'] = 'FPDF '.FPDF_VERSION;
	$this->metadata['CreationDate'] = 'D:'.@date('YmdHis');
	foreach($this->metadata as $key=>$value)
		$this->_put('/'.$key.' '.$this->_textstring($value));
}
```
Here the metadata with key: **'Producer'** it set to **"FPDF 1.83"**
and in the line 1875 it defines the function **_enddoc()**, in this functions it set all the properties and values of the document. This include the line 1883: **`$this->_putinfo();`**

So with this 2 information, we know that the metadata with the key = 'Producer' will have the text value: 'FPDF ' . FPDF_VERSION, for any document generated/saved by FPDF.

izabala: Let me know if with this is Ok.

izabala:
Is anything more to do on this?

k00ni (approved):

izabala:
I am OK with this!

PR #457, opened 2021-09-03
Prevent endless loops during Page->getText() because of faulty recursionStack

Nickmanbear:
The $recursionStack used in the getText() function of PDFObject to prevent "Do" commands from causing endless loops removes items from it's stack prematurely during the parsing of pages. Meaning endless loops can and are still being caused by the Do command.

Only reseting the recursionStack at the end of the getText() function for every page prevents this bug.

### DIFF
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -176,6 +176,8 @@ class Page extends PDFObject
 
     public function getText(self $page = null): string
     {
+        PDFObject::$recursionStack = [];
+        
         if ($contents = $this->get('Contents')) {
             if ($contents instanceof ElementMissing) {
                 return '';
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -210,9 +210,11 @@ class Page extends PDFObject
                 $contents = new PDFObject($this->document, $header, $new_content, $this->config);
             }
 
-            // Elements referecing eachother on the same page can cause endless loops during text parsing.
-            // To combat this we keep a recursionStack containing already parsed elements on the page.
-            // The stack is only emptied here after getting text from a page.
+            /*
+             * Elements referecing eachother on the same page can cause endless loops during text parsing.
+             * To combat this we keep a recursionStack containing already parsed elements on the page.
+             * The stack is only emptied here after getting text from a page.
+             */
             $contentsText = $contents->getText($this);
             PDFObject::$recursionStack = [];
 
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -176,8 +176,6 @@ class Page extends PDFObject
 
     public function getText(self $page = null): string
     {
-        PDFObject::$recursionStack = [];
-        
         if ($contents = $this->get('Contents')) {
             if ($contents instanceof ElementMissing) {
                 return '';
@@ -212,7 +210,10 @@ class Page extends PDFObject
                 $contents = new PDFObject($this->document, $header, $new_content, $this->config);
             }
 
-            return $contents->getText($this);
+            $contentsText = $contents->getText($this);
+            PDFObject::$recursionStack = [];
+
+            return $contentsText
         }
 
         return '';
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -211,7 +211,7 @@ class Page extends PDFObject
             }
 
             /*
-             * Elements referecing eachother on the same page can cause endless loops during text parsing.
+             * Elements referencing each other on the same page can cause endless loops during text parsing.
              * To combat this we keep a recursionStack containing already parsed elements on the page.
              * The stack is only emptied here after getting text from a page.
              */
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -210,6 +210,9 @@ class Page extends PDFObject
                 $contents = new PDFObject($this->document, $header, $new_content, $this->config);
             }
 
+            // Elements referecing eachother on the same page can cause endless loops during text parsing.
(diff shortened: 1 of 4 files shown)


amooij:
This great, @Nickmanbear. This solves headaches for many people, great work!

Nickmanbear:
@amooij Thanks!

k00ni:
@Nickmanbear thank you for your pull request!

Is there a way to "demonstrate" your change? Like a test which triggers an endless loop?

Nickmanbear:
@k00ni These are a few of the files we have that cause endless loops. The first two will get stuck when you call getText() on the first page and the third one when you call getText() on the second page.

[20210514101853.pdf](https://github.com/smalot/pdfparser/files/7114544/20210514101853.pdf)
[2021081795357.pdf](https://github.com/smalot/pdfparser/files/7114813/2021081795357.pdf)
[202108201456185.pdf](https://github.com/smalot/pdfparser/files/7114824/202108201456185.pdf)

With my fix these files are all able to be parsed just fine. I could add one of these files to the Page integration tests to prove it now works and make sure it continues to work in the future. But I'm not sure how I would write a meaningful test to trigger this endless loop when this fix should prevent it from occurring.

k00ni:
> But I'm not sure how I would write a meaningful test to trigger this endless loop when this fix should prevent it from occurring.

Its a good start to have such a test, so one can deactivate the fix and see if such a loop gets triggered.

tests/Integration/PageTest.php
```diff
@@ -141,6 +141,26 @@ public function testGetText()
         $this->assertStringContainsString('Verdana', $text);
     }
 
+    public function testGetTextPullRequest457()
+    {
```
k00ni, line 145:
```suggestion
    /**
      * @see https://github.com/smalot/pdfparser/pull/457
      */
    public function testGetTextPullRequest457()
    {
```

src/Smalot/PdfParser/Page.php
```diff
@@ -210,7 +210,10 @@ public function getText(self $page = null): string
                 $contents = new PDFObject($this->document, $header, $new_content, $this->config);
             }
 
-            return $contents->getText($this);
+            $contentsText = $contents->getText($this);
+            PDFObject::$recursionStack = [];
```
k00ni, line 214: Please add a short comment about those two lines.

Nickmanbear: Here in the code or above the integration test?

k00ni: In the code above the two lines I marked please, so developers get an idea why this code is used.

Nickmanbear: I've added a comment.

src/Smalot/PdfParser/Page.php
```diff
@@ -210,7 +210,13 @@ public function getText(self $page = null): string
                 $contents = new PDFObject($this->document, $header, $new_content, $this->config);
             }
 
-            return $contents->getText($this);
+            // Elements referecing eachother on the same page can cause endless loops during text parsing.
+            // To combat this we keep a recursionStack containing already parsed elements on the page.
+            // The stack is only emptied here after getting text from a page.
```
k00ni, line 215:
```suggestion
            /*
             * Elements referecing eachother on the same page can cause endless loops during text parsing.
             * To combat this we keep a recursionStack containing already parsed elements on the page.
             * The stack is only emptied here after getting text from a page.
             */
```

tests/Integration/PageTest.php
```diff
@@ -141,6 +141,38 @@ public function testGetText()
...
+        $this->assertTrue(1000 < \strlen($text));
+        $this->assertStringContainsString('SUPER', $text);
+        $this->assertStringContainsString('VOORDEEL', $text);
+        $this->assertStringContainsString('KRANT', $text);
+        $this->assertStringContainsString('DINSDAG', $text);
+        $this->assertStringContainsString('Snelfilterkoffie', $text);
+        $this->assertStringContainsString('AardappelenZak', $text);
+        $this->assertStringContainsString('ALL', $text);
+
+        // Force garbage collection to not interfere with other tests.
```
k00ni, line 166: Can you explain what problem you solve here exactly? Do you run in memory-leaks?

In general test have to be isolated and must not interfere with each other.

Nickmanbear: The unit tests are failing because `testRetainImageContentImpact()` from the [Parser integration tests](https://github.com/smalot/pdfparser/blob/master/tests/Integration/ParserTest.php) checks if memory usage is under a certain amount after executing its test. However it appears that PHP doesn't run garbage collection fast enough after our new test here (or something of that nature) causing the pdf parsed here to still be in memory during that test.

Nickmanbear: Perhaps it would be better to call gc_collect_cycles() before testRetainImageContentImpact() instead of after this test. However I was unsure if editing other tests to make a new test work would be appreciated.

k00ni: Editing other tests is fine if you found a problem. Please propose a change which works in your opinion.

Nickmanbear: It took me a bit of testing but I think I've finally found a good way to fix the issue.

testRetainImageContentImpact() now checks against a baseline memory it has at the beginning of the test. Instead of an arbitrary value.

Nickmanbear:
@k00ni Are there any more changes you think I should make?

k00ni:
I plan to get back to you next week. For now all I can say is that it is looking good.

k00ni:
I am still super busy and try to find some time to finalize my review. Hope its not urgent to merge it.

Nickmanbear:
@k00ni We have switched to using this fork for our production environment. So merging this is not that urgent to us. Of course we would still like it to be in the next release, so we can switch back to that.

k00ni (approved):
Looks good to me.

PR #461, opened 2021-09-24
Fix TypeError on Document::getFirstFont when no fonts are available

PrinsFrank:
When there is no font available in a document, an empty array is passed to the reset function. Reset can return ["false if the array is empty"](https://www.php.net/manual/en/function.reset), causing a TypeError as the return type of the function "getFirstFont" is set to "?Font". This PR fixes this issue.


k00ni (approved):
Good catch!

PR #466, opened 2021-10-08
Fix TypeError on default font when no fonts available

PrinsFrank:
When there is no font available for a page, 'getFirstFont' returns null. GetDefaultFont then returns that null value, where the returntype is 'Font', causing a TypeError. This PR fixes that issue.

### DIFF
--- a/tests/Unit/PDFObjectTest.php
+++ b/tests/Unit/PDFObjectTest.php
@@ -1,4 +1,5 @@
 <?php
+
 declare(strict_types=1);
 
 namespace Tests\Smalot\PdfParser\Unit;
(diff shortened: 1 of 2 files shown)


tests/Unit/PDFObjectTest.php
```diff
@@ -0,0 +1,25 @@
...
+    public function testGetTextOnNullPage(): void
+    {
+        static::assertSame(' ', (new PDFObject(new Document()))->getText());
+    }
+
+    public function testGetTextOnPageWithoutContent(): void
+    {
+        $document = new Document();
+
+        static::assertSame(' ', (new PDFObject($document, null, null))->getText(new Page($document)));
```
k00ni, line 23:
```suggestion
        $this->assertSame(' ', (new PDFObject(new Document()))->getText());
    }

    public function testGetTextOnPageWithoutContent(): void
    {
        $document = new Document();

        $this->assertSame(' ', (new PDFObject($document, null, null))->getText(new Page($document)));
```

Can you tell me why a white space is the default return value here?

k00ni (changes requested):
Good catch!

PR #467, opened 2021-10-14
Font.php: Optimization of the uchr function

mariuszkrzaczkowski:
I noticed that the mb_convert_encoding function is called multiple times for the same value, which is redundant

### DIFF
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -114,10 +114,9 @@ class Font extends PDFObject
     
     public static function uchr(int $code): string
     {
-        $code = (int) $code;
-		if (isset(self::$uchrCache[$code])) {
-			return self::$uchrCache[$code];
-		}
+        if (isset(self::$uchrCache[$code])) {
+		    return self::$uchrCache[$code];
+        }
         // html_entity_decode() will not work with UTF-16 or UTF-32 char entities,
         // therefore, we use mb_convert_encoding() instead
         return self::$uchrCache[$code] = mb_convert_encoding("&#{$code};", 'UTF-8', 'HTML-ENTITIES');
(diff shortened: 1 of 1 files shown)


src/Smalot/PdfParser/Font.php
```diff
@@ -110,11 +110,17 @@ public function translateChar(string $char, bool $use_default = true)
...
     }
 
+    private static $uchrCache = [];
+    
     public static function uchr(int $code): string
     {
+        $code = (int) $code;
+		if (isset(self::$uchrCache[$code])) {
+			return self::$uchrCache[$code];
+		}
```
k00ni, line 120:
```suggestion
        if (isset(self::$uchrCache[$code])) {
		    return self::$uchrCache[$code];
        }
```

k00ni: `$code = (int) $code;`

is not required because function sends you an `integer`.

mariuszkrzaczkowski:
I noticed that the mb_convert_encoding function is called multiple times for the same value, which is redundant

src/Smalot/PdfParser/Font.php
```diff
@@ -110,11 +110,16 @@ public function translateChar(string $char, bool $use_default = true)
         return $use_default ? self::MISSING : $fallbackDecoded;
     }
 
+    private static $uchrCache = [];
+    
     public static function uchr(int $code): string
     {
+        if (isset(self::$uchrCache[$code])) {
+		    return self::$uchrCache[$code];
+        }
```
k00ni, line 119:
```suggestion
        if (isset(self::$uchrCache[$code])) {
            return self::$uchrCache[$code];
        }
```
```suggestion
        if (isset(self::$uchrCache[$code])) {
		    return self::$uchrCache[$code];
        }
```

k00ni:
I made two commits to your branch, please check.

Issue #475, opened 2021-10-29
Crashing PHP process through memory exhaustion when decompressing images.


b3n-l:
Based on issues #104 and #357

I think we have an issue around decompression bombs, where the output file is many times larger than the input file. The demo on pdfparser.org gives the following fatal error: 
`Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 25708532 bytes) in /var/www/html/sites/all/modules/custom/pdfparser_demo/pdfparser/src/Smalot/PdfParser/RawData/FilterHelper.php on line 244`

The problem is, before it checks the `Config` class for disabled image parsing, you need to have called `gzuncompress` in `FilterHelper.php`. If this pushes it above the PHP memory limit, it'll simply die. This is why the config option doesn't stop the crash. 

xpdf, as example of another popular PHP library, checks for compression bombs by doing the following in C++. This is lifted directly from the source code in Stream.cc
```
// check for a 'decompression bomb'
  if (totalOut > 50000000 && totalIn < totalOut / 250) {
    error(errSyntaxError, getPos(), "Decompression bomb in flate stream");
    endOfBlock = eof = gTrue;
    remain = 0;
  }
```

I've tested this in FilterHelper by adding a max limit to the number of bytes `gzuncompress` can parse on line 254 as an argument. This will throw an Exception then, rather than going to a fatal error. I don't know what the optimum values would be, but, this could be useful. 

Potentially it'd be something to add to the Config class, as an optional memory limit. I wanted to get the dev team's thoughts first before submitting a PR though. 

The PDF attached is based on that posted by yaspr in issue #104 
[document_with_text_and_png_image_4pages.pdf](https://github.com/smalot/pdfparser/files/7389789/document_with_text_and_png_image_4pages.pdf)

_Originally posted by @b3n-l in https://github.com/smalot/pdfparser/issues/104#issuecomment-948669261_

PR #476, opened 2021-10-29
Added decodeMemoryLimit to Config to avoid memory leaks.

b3n-l:
Per issue #475, we have potential for a memory leak, especially with Flate compressed data. 

This PR adds an optional memory limit in the config, that can be used by any filter. 

By default, the limit is set to 0 or unlimited. 

This may not be the best way to propagate the setting, but I wanted it to be a configurable option, rather than hard-coded.

### DIFF
--- a/src/Smalot/PdfParser/RawData/FilterHelper.php
+++ b/src/Smalot/PdfParser/RawData/FilterHelper.php
@@ -56,7 +56,7 @@ class FilterHelper
      *
      * @throws Exception if a certain decode function is not implemented yet
      */
-    public function decodeFilter(string $filter, string $data, $decodeMemoryLimit): string
+    public function decodeFilter(string $filter, string $data, $decodeMemoryLimit=0): string
     {
         switch ($filter) {
             case 'ASCIIHexDecode':
--- a/src/Smalot/PdfParser/RawData/FilterHelper.php
+++ b/src/Smalot/PdfParser/RawData/FilterHelper.php
@@ -56,7 +56,7 @@ class FilterHelper
      *
      * @throws Exception if a certain decode function is not implemented yet
      */
-    public function decodeFilter(string $filter, string $data): string
+    public function decodeFilter(string $filter, string $data, $decodeMemoryLimit): string
     {
         switch ($filter) {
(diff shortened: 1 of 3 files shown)


k00ni:
I am busy in the next 2 weeks, but will try to come back to you until end of November.

A few remarks:
* Does this PR closes #475?
* If you add new properties/parameters, please use a type. For instance in `Config` for `private $decodeMemoryLimit = 0;`.

b3n-l:
To confirm, this fixes the issue described in #475. 

I made it optional, so that it didn't break compatibility if people are expecting to be able to deflate huge files. 

I will update the PR with appropriate types.

b3n-l:
Just a quick note, 

I've investigated why it's failing the tests and it's because there's now the wrong number of arguments in some of the test calls. I'd appreciate input into the best format/structure for adding these arguments, as not all decompression methods will use/need them. 

See `decodeFilter` as an example.

k00ni:
> I'd appreciate input into the best format/structure for adding these arguments, as not all decompression methods will use/need them.

There is no official statement. In my opinion the most important parameters should be first. Optional parameters always at the end.

If that doesn't help, can you please more clear what you mean?

b3n-l:
I've updated the PR, and simplified it slightly too, I'm now happy with it. 

I think this should fix the tests failing too.

src/Smalot/PdfParser/RawData/FilterHelper.php
```diff
@@ -56,7 +56,7 @@ class FilterHelper
      *
      * @throws Exception if a certain decode function is not implemented yet
      */
-    public function decodeFilter(string $filter, string $data): string
+    public function decodeFilter(string $filter, string $data, $decodeMemoryLimit=0): string
```
k00ni, line 59:
```suggestion
    public function decodeFilter(string $filter, string $data, int $decodeMemoryLimit = 0): string
```

k00ni:
I took the liberty to fix remaining coding style issues. Any remarks @b3n-l?

If there are no objections I will merge it soon.

b3n-l:
Thanks for that, all looks good to me.

PR #479, opened 2021-11-11
Fixes #478 (/Index problem)

### LINKED ISSUE #478: Fix for /Index problem
After editing the document description of a PDF file, I've had some cases where getDetails() stopped reporting the /Info data to the Document class. I found the problem in the incomplete handling of the /Index area. So far it was assumed that the XRef is continuous (only one entry in the area /Index). In my cases there was more than one entry, e.g. "/Index[162 1 165 4]", which means 1 object starting at number 162 and 4 objects starting at number 165. I eliminated this problem in the file "RawDataParser.php", but I have no idea how I can upload this file.

This is the function where I fixed the problem: 

    protected function decodeXrefStream(string $pdfData, int $startxref, array $xref = []): array
    {
        // try to read Cross-Reference Stream
        $xrefobj = $this->getRawObject($pdfData, $startxref);
        $xrefcrs = $this->getIndirectObject($pdfData, $xref, $xrefobj[1], $startxref, true);
        if (!isset($xref['trailer']) || empty($xref['trailer'])) {
            // get only the last updated version
            $xref['trailer'] = [];
            $filltrailer = true;
        } else {
            $filltrailer = false;
        }
        if (!isset($xref['xref'])) {
            $xref['xref'] = [];
        }
        $valid_crs = false;
        $columns = 0;
        $predictor = null;
        $sarr = $xrefcrs[0][1];
        if (!\is_array($sarr)) {
            $sarr = [];
        }

        $wb = [];

        foreach ($sarr as $k => $v) {
            if (
                ('/' == $v[0])
                && ('Type' == $v[1])
                && (
                    isset($sarr[($k + 1)])
                    && '/' == $sarr[($k + 1)][0]
                    && 'XRef' == $sarr[($k + 1)][1]
                )
            ) {
                $valid_crs = true;
            } elseif (('/' == $v[0]) && ('Index' == $v[1]) && (isset($sarr[($k + 1)]))) {
                // list of: first object number in the subsection / number of objects
				$index_blocks = [];
				for($m = 0; $m < count($sarr[($k + 1)][1]); $m += 2) {
					$index_blocks[] = [$sarr[($k + 1)][1][$m][1], $sarr[($k + 1)][1][$m + 1][1]];
				}
            } elseif (('/' == $v[0]) && ('Prev' == $v[1]) && (isset($sarr[($k + 1)]) && ('numeric' == $sarr[($k + 1)][0]))) {
                // get previous xref offset
                $prevxref = (int) ($sarr[($k + 1)][1]);
            } elseif (('/' == $v[0]) && ('W' == $v[1]) && (isset($sarr[($k + 1)]))) {
                // number of bytes (in the decoded stream) of the corresponding field
                $wb[0] = (int) ($sarr[($k + 1)][1][0][1]);
                $wb[1] = (int) ($sarr[($k + 1)][1][1][1]);
                $wb[2] = (int) ($sarr[($k + 1)][1][2][1]);
            } elseif (('/' == $v[0]) && ('DecodeParms' == $v[1]) && (isset($sarr[($k + 1)][1]))) {
                $decpar = $sarr[($k + 1)][1];
                foreach ($decpar as $kdc => $vdc) {
                    if (
                        '/' == $vdc[0]
                        && 'Columns' == $vdc[1]
                        && (
                            isset($decpar[($kdc + 1)])
                            && 'numeric' == $decpar[($kdc + 1)][0]
                        )
                    ) {
                        $columns = (int) ($decpar[($kdc + 1)][1]);
                    } elseif (
                        '/' == $vdc[0]
                        && 'Predictor' == $vdc[1]
                        && (
                            isset($decpar[($kdc + 1)])
                            && 'numeric' == $decpar[($kdc + 1)][0]
                        )
                    ) {
                        $predictor = (int) ($decpar[($kdc + 1)][1]);
                    }
                }
            } elseif ($filltrailer) {
                if (('/' == $v[0]) && ('Size' == $v[1]) && (isset($sarr[($k + 1)]) && ('numeric' == $sarr[($k + 1)][0]))) {
                    $xref['trailer']['size'] = $sarr[($k + 1)][1];
                } elseif (('/' == $v[0]) && ('Root' == $v[1]) && (isset($sarr[($k + 1)]) && ('objref' == $sarr[($k + 1)][0]))) {
                    $xref['trailer']['root'] = $sarr[($k + 1)][1];
                } elseif (('/' == $v[0]) && ('Info' == $v[1]) && (isset($sarr[($k + 1)]) && ('objref' == $sarr[($k + 1)][0]))) {
                    $xref['trailer']['info'] = $sarr[($k + 1)][1];
                } elseif (('/' == $v[0]) && ('Encrypt' == $v[1]) && (isset($sarr[($k + 1)]) && ('objref' == $sarr[($k + 1)][0]))) {
                    $xref['trailer']['encrypt'] = $sarr[($k + 1)][1];
                } elseif (('/' == $v[0]) && ('ID' == $v[1]) && (isset($sarr[($k + 1)]))) {
                    $xref['trailer']['id'] = [];
                    $xref['trailer']['id'][0] = $sarr[($k + 1)][1][0][1];
                    $xref['trailer']['id'][1] = $sarr[($k + 1)][1][1][1];
                }
            }
        }

        // decode data
        if ($valid_crs && isset($xrefcrs[1][3][0])) {
            if (null !== $predictor) {
                // number of bytes in a row
                $rowlen = ($columns + 1);
                // convert the stream into an array of integers
                $sdata = unpack('C*', $xrefcrs[1][3][0]);
                // split the rows
                $sdata = array_chunk($sdata, $rowlen);

                // initialize decoded array
                $ddata = [];
                // initialize first row with zeros
                $prev_row = array_fill(0, $rowlen, 0);
                // for each row apply PNG unpredictor
                foreach ($sdata as $k => $row) {
                    // initialize new row
                    $ddata[$k] = [];
                    // get PNG predictor value
                    $predictor = (10 + $row[0]);
                    // for each byte on the row
                    for ($i = 1; $i <= $columns; ++$i) {
                        // new index
                        $j = ($i - 1);
                        $row_up = $prev_row[$j];
                        if (1 == $i) {
                            $row_left = 0;
                            $row_upleft = 0;
                        } else {
                            $row_left = $row[($i - 1)];
                            $row_upleft = $prev_row[($j - 1)];
                        }
                        switch ($predictor) {
                            case 10:  // PNG prediction (on encoding, PNG None on all rows)
                                $ddata[$k][$j] = $row[$i];
                                break;

                            case 11:  // PNG prediction (on encoding, PNG Sub on all rows)
                                $ddata[$k][$j] = (($row[$i] + $row_left) & 0xff);
                                break;

                            case 12:  // PNG prediction (on encoding, PNG Up on all rows)
                                $ddata[$k][$j] = (($row[$i] + $row_up) & 0xff);
                                break;

                            case 13:  // PNG prediction (on encoding, PNG Average on all rows)
                                $ddata[$k][$j] = (($row[$i] + (($row_left + $row_up) / 2)) & 0xff);
                                break;

                            case 14:  // PNG prediction (on encoding, PNG Paeth on all rows)
                                // initial estimate
                                $p = ($row_left + $row_up - $row_upleft);
                                // distances
                                $pa = abs($p - $row_left);
                                $pb = abs($p - $row_up);
                                $pc = abs($p - $row_upleft);
                                $pmin = min($pa, $pb, $pc);
                                // return minimum distance
                                switch ($pmin) {
                                    case $pa:
                                        $ddata[$k][$j] = (($row[$i] + $row_left) & 0xff);
                                        break;

                                    case $pb:
                                        $ddata[$k][$j] = (($row[$i] + $row_up) & 0xff);
                                        break;

                                    case $pc:
                                        $ddata[$k][$j] = (($row[$i] + $row_upleft) & 0xff);
                                        break;
                                }
                                break;

                            default:  // PNG prediction (on encoding, PNG optimum)
                                throw new Exception('Unknown PNG predictor: '.$predictor);
                        }
                    }
                    $prev_row = $ddata[$k];
                } // end for each row
                // complete decoding
            } else {
                // number of bytes in a row
                $rowlen = array_sum($wb);
                // convert the stream into an array of integers
                $sdata = unpack('C*', $xrefcrs[1][3][0]);
                // split the rows
                $ddata = array_chunk($sdata, $rowlen);
            }

            $sdata = [];

            // for every row
            foreach ($ddata as $k => $row) {
                // initialize new row
                $sdata[$k] = [0, 0, 0];
                if (0 == $wb[0]) {
                    // default type field
                    $sdata[$k][0] = 1;
                }
                $i = 0; // count bytes in the row
                // for every column
                for ($c = 0; $c < 3; ++$c) {
                    // for every byte on the column
                    for ($b = 0; $b < $wb[$c]; ++$b) {
                        if (isset($row[$i])) {
                            $sdata[$k][$c] += ($row[$i] << (($wb[$c] - 1 - $b) * 8));
                        }
                        ++$i;
                    }
                }
            }

            // fill xref
			if (isset($index_blocks)) {
				$obj_num = $index_blocks[0][0];
            } else {
                $obj_num = 0;
            }
            foreach ($sdata as $k => $row) {
                switch ($row[0]) {
                    case 0:  // (f) linked list of free objects
                            break;

                    case 1:  // (n) objects that are in use but are not compressed
                            // create unique object index: [object number]_[generation number]
                            $index = $obj_num.'_'.$row[2];
                            // check if object already exist
                            if (!isset($xref['xref'][$index])) {
                                // store object offset position
                                $xref['xref'][$index] = $row[1];
                            }
                            break;

                    case 2:  // compressed objects
                            // $row[1] = object number of the object stream in which this object is stored
                            // $row[2] = index of this object within the object stream
                            $index = $row[1].'_0_'.$row[2];
                            $xref['xref'][$index] = -1;
                            break;

                    default:  // null objects
                            break;
                }
                ++$obj_num;
				if (isset($index_blocks)) {
					if (--$index_blocks[0][1] == 0) {
						array_shift($index_blocks);
						if (count($index_blocks) > 0) {
							$obj_num = $index_blocks[0][0];
						} else {
							unset($index_blocks);
						}
					}
				}
            }
        } // end decoding data
        if (isset($prevxref)) {
            // get previous xref
            $xref = $this->getXrefData($pdfData, $prevxref, $xref);
        }

        return $xref;
    }

yasheena:
This is the requested pull request from issue #478. Sorry, I was not yet familiar with pull requests.

After editing the document description of a PDF file, I've had some cases where getDetails() stopped reporting the /Info data to the Document class. I found the problem in the incomplete handling of the /Index area. So far it was assumed that the XRef is continuous (only one entry in the area /Index). In my cases there was more than one entry, e.g. "/Index[162 1 165 4]", which means 1 object starting at number 162 and 4 objects starting at number 165. I eliminated this problem in the file "RawDataParser.php" with this pull request.


tests/Integration/RawData/RawDataParserTest.php
```diff
@@ -119,4 +119,23 @@ public function testDecodeObjectHeaderIssue405()
 
         $this->assertStringContainsString('Bug fix: PR #405', $pages[0]->getText());
     }
+
+    /**
+     * Tests buggy behavior of decodeXrefStream.
+     *
+     * When PDF has more than one entry in the /Index area (for example by changing the document description), only the first entry is used.
+     *
+     * @see https://github.com/smalot/pdfparser/pull/479
```
k00ni, line 128: What exactly would happen in your test if the fix is not used?

yasheena: If the fix is not used the array returned by getDetails() contains only the entry with the key 'Pages'. All other entries like 'Author', 'Creator', 'Title', 'Subject' (which come from the 'Info' object) are not listed, because the 'Info' object gets a wrong object id during parsing the data into the xref structure. So the object id listed at the /Info entry is not valid and the data of the info object cannot be loaded during executing Document::buildDetails().

k00ni: * [ ] Can you add this to the function header too? It helps a lot.

* [ ] Also, can you check for all expected keys not just `Subject`?

yasheena: I added the rest of the keys, but they available only for this PDF.Other PDFs may have more or less keys.

Question: Several tests fail, but it seems that my code isn't causing it. Is that ok?

k00ni: > Question: Several tests fail, but it seems that my code isn't causing it. Is that ok?

No, unfortunately not.

yasheena:
I checked the failed test: My fix also fixes the issue #331. The test was written to fail if something changes in the behavior related to issue # 331. I have adapted and commented on the test accordingly.

tests/Integration/PageTest.php
```diff
@@ -484,13 +484,23 @@ public function testGetPages()
...
+        // $this->assertCount(2, $pages);
+
+        // The above problem is fixed by the pull request of the issue #479.
+        // The original Issue331.pdf was modified so for the updated version (actual
+        // version) a new xref was added and now the valid /Index has the following value:
+        //    [1 1 3 1 7 1 175 1 178 1 219 2]
+        // This means, that there a 6 pairs containing the values for 'first object id'
+        // and 'number of objects'. Till now only the first entry was used and so the
+        // objects of all following entries gots a wrong id.
+        // By the fix of issue #479 now the expected number of pages is counted.
```
k00ni, line 501:
```suggestion
        /*
         * The above problem is fixed by the pull request of the issue #479.
         * The original Issue331.pdf was modified so for the updated version (actual
         * version) a new xref was added and now the valid /Index has the following value:
         *    [1 1 3 1 7 1 175 1 178 1 219 2]
         * This means, that there a 6 pairs containing the values for 'first object id'
         * and 'number of objects'. Till now only the first entry was used and so the
         * objects of all following entries gots a wrong id.
         * By the fix of issue #479 now the expected number of pages is counted.
         */
```

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -463,6 +466,16 @@ protected function decodeXrefStream(string $pdfData, int $startxref, array $xref
...
+                if (isset($index_blocks)) {
+                    if (0 == --$index_blocks[0][1]) {
+                        array_shift($index_blocks);
+                        if (0 < count($index_blocks)) {
+                            $obj_num = $index_blocks[0][0];
+                        } else {
+                            unset($index_blocks);
+                        }
+                    }
+                }
```
k00ni, line 478: 3 if-clauses inside each other is a code smell in my opinion.

* [ ] Please add comment which explains what this part does.
* [ ] Reduce number of if-clauses. For instance move some operations in front of the first clause and combine all their conditions?
* [ ] `if (0 == --$index_blocks[0][1]) {` is a bit problematic, because one might overlook the `--` before `$index_blocks` and assumes a different value. Such operations should be split up.

k00ni (changes requested):
First of all, thanks for your quick responses and good progression here! Just a few remarks.

src/Smalot/PdfParser/RawData/RawDataParser.php
yasheena: I added comments and splitted the code with the --.
But I do not know, how I should reduce the if clauses: The first is neccessary to avoid problems, if $index_blocks does not exists. The second checks the downcounting and only if the count is 0 there must be a check with the third if whether or not there are additional entries.

k00ni:
@Connum @izabala @j0k3r and others, it would be good to get a second opinion on this. For me it looks good.

j0k3r (approved):
Looks ok, just ran `php-cs-fixer` to fix the build

Connum:
Didn't have much time to look into this, but it seems fine to me - congrats for fixing this @yasheena
Only comment I have is that "The above problem" in the comment in `tests/Integration/PageTest.php
` doesn't relate to anything after the comment block above has been removed, so the wording should be adapted.

tests/Integration/PageTest.php
```diff
@@ -484,12 +484,17 @@ public function testGetPages()
...
         $pages = $document->getPages();
 
-        // This should actually be 3 pages, but as long as the cause for issue #331
-        // has not been found and the issue is not fixed, we'll settle for 2 here.
-        // We still test for the count, so in case the bug should be fixed
-        // unknowingly, we don't forget to resolve the issue as well and make sure
-        // this assertion is present.
-        $this->assertCount(2, $pages);
+        /*
+         * The problem of issue 331 is fixed by the pull request of the issue #479.
```
k00ni, line 488:
```suggestion
         * The problem of issue #331 is fixed by the pull request of the issue #479.
```

k00ni (approved):
Thank you for your latest changes. It looks good now.

As always, we will keep this open for a few days so the community has a chance to comment, before we merge it.

Reqrefusion:
Thanks. I've been waiting for this for a long time.

PR #485, opened 2021-11-18
fix for Issue #482

### LINKED ISSUE #482: Undefined array key when calling getDataTm()
I have a simple code like this:

```php
require_once 'vendor/autoload.php';

$file = 'path/to/file.pdf';

$pdf = $parser->parseFile($file);
$pages = $pdf->getPages();

$firstPage = $pages[0];

// Get data matrix
$firstPageDm = $firstPage->getDataTm();

echo '<pre>';
var_dump($firstPageDm);
```

And got this error message:

Error

Is this caused by wrong array access?
From what I found in the code, is that mean want to get the last array element?

I use version ^2.0, the code in line 675.

https://github.com/smalot/pdfparser/blob/ce6aa779dd67fadf3887dace6f31ea70882dba73/src/Smalot/PdfParser/Page.php#L584

Any solution?

Btw thanks for this awesome library.

izabala:
Fixes #482. The 'td' command has a unexpected "\n" between the 2 arguments (that are numbers).


src/Smalot/PdfParser/Page.php
```diff
@@ -711,7 +711,8 @@ public function getDataTm(array $dataCommands = null): array
                  * current line by tx, ty.
                  */
                 case 'Td':
-                    $coord = explode(' ', $command['c']);
+                    $commandLine = str_replace("\n", ' ', $command['c']);
```
k00ni, line 714: Is it only `\n` or could there be other newline characters that this point?

izabala: Ok... This will be a good discussion.
This is really an unexpected \n between the operands (numbers) of the Td operator. 
In the [Pdf Specification](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf) it defines the Td operation in page 249 table 108. Define the operands (numbers that could be integers or floats) in one line, separated by one space.  If you have a look at the examples shown in the same specification it looks like: 
```
BT 
    40 50 Td
    (ABC) Tj
ET
```
or 
```
BT
    /F1 12 Tf
    72 712 Td
   (A stream with an indirect length) Tj
ET
```
Both are examples from the PDF Specification.
So this is the first time I saw a "strange" character between the both operands.   For me, (and I am not really an expert on this, I am just using the specification and my experience, but I could be wrong), the file that was sent in Issue#482 is a wrong formatted PDF file. But this is just Opinion. 

So If I am right, should we change all the code where should be operands separated with one space like: 'Td, TD, Tm, " ' ?(By the way the operator ' " ' has three operands aw ac string, is not an extra " that i put before).
Should we use regex in all the cases to take out whatever is put there for an Pdf writer or editor that creates wrong formatted PDF files? (with the extra processor time that using regex put in every time we used it?).  How far we should go? Do we still make this fix, if the Pdf file is wrong format? (As I said before, I made a research and using my experience, for giving this opinion,  but I could be wrong).

So, I decided to put an extra line, replacing just the "\n" for a " ", so for the Issue #482 could be fixed, and if there is no "\n" in the line, it do nothing.  But maybe the call is to do nothing, and answer that the PDF file is wrong formatted.  But I am not the right person to make that call, I just give my thoughts. 

Does this help?

ngekoding: Sorry guys, I want to confirmed about what @izabala said above.

> the file that was sent in Issue#482 is a wrong formatted PDF file.

I generated the PDF document with Ms. Word 365, so maybe it is not the problem by the PDF.

izabala: @k00ni If this is a wrong formatted PDF, do we need to close this Pull request (so no changes in the code)  and close the Issue#482?

k00ni: > so maybe it is not the problem by the PDF.

There are a couple of programs which do not generate 100% compatible PDFs. It is sometimes hard to see, if PDFParser lacks functionality, has a bug or if just invalid characters were used.

In general PDFParser should follow the specification as close as possible. But on the other hand it should be forgiving to some extent, to enable users a wider range of PDFs to use. 

Could #482 be a case in which we can be "forgiving" and allow invalid characters at this point? I am torn, but tend to **not** allow it, because it looks like a hack to get one PDF working. In other PDFs this might lead undefined behavior of PDFParser.

Are there known adaptions MS Word does which might be relevant here? We could think about a `Config` option, which when enabled, handles those.

What do you think?

izabala: Hi, the problem here is How far to go? It will be just "\n" character, Do we have to use regex just for something that occurs if the file is just wrong formatted? In which PDF Operations (commands) do we have to do it... just in "Td" or for all?

I made the "FIX" because I saw that was just 1 line to change... If we need to go farther, I will need extra time, that I dont have at this very moment. 

By the way, I don'tknow what do you mean with: "Are there known adaptions MS Word does which might be relevant here?".  Which adaptations? How are them relevant for this?  (I really don't know them).

Thanks you all for your involvement and opinion on this one!!!

k00ni: >  the problem here is How far to go?

Yes, good question. Its good that we have a chance here to elaborate on it a bit. Your feedback helped to better understand the situation. I am not a fan of hacks, therefore its good that we have no need to merge a pull request which fixes 1 PDF.

PDFParser is the kind of project which tries to satisfy a specification, but has to deal with quirky PDF also. Discussing the magnitude of changes before making them is one of the things I like about this Community. Its always a trade off between time, effort and goals. We are doing a fine job here.

Closing this, because I think we decided that we won't fix something which isn't broken. Thank you guys.

PR #486, opened 2021-11-19
Replace usage of str_starts_with in Page.php::isFpdf

### LINKED ISSUE #483: PHP 8 seems to be required for v2.0.0
Rather than just supported, it seems PHP 8 is a _requirement_ for v2.0.0 (and for 1.2.0-RC2) based on the usage of `str_starts_with()` here: https://github.com/smalot/pdfparser/blob/f5fa4d3e6415cbf558d1e9c46804ab79ebe2574f/src/Smalot/PdfParser/Page.php#L239

As per PHP docs, that function has only existed since PHP 8 (https://www.php.net/manual/en/function.str-starts-with.php)

So I believe the PHP version requirement in `composer.json` needs to be upped to 8.0 rather than 7.1

k00ni:
Fixes #483 reported by @dpanzer.

Problem was usage of `str_starts_with` which is only available in PHP 8.

### DIFF
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -236,7 +236,7 @@ class Page extends PDFObject
     {
         if (\array_key_exists('Producer', $this->document->getDetails()) &&
             \is_string($this->document->getDetails()['Producer']) &&
-            0 === strncmp($this->document->getDetails()['Producer'], 'FPDF', \strlen('FPDF'))) {
+            0 === strncmp($this->document->getDetails()['Producer'], 'FPDF', 4)) {
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


j0k3r (approved):
Which means that line isn't covered by tests.

k00ni:
> Which means that line isn't covered by tests.

Good point! Based on Scrutinizer this line is covered: https://scrutinizer-ci.com/g/smalot/pdfparser/code-structure/master/code-coverage/src/Smalot/PdfParser/Page.php

But our PHP 7.x tests should have failed. 

Any idea?

k00ni:
In our CI it install `symfony/polyfill-php80` before running tests on PHP 7.1: https://github.com/smalot/pdfparser/runs/4261468787?check_suite_focus=true#step:5:57

Thats why we didn't saw it, because the polyfill provided the function.

PR #493, opened 2021-11-29
added function to parse base64 encoded PDFs

granjero:
This function instead of receiving a filename like parseFile it receives a base 64 enconded PDF.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -38,3 +38,20 @@ To check coding styles run:
 To run a static code analysis use:
 
 > make run-phpstan
+
+## Base64 encoded PDFs
+
+If working with [Base64](https://en.wikipedia.org/wiki/Base64) encoded PDFs you might want to parse the PDF without saving the file on disk. This sample will parse the Base64 encoded PDF and extract text from each page.
+
+```php
+<?php
+// Include Composer autoloader if not already done.
+include "vendor/autoload.php";
+
+// Parse Base64 encoded PDF string and build necessary objects.
+$parser = new \Smalot\PdfParser\Parser();
+$pdf = $parser->parseContent(base64_decode($base64PDF));
+
+$text = $pdf->getText();
+echo $text;
+```
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# PdfParser
+# PdfParser #
 
 Pdf Parser, a standalone PHP library, provides various tools to extract data from a PDF file.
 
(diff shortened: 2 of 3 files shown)


k00ni (changes requested):
Hello @granjero, thank you for you pull request.

Your changes seem too small and can be implemented with just 1 line of code, like

```php
$decoded = (new Parser())->parseContent(base64_decode($base64EncodedPdf));
```

What is the benefit having this as a new function?

granjero:
Hello @k00ni, thank you for taking the time in reviewing the request.
I see that your approach works. And does not require the function I proposed. 
Nevertheless I think that the function I submitted may be "user friendlier".

I found this library while working with an API that returns a PDF in a base64 string. I could not find any mention in the docs for base64 strings. At first I thought I had to save the file and then read it with parseFile() but wondered if i could do it in the fly. So I came with that solution, wrote the function and it worked.  

If you think it's worth it I can write the tests and meet the other labeled requirements.

Thank you.

jm

PS: Forgive my rusty english.

k00ni:
Thanks for getting back to me quickly.

> If you think it's worth it I can write the tests and meet the other labeled requirements.

My **suggestion** would be, instead of adding a new function, to extend `README.md` with a new section which contains the case you described and some example code how to solve it. 

What do you think? If you agree, could you propose such a change?

granjero:
Ok, so you are saying something like this at the end of [DEVELOPER.md](https://github.com/smalot/pdfparser/blob/master/DEVELOPER.md) or in the [documentation](https://www.pdfparser.org/documentation)



## Base64 encoded PDFs

If working with base64 encoded PDFs you might want to parse the pdf without saving the file on disk.
This sample will parse the base 64 encoded pdf and extract text from each page.

```php
<?php
// Include Composer autoloader if not already done.
include 'vendor/autoload.php';

// Parse pdf base64 enconded string and build necessary objects.
$pdf    = $parser->parseContent(base64_decode($base64PDF));

$text = $pdf->getText();
echo $text;
?>
```

k00ni:
Good suggestion. I refined it a bit.

We don't have control over pdfparser.org, therefore DEVELOPER.md is the way to go. 

* [ ] Please add your changes to DEVELOPER.md.
* [ ] Please add a sentence to README.md which points to Base64 section in DEVELOPER.md. 

Suggestion:
```
**For developers:** Please read DEVELOPER.md for more information about integration, usage and local development of the PDFParser library. There you will also find information about Base64 PDFs.
```

granjero:
Ok. I hope I did it right. 
@k00ni please tell me if it ok.
Thanks!

k00ni:
Its looks good! But I dont have the time right now to look into it further. WIll do it hope fully in the first two weeks of January. Bear with me.

k00ni (approved):
I reverted many of your changes in README.md, only kept reference to DEVERLOPER.md section so this PR is only about Base64 information. 

We plan to update documentation overall, which will include some of your changes in README.md (see #478). You are welcome to contribute.

Is this PR OK for you @granjero?

granjero:
Sorry, I was afk.
I'm glad I could contribute!
Keep it going!

granjero (commented):
Line 47 Base63 instead of Base64

granjero:
@k00ni 
There is a typo in the new [README.md](https://github.com/smalot/pdfparser/blob/master/README.md)

Line 47 -> Base63 instead of Base64

PR #500, opened 2021-12-25
Fix encoding for encoding dictionary without Type item.

likemusic:
image

In PDF-file in Font internal encryption point to dictionary without Type item, therefore encryption treated not like encryption but just like PDFObject and as result text incorrectly decoded (because both BaseEncoding item and differences array item ignored). 

This PR fixes it.

Also has been deleted unnecessary unicode string decode test for file with WinAnsiEncoding text encoding.

Also deleted for $unicode passed by reference parameters in Font class because seems that it have no sense (not sure). 

Potentially this PR can fix some of the many other opened issues related to incorrect result text encoding.


composer.json
```diff
@@ -17,7 +17,8 @@
     "require": {
         "php": ">=7.1",
         "symfony/polyfill-mbstring": "^1.18",
-        "ext-zlib": "*"
+        "ext-zlib": "*",
+        "ext-iconv": "*"
```
k00ni, line 21: Why did you add this line?

src/Smalot/PdfParser/Font.php
```diff
@@ -57,6 +57,9 @@ class Font extends PDFObject
      */
     private static $uchrCache = [];
 
+    /** @var Encoding */
+    private $initializedEncodingByPdfObject;
```
k00ni, line 61:
```suggestion
    private Encoding $initializedEncodingByPdfObject;
```

src/Smalot/PdfParser/Font.php
```diff
@@ -381,86 +384,151 @@ public function decodeText(array $commands): string
...
     {
         if ($this->has('ToUnicode')) {
-            $bytes = $this->tableSizes['from'];
+            return $this->decodeContentToUnicode($text);
+        }
 
-            if ($bytes) {
-                $result = '';
-                $length = \strlen($text);
+        if ($this->has('Encoding')) {
```
k00ni, line 390: Please use if-elseif here.

k00ni (changes requested):
@likemusic First of all, happy new year and thank you for your pull request!

Please check my questions and remarks below. 

* [x] You added a lot of code. Please describe inline (e.g. function headers) what each code part does and why you added/changed it.
* [ ] Add tests to cover your changes.
* [ ] Explain `Also deleted for $unicode passed by reference parameters in Font class because seems that it have no sense (not sure).`

Let me know if you need help.

likemusic:
@k00ni Happy New Year to you, too!

And thank you too for active maintenance this helpfull project!

likemusic:
@k00ni 
> Please describe inline (e.g. function headers) what each code part does and why you added/changed it.

I must add comment like you do with you reamarks and question and I shouldn't commit any comment to source code. Did I understand you correctly?

likemusic:
> Add tests to cover your changes.

@k00ni Unfotrunaly I can't commit pdf-file bacause it's a bank statement :) But I'll try to find suitable pdf-file in some of opened issues related to incorect output encodings. By the way I will test those issues and mark wich one this PR will fix.

likemusic:
@k00ni I want to mark encoding related issued by `encoding` label to have quick filter and mayby to fix all of them in future.

I want to add #encoding tag in comments for issues related to encoding and point you (for adding label `encoding` to issue).

Is it good idea? Can I do this?

composer.json
likemusic: Because `iconv()` method from iconv extension used during converting pdf-file encoding to utf-8 in [Font class] (https://github.com/smalot/pdfparser/pull/500/files#diff-27bc594b24bd5e2779e8d81ee79810d0ffda03f200f0d53be007a44a9d2cb2deR409)

src/Smalot/PdfParser/Font.php
likemusic: I think if-return variant has better readability, hasn't it?

I googled a bit and found that some developer prefer if-elseif to have just one point of return statement. But I think it is not a good reason for this case because there is no any common code for some if-branches or all-branches common post-processing.

 Also I also looked through how elseif already used in project and not found where it used with return operator.

Also there is already exists using if-return some times:
- [Document::getObjectsByType()](https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/Document.php#L188) and [Docuemnt::getPages()](https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/Document.php#L228);
- [Page::getForns()](https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/Page.php#L58);

Therefore I think it would be better to keep if-return statements as is.

Do you agree with me?

src/Smalot/PdfParser/Font.php
likemusic: PHP supports type definitions for class properties just from [version 7.4](https://www.php.net/manual/en/migration74.new-features.php). But currently this project must support php strating from version 7.1 (by composer restriction ("php": ">=7.1")).

And at the moment project hasn't any typed properties.

Therefore I think It would be better keep it as is.

And maybe add issue with task to migrate to php v7.4 and add propetries types everywere in sources.

k00ni: Ah yes, my bad.

src/Smalot/PdfParser/Font.php
k00ni: Good points, I agree.

likemusic:
> What do you think about this: You post in related issues a short hint and I will add the label for you. I already created "encoding issues", which is more clear. You agree?

@k00ni It would be great! I meant the same, but my English is not as good as I would like :)

At the moment I sterted and hangs on the investigeting for first issue, not related to but fixed in this PR.

likemusic:
> Please add information, links, thoughts etc. next to your new code (via further commits).

@k00ni I made small refactoring and add coments for not unobvious logic and methods names.

likemusic:
> Can you create a minimal string which represents a "broken" PDF? You then pass it to the Parser and see how it reacts.

@k00ni I don't like this idea because It allows create combinations not compartible with real PDF-file format. Therefore real pdf file would be preferable, bacause it can be tested visually by Adobe Acrobat Reader.

Frst I will try to find suitable file in opened encoding-related issues. If I have no luck I'll try to "rebuld" manually my bank statement's pdf to small suitable file with one string. If it would be too difficult I'll make synthetic test as you suggested.

src/Smalot/PdfParser/Font.php
```diff
@@ -57,6 +58,17 @@ class Font extends PDFObject
      */
     private static $uchrCache = [];
 
+    /**
+     * In some pdf-files (@see https://github.com/smalot/pdfparser/pull/500) encoding could be referenced by object id
+     * but object itself not contains `/Type /Encoding` in its dictionary. That objects wouldn't be initialized as
```
k00ni, line 63:
```suggestion
     * but object itself not contains `/Type /Encoding` in its dictionary. These objects wouldn't be initialized as
```

src/Smalot/PdfParser/Font.php
```diff
@@ -381,86 +393,182 @@ public function decodeText(array $commands): string
...
     }
+
+    /**
+     * Decode content by any type of Encoding (dictionary's item) instance.
+     */
+    private function decodeContentByEncoding(string $text): ?string
+    {
+        $encoding = $this->get('Encoding');
+
+        // When Encoding referenced by object id but object itself not contains `/Type /Encoding` in it's dictionary.
```
k00ni, line 465:
```suggestion
        // When Encoding referenced by object id but object itself does not contain `/Type /Encoding` in it's dictionary.
```

src/Smalot/PdfParser/Font.php
```diff
@@ -381,86 +393,182 @@ public function decodeText(array $commands): string
...
+        if ($encoding instanceof Element) { //todo: ElementString class must by used?
+            return $this->decodeContentByEncodingElement($text, $encoding);
+        }
+
+        // Encoding has unintended type.
+        $encodingClassName = \get_class($encoding);
+        throw new LogicException("Unknown encoding instance type: {$encodingClassName}");
+    }
+
+    private function getInitializedEncodingByPdfObject(PDFObject $PDFObject): Encoding
```
k00ni, line 485: Why did you add this function? It is only used once.

src/Smalot/PdfParser/Font.php
```diff
@@ -381,86 +393,182 @@ public function decodeText(array $commands): string
...
+    {
+        if (!$this->initializedEncodingByPdfObject) {
+            $this->initializedEncodingByPdfObject = $this->createInitializedEncodingByPdfObject($PDFObject);
+        }
+
+        return $this->initializedEncodingByPdfObject;
+    }
+
+    /**
+     * Decode content when Encoding is instance of Encoding.
```
k00ni, line 495: This sentence does not make sense in my opinion. `$encoding` is always of type `Encoding`, isn't it?

src/Smalot/PdfParser/Font.php
```diff
@@ -381,86 +393,182 @@ public function decodeText(array $commands): string
...
+            $dec_av = hexdec(bin2hex($text[$i]));
+            $dec_ap = $encoding->translateChar($dec_av);
+            $result .= self::uchr($dec_ap ?? $dec_av);
+        }
+
+        return $result;
+    }
+
+    /**
+     * Decode content when Encoding is instance of Element.
```
k00ni, line 512: Same here, it is a little bit confusing. I get what you want to say, but you should reference the point, where this decision is made, at least.

src/Smalot/PdfParser/Font.php
```diff
@@ -381,86 +393,182 @@ public function decodeText(array $commands): string
...
+     */
+    private function createEncodingByPdfObject(PDFObject $PDFObject): Encoding
+    {
+        $document = $PDFObject->getDocument();
+        $header = $PDFObject->getHeader();
+        $content = $PDFObject->getContent();
+        $config = $PDFObject->getConfig();
+
+        return new Encoding($document, $header, $content, $config);
+    }
```
k00ni, line 573: That is a bit over engineered in my opinion. `createEncodingByPdfObject` and `createInitializedEncodingByPdfObject` are only used once and are basically shortcuts. What is the benefit of having them? Why not init related objects at the place where they are need?

k00ni:
Can you explain what you meant with

> Also deleted for $unicode passed by reference parameters in Font class because seems that it have no sense (not sure).

src/Smalot/PdfParser/Font.php
likemusic: > Why did you add this function?

Reason for this function is described in method where it used.

https://github.com/likemusic/pdfparser/blob/21c4797009c23979005fc6fb1ee358c2abd8dfb8/src/Smalot/PdfParser/Font.php#L465-L468

Should I add some comments for this method? What exactly should I add to the comment?

likemusic: > It is only used once.

Is there anything wrong with that?

src/Smalot/PdfParser/Font.php
likemusic: >  you should reference the point where this decision is made at least.

Cases for encoding types pointed in method decodeContentByEncoding() where all this methods used. All this types created during parsing file.

src/Smalot/PdfParser/Font.php
likemusic: > That is a bit over engineered in my opinion ... Why not init related objects at the place where they are needed?

I don't think so. This methods just implements "details" in "main" algorithm. This reduces cognitive load when read "main" algorithm logic.

If you implement difficult algorithm in just one method it could grow up to thousands of lines and it would be too difficult or even impossible to understand what method realy do, because you must scrool too long and during reading code sort in your mind which lines are really important for implemented algorithm and which is just noise with unimportant details implementation.

Imho, it's much better when "main" method has just 10 lines and 10 nested methods with 10 lines for each, then one method with has 100 lines of code.

Look at Font::decodeContent() method before (84 lines of code with many deep nested if/else/foreach):
https://github.com/smalot/pdfparser/blob/e056671d50b31d6320a26aa8fbfafd2c02e6bff1/src/Smalot/PdfParser/Font.php#L381-L465

and after this PR (15 lines of code with just 2 simple if-return statements):

https://github.com/likemusic/pdfparser/blob/21c4797009c23979005fc6fb1ee358c2abd8dfb8/src/Smalot/PdfParser/Font.php#L393-L408

I think it's obvious that extracting some line groups to separated methods made it much easier to understand the logic of the method.


Splitting to createEncodingByPdfObject() and createInitializedEncodingByPdfObject() made to show that created object could has different states (not initialized and initialized).


By the way, I think that using initialized and not initialized object is not good idea, because in this case after geting object you must always check his state and init it if requred. It's more easy to use objects when they ready for usage after just creating it by constructor. So I think it would be much better if initialized and not initialized objects would be presented by different type groups, where initialized inherits or aggregate not initialized type.

likemusic:
> Also deleted for $unicode passed by reference parameters in Font class because seems that it have no sense (not sure).

> Can you explain what you meant with

@k00ni I was saying this about extracted methods, because it has no sense.  This paramer in first cases not used at all:
https://github.com/smalot/pdfparser/blob/e056671d50b31d6320a26aa8fbfafd2c02e6bff1/src/Smalot/PdfParser/Font.php#L383

or ovewriten immideately internally in other case:
https://github.com/smalot/pdfparser/blob/e056671d50b31d6320a26aa8fbfafd2c02e6bff1/src/Smalot/PdfParser/Font.php#L426-L429

So I just keep `$unicode` parameter in public  decodeContent() for backward compatibility (as it was before) and this PR has no any effect for this parameter.

likemusic:
I made pdf-file manually and use it in added test - [issue500.pdf](https://github.com/smalot/pdfparser/files/7819207/issue500.pdf)

There is a differenses in Page::getText() results before and after PR:
image

How this file shown by [Sumatra PDF Reader](https://www.sumatrapdfreader.org/free-pdf-reader)
image

But unfortunately in Adobe Acrobat Reader this file shown with spaces :(
image

I'll try to fix it in future.

likemusic:
This is differences for original pdf-file which shown the same in all pdf-readers.
image

src/Smalot/PdfParser/Font.php
k00ni: > Should I add some comments for this method? What exactly should I add to the comment?

Its usually the way around: each function should be described in its function header. If one uses a function, then there could be more information about the why's, what's etc. Having a descriptive function header also benefits users, who use an IDE. They will get information when using a function beforehand.

> Is there anything wrong with that?

There is a good statement about usage of functions [here](https://www.cs.utah.edu/~zachary/computing/lessons/uces-10/uces-10/node11.html#:~:text=The%20first%20reason%20is%20reusability,program%2C%20which%20saves%20you%20work.&text=Another%20aspect%20of%20reusability%20is,different%20(and%20separate)%20programs.). It talks about reusability and abstraction. In the respective [code line](https://github.com/likemusic/pdfparser/blob/21c4797009c23979005fc6fb1ee358c2abd8dfb8/src/Smalot/PdfParser/Font.php#L467) you introduce a function which uses another introduced function which just initialize an instance. Combined with the fact that you only using them once, that is over engineered in my opinion. It suggests complexity/structures which are not needed.

Please combined them.

src/Smalot/PdfParser/Font.php
k00ni: I assumed it, but the sentence is weird because you say X is and instance of X. But you refrain to `$this->get('Encoding')`, don't you?

My suggestion would be:

> Decode content when $this->get('Encoding') is instance of Encoding.

Would you agree to that?

src/Smalot/PdfParser/Font.php
k00ni: I agree with the fact that this library has code parts which are not good structured and have smell.

But this fact doesn't support your suggested change. When I take the code of your functions and combine it, it boils down to:

```php
$encoding = new Encoding(
    $PDFObject->getDocument(), 
    $PDFObject->getHeader(), 
    $PDFObject->getContent(), 
    $PDFObject->getConfig()
);
$encoding->init();
```

Its size is small and everyone get its meaning quickly, in my opinion. If in doubt a comment might help.

---

If you still don't agree with me I can accept the fact that you want to keep these 2 functions. Your underlying approach is to apply a better structure to the library and improve code readability which I welcome and support. I don't want to risk loosing your commitment over a two functions :)

k00ni:
> But unfortunately in Adobe Acrobat Reader this file shown with spaces :(

There are some issues regarding (unnecessary) white spaces. If you wanna work on those, please create dedicated PR. In this PR focus only on the encoding please.

tests/Integration/FontTest.php
```diff
@@ -340,14 +340,41 @@ public function testDecodeText()
...
-        $this->assertEquals('æöü', $font->decodeText($commands));
+    /**
+     * Font could have indirect encoding without `/Type /Encoding`
+     * which would be instance of PDFObject class (but not Encoding or ElementString).
+     *
+     * @see https://github.com/smalot/pdfparser/pull/500
+     *
+     * @return void
+     *
+     * @throws \Exception
```
k00ni, line 353: Tests are not meant to throw exceptions. Might be a leftover, please remove.

src/Smalot/PdfParser/Font.php
```diff
@@ -57,6 +58,17 @@ class Font extends PDFObject
...
     private static $uchrCache = [];
 
+    /**
+     * In some pdf-files (@see https://github.com/smalot/pdfparser/pull/500) encoding could be referenced by object id
+     * but object itself not contains `/Type /Encoding` in its dictionary. These objects wouldn't be initialized as
+     * Encoding in \Smalot\PdfParser\PDFObject::factory() during file parsing (they would be just PDFObject).
+     *
+     * Therefore, we create Encoding instance from them during decoding and cache this value in this property.
+     *
+     * @var Encoding
```
k00ni, line 68:
```suggestion
     * In some PDF-files encoding could be referenced by object id but object itself does not contain
     * `/Type /Encoding` in its dictionary. These objects wouldn't be initialized as Encoding in
     * \Smalot\PdfParser\PDFObject::factory() during file parsing (they would be just PDFObject).
     *
     * Therefore, we create an instance of Encoding from them during decoding and cache this value in this property.
     *
     * @var Encoding
     *
     * @see https://github.com/smalot/pdfparser/pull/500
```

src/Smalot/PdfParser/Font.php
likemusic: > each function should be described in its function header

Imho, the function should be described if the description has any sense, any benefit.

I don't see any sense to add comments which just repeat the name of the function, like below:

```
/**
* Get user by user id
* 
* @param int $userId
* @return User
**/

public function getUserById(int $userId): User
```

likemusic: > There is a good statement about usage of functions here. It talks about reusability and abstraction.

It's doesn't mean that you should use function only if you use both reasons. Just one of them is enough to use a function.

likemusic: > you introduce a function which uses another introduced function which just initialize an instance. Combined with the fact that you only using them once, that is over engineered in my opinion.

In this case, it's just abstraction. 
In my code, I adhere to the following rule:
- If created by function object must be configured, then it should be split to function which just creates an object.

It helps to understand just by function name if something else happens during creating an object. Also if you'll need to have an object with another configuration, you don't need to refactor it in the future.

likemusic: > It suggests complexity/structures which are not there.

As I argued earlier, splitting function to other do not grow complexity. It just adds some private methods to a class, which do not affect understanding the algorithm of the public methods in any way.

Otherwise, when you use inline many lines of code instead of one line function call, it very much increases function size and cognitive load to understand what a function really do.

src/Smalot/PdfParser/Font.php
likemusic: I know what you want to say but in the sentence you basically say "X is an instance of X". But you mean $this->get('Encoding') instead, don't you?

Just now I understand you))

> Decode content when $this->get('Encoding') is instance of Encoding.

Definitely, it would be much better than before!)

I used a quite changed version like:
```
Decode content when $encoding (given by $this->get('Encoding')) is instance of Encoding.
```

src/Smalot/PdfParser/Font.php
likemusic: > When I take the code of your functions and combine it ...

You mix in one function two different operations: instantiating and configuring.
Imho, it is always better to have a low-level function that does just only one thing and hi-level functions that use the low-level function.

In this case, configuring is just calling `init()`. Therefore I don't extract it to another function)) But often configuring objects after creation is not so simple.

I think my version is better because it could help for the next factoring iteration (when $encoding would be private class member instantiated always like Encoding in Font constructor or when objects would not require call init() before using them).

likemusic: > I don't want to risk loosing your commitment over two functions :)

Thank you! I hope it will inspire me to fix some another encoding issues :)

And I understand that introducing a new function just for abstract one-line object configuration from creation seems like over-engineering, but I think for the long term my variant is better.

I feel that soon I'll understand how pdf-files encodings must be decoded and how it does most popular readers/viewers))

tests/Integration/FontTest.php
likemusic: )) Yes. Sorry, my bad.

likemusic:
I added comments to some existing before functions. And in some of them add //todo. But I'm not sure that phpDoc-bock is the right place for this.

src/Smalot/PdfParser/Font.php
k00ni: > Decode content when $encoding (given by $this->get('Encoding')) is instance of Encoding.

That is a good description.

src/Smalot/PdfParser/Font.php
```diff
@@ -306,6 +357,15 @@ public static function decodeEntities(string $text): string
         return $text;
     }
 
+    /**
+     * Check if given string is Unicode text (by BOM);
+     * If true - decode to "utf-8" encoded string.
+     * Otherwise - return text as is.
+     *
+     * //todo: rename in next major release to make the name correspond to reality (for ex. decodeIfUnicode())
```
k00ni, line 365:
```suggestion
     * @todo rename in next major release to make the name correspond to reality (for ex. decodeIfUnicode())
```

src/Smalot/PdfParser/Font.php
```diff
@@ -290,6 +336,11 @@ public static function decodeOctal(string $text): string
         return $text;
     }
 
+    /**
+     * Decode string with html entity encoded chars.
+     * @param string $text
```
k00ni, line 341:
```suggestion
     * Decode string with html entity encoded chars.
     *
     * @param string $text
```

src/Smalot/PdfParser/Font.php
```diff
@@ -236,11 +261,26 @@ public function loadTranslateTable(): array
         return $this->table;
     }
 
+    /**
+     * Set custom char translation table where:
+     * - key - integer character code;
+     * - value - "utf-8" encoded value;
+     *
+     * @param array $table
```
k00ni, line 269:
```suggestion
```
Its redundant.

src/Smalot/PdfParser/Font.php
```diff
@@ -376,91 +442,213 @@ public function decodeText(array $commands): string
...
+    }
+
+    /**
+     * First try to decode $text by ToUnicode CMap.
+     * If char translation not found in ToUnicode CMap tries:
+     *  - If DescendantFonts exists tries to decode char by one of that fonts.
+     *      - If have no success to decode by DescendantFonts interpret $text as a string with "Windows-1252" encoding.
+     *  - If DescendantFonts does not exist just return "?" as decoded char.
+     *
+     * //todo: Seems this is invalid algorithm that do not follow pdf-format specification. Must be rewritten.
```
k00ni, line 473:
```suggestion
     * @todo Seems this is invalid algorithm that do not follow pdf-format specification. Must be rewritten.
```

tests/Integration/FontTest.php
```diff
@@ -340,14 +340,39 @@ public function testDecodeText()
...
-        $this->assertEquals('æöü', $font->decodeText($commands));
+    /**
+     * Font could have indirect encoding without `/Type /Encoding`
+     * which would be instance of PDFObject class (but not Encoding or ElementString).
+     *
+     * @see https://github.com/smalot/pdfparser/pull/500
+     *
+     * @return void
+     */
+    public function testDecodeTextForFontWithIndirectEncodingWithoutTypeEncoding()
```
k00ni, line 353:
```suggestion
     */
    public function testDecodeTextForFontWithIndirectEncodingWithoutTypeEncoding(): void
```

k00ni (commented):
This PR becoming better and better. You didn't had to add all these new function headers. 

They are a welcomed addition, but we should really focus on the encoding fixes in the PR.

About `todo` statements: We currently collect them mostly in function header using `@todo` or inline via `// TODO`.

src/Smalot/PdfParser/Font.php
```diff
@@ -376,91 +442,213 @@ public function decodeText(array $commands): string
...
-            // don't double-encode strings already in UTF-8
-            $text = mb_convert_encoding($text, 'UTF-8', 'Windows-1252');
         }
 
         return $text;
     }
+
+    /**
+     * Decode content by any type of Encoding (dictionary's item) instance.
+     */
```
k00ni, line 527:
```suggestion
     *
     * @throws LogicException if unknown Encoding instance type is used.
     */
```

src/Smalot/PdfParser/Font.php
k00ni: > Imho, the function should be described if the description has any sense, any benefit.
> I don't see any sense to add comments which just repeat the name of the function, like below:

You are right, I wouldn't bother to add function headers to getter-functions either. What I meant and referring to were functions which you introduced and do more than just returning a value. When writing code which is used by others, I ask myself "What do others need to know about my functions to properly use them?".
---

I think we got a good idea what the other person thinks about the topic. Also, I don't wanna start a philosophical discussion here. I follow your arguments here and therefore, this conversation is resolved. Nevertheless, don't hesitate to add further comments.

src/Smalot/PdfParser/Font.php
k00ni: 👍🏿 Resolved for me, see my comments at https://github.com/smalot/pdfparser/pull/500#discussion_r780224706

tests/Integration/FontTest.php
likemusic: By the way, I added a return type for all test's methods.

k00ni: Thats good

k00ni:
Code is in a good shape right now. We can focus on functionality. I would like hear other opinions about this PR.

Maybe @Connum @j0k3r @izabala @Reqrefusion

j0k3r:
First of all, can you remove all the `: void` noise from the PR?
It's better to suggest such changes in a different PR.

Connum:
A happy and healthy new year to all of you! And lots of thanks for the great work @likemusic - this really has the potential to fix some of the other existing encoding bugs, and I sure hope it will!

I'm neither an expert in character encodings nor in the PDF spec, so I don't understand all the details of the changes that were made. But there's nothing that caught my attention, so seeing that it doesn't break any existing tests and fixes the encoding for the exemplary PDF, it looks fine to me!

If I had more time at my hands right now, I would check some of the other issues to see if it either fixes those too or at least doesn't make matters any worse, but I don't think I'll be able to find that time, unfortunately.

Connum:
If anyone finds the time, the test file samples/InternationalChars.pdf I created a while ago would be a starting point for additional testing. There are tests for 6 of the languages in https://github.com/smalot/pdfparser/blob/1b3b6eb5766da659bb514f090b00a5eac72cf4db/tests/Integration/ParserTest.php#L92 but Hindi, Simplified Chinese, Japanese, Korean, Arabic and Hebrew didn't work back then so I left them out of the test. Though I'm not sure this fix will work for those language as well, as they use UTF-16 and as far as I can tell, this fix specifically handles UTF-8 strings?

Reqrefusion:
Actually writing I have been following this PR from the beginning. I didn't want to write from the beginning. However, I am writing as a courtesy since I was invited, thanks for @k00ni. The reason I wanted to write was that I could not bring a new thought to the subject. I disagree with @j0k3r about ":void noise". It may not be correct to add this as an addition here, but it doesn't make sense to do other PR. If it is already done here, there is no need to do it again. After all, it's not such a big change that it needs to be separated. It's fine, except that it makes it harder to skim over this change later on. If this is seen as a problem, necessary actions can be taken.

Like my dear friend @Connum, I do not have much idea about the corrections made here. Looking at the PRs in the past, I can't find any change in this direction. That's why I find this work of @likemusic very valuable. I hope he continues these efforts. I agree with @Connum on tests. In this regard, some tests that have been added before can be passed. It can also be added in a few tests. There are many tests I want to add, but unfortunately I can't find the time.

Finally, I can say that @likemusic increases the readability considerably in some encodings. I see no obstacle to the acceptance of this PR in its current form. I hope @likemusic makes more corrections like this with valuable expertise.

k00ni:
_A bit off topic:_ I hope I didn't make the impression, that anyone needs my permission to contribute. We welcome everyone to contribute, comment and ask on issues and pull requests, invitations not needed here :)

k00ni:
@likemusic how are you doing? Can you tell us your plans for this PR?

@j0k3r @Connum and @Reqrefusion thank you for your feedback.

I can live with the `: void` noise too. I made another commit suggestion, performance tests keep failing and there was a question from @Connum. Can you get back to us in these regards @likemusic?

likemusic:
@k00ni I'll get back to this PR on the Weekend or during next week.

likemusic:
I've updated the test pdf file. Now it opens without errors both in Sumatra PDF Viewer and Adobe Acrobat Reader.

Sumatra Pdf Viewer:
SumatraPdfViewer

Adobe Acrobat Reader
AdobeAcrobatReader

likemusic:
Here is extracted text differences for the test file before and after this PR:
image

likemusic:
> but Hindi, Simplified Chinese, Japanese, Korean, Arabic and Hebrew didn't work back then so I left them out of the test. Though I'm not sure this fix will work for those language as well, as they use UTF-16 and as far as I can tell, this fix specifically handles UTF-8 strings?

@Connum This PR doesn't relate to these issues. As far as I understand issues in InternationalChars.pdf are related to Identity-h/Identity-v encodings handling.

Thank you for this useful notice! It will help to test Identity-h/Identity-v fixes.

likemusic:
To not mix in this PR changes with tests return types, todos, and methods descriptions I've made separate PR  https://github.com/smalot/pdfparser/pull/509 with cherry-picked https://github.com/smalot/pdfparser/pull/500/commits/1ee85771a4a0abe09024f7f86d971f018619fbe8

If it would be merged to master-branch firstly, we could squash commits in this PR to not have not related void-noise in return types.

likemusic:
> performance tests keep failing

@k00ni I don't really understand why this could happen. I've run tests on my laptop under PHP v7.4 without errors.

If this happens again I will try to reproduce it, profile by xDebug, and fix it.

j0k3r:
There is one remaining failing test, could you fix it locally?

```
php tests/Performance/runPerformanceTests.php
```

k00ni:
@likemusic you made the right call to do these off-topic changes in a separate PR. I approved and merged it right away. After merging `master` in this PR should smaller.

j0k3r:
I updated the code to handle `$text` when the encoding isn't `instanceof` something. It was the previous behavior so I think we must keep it: https://github.com/smalot/pdfparser/blob/43ca68fe8e6a179b43b803182c655ba74450239c/src/Smalot/PdfParser/Font.php#L491-L496

This then fix the `tests/Performance/runPerformanceTests.php` script.

Reqrefusion:
I think there has been a problem with old coding related to performance testing, I hope something will be done about this in the future.

k00ni (approved):
Good addition @j0k3r

I merged in `master` branch which leads to a way smaller PR.

PR is accepted. Is there anything left @likemusic?

j0k3r (approved):

likemusic:
@k00ni I don't like @j0k3r fix because it seems like suppressing Logic Error. But in any way, you could merge this PR, and I will make a new one after researching why Header instance is passed during the performance test.

k00ni:
@likemusic sounds good. 

Thank you for being patient and your work here.

Issue #193, opened 2018-03-21
layout Maintain (as best as possible) the original physical layout


TnCoders:
Hi, Can you tell me how can i Maintain (as best as possible) the original physical layout of the text. So keep the Tabulation etc.

antwal:
Hi, i have a problem with a pdf, not return lines of valid table with tabulation but return lines with whitespaces

rubenvanerk:
For me, changing this value helped separating columns: https://github.com/smalot/pdfparser/blob/7f2d319eab7c5b198611cf1a3de13e0ac1dd8288/src/Smalot/PdfParser/PDFObject.php#L287

k00ni:
Hi @rubenvanerk, what exactly has be changed here?

The code `$text .= ' ';` already exists in the current master branch:
https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/PDFObject.php#L287

rubenvanerk:
Ideally it would be nice to programmatically modify this value.

When parsing [this file](https://github.com/smalot/pdfparser/files/4710183/2018-world-championschips-national-teams-open-page1.pdf), the fourth line would result in:
`7 1 NZL/12 1988 Kent, Steven New Zealand  0:45,13 0.0 Q`

It is difficult to separate the name `Kent, Steven` from the country `New Zealand`. Now when I change the horizontal offset to for example `"\t"`:
`7---1---NZL/12---1988---Kent, Steven---New Zealand---0:45,13---0.0---Q`
_Note: I replaced `"\t"` with --- in this example for demonstration purposes._

It suddenly becomes a lot easier to separate the columns.

Overall I think there should be some options to modify the behavior of the parser. I've also seen some issues (https://github.com/smalot/pdfparser/issues/132, https://github.com/smalot/pdfparser/issues/82) recommending manually changing this value:
https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/Font.php#L341

k00ni:
I can't take care of it right now, sorry. 

But i find the idea good to override the value of `getFontSpaceLimit` as mentioned in #259. If someone wants to suggest something via a pull request i would be happy to help finding a solution.

Issue #494, opened 2021-11-30
Random spaces in text


Reqrefusion:
I got this error while trying the library on different pdfs. There are various gaps in the entire text without being connected to a system.

Sample:
```
11 ADET D ERGİ İ ÇİN 3 K ALE M H İZ M ET AL IM I İHALE EDİLECEKTİR
İstanbul Ün iversi tesi-Cerra hpaşa Rek tö rlüğünde n:
İ.Ü .-Cer rah paşa Üniv ers ite sine 11 Adet D ergi İç in 3 K ale m Hi zm et Al ım ı, 4734 sa y ılı
K am u İha le K anununun 3. m addesi (f) bendi ile geçi ci 4. m addesi hük üm le ri ne da yan ıl arak
ha zırl anan 1/1 2/2003 ta ri hli 2003/ 6554 sa y ılı K ara rna me ek i e sa sla rı nın 2 0 /a m adde sine gö re
P aza rlık U su lüyle ih ale ed ile ce k ti r.
İ ha le y e iliş kin a y rı ntılı b ilg iler aşa ğ ıda yer a lm ak tad ır :
1 - İda ren in
a) A dre si : İs tanb ul Ü niv ers ite si-Ce rrah paşa Re ktö rl üğü B ilim se l
Ara ştır m a Pr oje le ri K oord in asy on Birim i/İ da ri Bi rim ler
Binası/A vcıla r Y erl eşk esi/A vcıla r/İS T ANBU L
b) T ele fon ve F ak s Nu mara sı : Tel: 0 212 404 07 95 / Fa ks: 0 212 404 0701
c) E le k tr onik Pos ta A dre si : bap sa tin alm a@ iu c.e du. tr
ç) İ ha le do kü m an ın ın gö rü leb il ece ği i n te rn et ad resi (v ars a): -
2 - İha le kon usu malın/ hiz m etin
1 - İle ri düz ey y ay ınc ılık v e se k re te ry a hiz m eti (1 1 de rgi i ç in) 12 Ay
2- Dil k ont ro lle ri ( 11 d erg i i ç in) 1.000.000 Keli m e
3 - Web t ab an lı h iz m et a lı m ı (Scho la rO ne, P ub lo ns, Web sa y fa sı) 11 Derg i
a) İ ha le K ay ıt N o : 202 1/779 742
b) T esli m y erl eri : İs tanbul Üniv ers ite si- Ce rrah pa şa Re ktö rl üğü /
[...]

PR #505, opened 2022-01-08
Make horizontal offset configurable

rubenvanerk:
Solves the issue as described in this comment: https://github.com/smalot/pdfparser/issues/193#issuecomment-636697897


src/Smalot/PdfParser/Config.php
```diff
@@ -40,6 +40,8 @@ class Config
 {
     private $fontSpaceLimit = -50;
 
+    private $horizontalOffset = ' ';
+
```
k00ni, line 44: Which type is it? string...? Please add `@var` PHPDoc comment.

src/Smalot/PdfParser/Config.php
```diff
@@ -78,6 +80,16 @@ public function setFontSpaceLimit($value)
         $this->fontSpaceLimit = $value;
     }
 
+    public function getHorizontalOffset()
+    {
+        return $this->horizontalOffset;
+    }
+
+    public function setHorizontalOffset($value)
```
k00ni, line 88: Based on the type of the property, use appropriate parameter / return value.

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -280,8 +280,7 @@ public function getText(?Page $page = null): string
                                 $current_position_td['x']
                             )
                         ) {
-                            // horizontal offset
-                            $text .= ' ';
+                            $text .= $this->config->getHorizontalOffset();
```
k00ni, line 283: Can you construct a basic test which includes this change?

k00ni (changes requested):
Thank you!

k00ni:
Does this fix #193?

k00ni (approved):
Thank you for your changes, it looks good to me.

If you have the time can you tell me if this PR fixes #193 and #72?

Any remarks @j0k3r?

rubenvanerk:
As I don't have access to the pdf's in those issues, I can't say for sure. But I think it does.

Since I used the pdf of #494, I can say that it fixes _that_ issue.

j0k3r (approved):

Issue #504, opened 2022-01-05
https://pdfparser.org/ is  is unavailable


likemusic:
When trying to open site I've got Chrome error page with code ERR_CONNECTION_REFUSED.

k00ni:
Ping @smalot

niclm:
Can confirm the same
```
This site can’t be reachedpdfparser.org refused to connect.

Try:
- Checking the connection
- Checking the proxy and the firewall

ERR_CONNECTION_REFUSED
```

zhanghuanchong:
@smalot Same here. The website is down.

suyogricha:
http://web.archive.org/web/20211001174117/https://www.pdfparser.org/documentation

rubenvanerk:
Here's a demo I hacked together for the time being: https://pdfparser.wrve.nl/

Source: https://github.com/rubenvanerk/pdfparser-demo

k00ni:
Out of interest: How important is the website for you guys? Is it so important that we need a replacement?

rubenvanerk:
I'd say only a demo is valuable to quickly determine if the library is suitable for someone's needs. The rest of the website content is just an outdated readme at this point.

Reqrefusion:
The site was fine for demo only. Other than that, there was nothing else.

PR #506, opened 2022-01-17
Link docs to wiki instead of pdfparser.org

rubenvanerk:
Since pdfparser.org is down (#504), I think it's better to link to the wiki instead of the (outdated) site.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ There is no active development by the author of this library (at the moment), bu
 
 ## Documentation ##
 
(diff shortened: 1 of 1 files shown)


Reqrefusion:
As the person who transferred the information on the site completely to the wiki, I find this idea logical until the documentation is prepared. Also, unlike the site, everything on the wiki is up to date. I hope the site will be online soon.

k00ni:
Thank you for this change @rubenvanerk.

> I hope the site will be online soon.

I don't think so. My assumption is that @smalot is busy and this project + site doesn't have the highest priority currently, which is fine. It supports the goal to merge all documentation available and make it easier to maintain and extend.

k00ni (approved):

PR #509, opened 2022-01-28
Add return types to tests methods. Fix todos in phpDocs. Add method's descriptions for Font class.

likemusic:
To not mix these changes with https://github.com/smalot/pdfparser/pull/500


k00ni (approved):
Good job!

PR #510, opened 2022-01-31
Avoid deprecated message in PHP >= 7.4

j0k3r:
If the `$hex` isn't a valid character, a deprecated message is thrown since PHP 7.4:

```
Invalid characters passed for attempted conversion, these have been ignored
```

Checking the string with `ctype_xdigit` first, avoid the message (found here: https://github.com/dompdf/dompdf/issues/2003)

Might fix https://github.com/wallabag/wallabag/issues/4221


j0k3r:
I'm not 100% the condition with `ctype_xdigit` is a proper fix and won't lead to some badly rendered PDF text. Because it the `text` isn't a valid character, it might means that the full sentence will be broken because that `text` won't be included (ie: skipped).

The PDF I used during test isn't properly parsed anyway (the text isn't extracted at all). I found only few online extractor which were able to get the text of the PDF. That means I can't be sure of the behavior in case of `text` skipped and content looking good..

k00ni:
What do you think about the idea to make this an configurable option (using `Config`)?

j0k3r:
[The PDF is public](https://www.geeek.org/content/files/2022/01/ebook_sept2003.pdf), I don't think it's a problem. And I also don't have an other PDF which generates the error.

What part should I move behind the config? The workaround for the error or the encoding fix?

k00ni:
You wrote:

> I'm not 100% the condition with ctype_xdigit is a proper fix and won't lead to some badly rendered PDF text. Because it the text isn't a valid character, it might means that the full sentence will be broken because that text won't be included (ie: skipped).

and 

> The PDF I used during test isn't properly parsed anyway (the text isn't extracted at all). I found only few online extractor which were able to get the text of the PDF. That means I can't be sure of the behavior in case of text skipped and content looking good..

To me it sounds like edge cases which might lead to unintended behavior if not handled correctly. So first of all, we should split this up: one PR to fix the PHP 7.4 error message and another one for the "non badly" encoding issue.

My question regarding "ctype_xdigit" is: does it make sense to make that fix enable able via `Config` class?

j0k3r:
I agree that we should split the PR BUT I don't have an other PDF with badly encoded details. So I can create an other PR, but I can't provide tests to cover that fix.

About the `ctype_xdigit` behind a config, I don't know.
On one side, it'll avoid unintended behavior because not everyone will enable it.
On the other side, if user got the error and then enable the config, it'll be enable for all other PDF parsed by the user.

The best solution should be to catch that PHP 7.4 deprecated message, but I don't think we can.

k00ni:
Can you reduce the PR so we only have changes regarding PHP 7.4 deprecation message?

j0k3r:
@k00ni PR reduced

j0k3r:
👋 @k00ni

k00ni:
Sorry @j0k3r I am super busy currently and don't know when to help out with this issue. Based on your feedback we have either make a hard decision or do more research.

k00ni:
Is there a way to replace the given PDF with a basic test string that raises the deprecation message? Looking at the PDF I am worried that it contains copyrighted material.

k00ni:
> The best solution should be to catch that PHP 7.4 deprecated message, but I don't think we can.

If we have issues which reference the warning you mentioned, we can continue here. Until now I would close to reduce number of stale PRs. Hope its OK for you.

Issue #498, opened 2021-12-20
Updating documentation


likemusic:
This library has some helpfull config options and methods that not covered in documentation ($cfg's 'ignore_filter_decoding_errors' and 'ignore_missing_filter_decoders' options, Page's `getDataTm()` and `getTextXY()` methods e.t.c.).

k00ni:
That is on my TODO list for a while and I plan to propose some improvements in January 2022.

Reqrefusion:
Hello, I also made some suggestions about documentation before. I thought it would be good to solve this problem with the wiki page. This is quite appropriate, especially considering the way the project is progressing. I invite you to contribute to the wiki.

k00ni:
Thank you for your contributions here @Reqrefusion.

I personally don't like usage of our Wiki, because it can be changed by everyone (with no way to approve/decline changes). My suggestion would be to utilize Markdown-files (like DEVELOPER.md), because we can use pull requests. 

Its sad that @smalot is off for a while again, because documentation at pdfparser.org's is also outdated and we should discuss a way how to organize documentation in general.

k00ni:
Note: include #505 when updating documentation.

Reqrefusion:
- [ ] Preparing a more detailed briefing for the config.
I found this necessary because @rubenvanerk  provides solutions to some problems by using the config feature. He had already made an addition in this direction on the [wiki page](https://github.com/smalot/pdfparser/wiki/Use). I also added a table. But still work in process.

rubenvanerk:
Maybe it's also an idea to have a 'Common problems' section with possible solutions.

k00ni:
Thanks again for the work on the wiki folks.

After #500 is done (hopefully end of January) we should focus on documentation. Until then, is there anything "missing" in documentation and/or wiki which you would like to include?

Would it be sufficient to just "copy" the wiki and merge it into the README.md with a few style improvements?

rubenvanerk:
A section with limitations would also be useful. Like parsing form fields.

Adding all the wiki documentation to the readme might be a bit much, but I think it could work.

k00ni:
I haven't forgotten but I am very busy currently. If anyone want to propose something via pull request, feel free.

My approach would be to split (wiki) documentation into different files in `doc` folder. I would also remove some stuff from README.md and maybe update other info.

PR #513, opened 2022-02-12
Rework documentation

rubenvanerk:
Here's my shot at a new README. We might be able to get away with putting the documentation in here, but I'm not too sure.

Related: #498

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ There is no active development by the author of this library (at the moment), bu
 - Handling of hexa and octal encoding in text sections
 - Create custom configurations (see [/doc/CustomConfig.md]).
 
-Currently, secured documents are not supported.
+Currently, secured documents and extracting form data are not supported.
 
 ## License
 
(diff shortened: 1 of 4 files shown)


k00ni:
Thank you!

I took the liberty and made some changes in your branch. Hope its OK. English is not my mother tongue so you might find instances to improve.

k00ni:
_I refined some things and added latest wiki-content made by @Reqrefusion._

What is left to do here?

@Reqrefusion After this got accepted and merged I would like to remove all content from wiki and only leave a hint to our new documentation. That way there is no confusion where to find latest documentation. What do you think? 

If there is anything I missed, please give me a hint.

k00ni:
OK, I will leave this open until end of this week. If there are no objections or comments, this PR gets merged afterwards. Documentation is way from finished but with this work there is a good basement in my opinion.

k00ni (approved):

Reqrefusion:
I feel bad seeing this so late. It seems like the end of Wiki has come. One of Github's nice but insufficient features. I will still make a few changes to the config. The current config is not very descriptive. The table I created beforehand makes trial and error easier.

PR #516, opened 2022-02-24
Added font info to dataTm

shtayerc:
Font id and size are useful when searching for text in pdf.
With that info you can calculate text height and width (now we only have x and y).

### DIFF
--- a/doc/Usage.md
+++ b/doc/Usage.md
@@ -62,7 +62,7 @@ Array
 )
 ```
 
-When activated via Config setting (Config::setDataTmFontInfoHasToBeIncluded(true)) font identifier (index 2) and font size (index 3) are added to dataTm.
+When activated via Config setting (`Config::setDataTmFontInfoHasToBeIncluded(true)`) font identifier (index 2) and font size (index 3) are added to dataTm.
 
 ```php
 // create config
--- a/doc/Usage.md
+++ b/doc/Usage.md
@@ -21,6 +21,109 @@ $text = $pdf->getText();
 $text = $pdf->getPages()[0]->getText();
 ```
 
+## Extract text positions
+
+You can extract transformation matrix (indexes 0-3) and x,y position of text objects (indexes 4,5).
+
+```php
+$data = $pdf->getPages()[0]->getDataTm();
+
+Array
+(
+    [0] => Array
+        (
+            [0] => Array
+                (
+                    [0] => 0.999429
+                    [1] => 0
+                    [2] => 0
+                    [3] => 1
+                    [4] => 201.96
+                    [5] => 720.68
+                )
+
+            [1] => Document title
+        )
+
+    [1] => Array
+        (
+            [0] => Array
+                (
+                    [0] => 0.999402
+                    [1] => 0
+                    [2] => 0
+                    [3] => 1
+                    [4] => 70.8
+                    [5] => 673.64
+                )
+
+            [1] => Calibri : Lorem ipsum dolor sit amet, consectetur a
+        )
+)
+```
+
+When activated via Config setting (Config::setDataTmFontInfoHasToBeIncluded(true)) font identifier (index 2) and font size (index 3) are added to dataTm.
+
+```php
+// create config
+$config = new Smalot\PdfParser\Config();
+$config->setDataTmFontInfoHasToBeIncluded(true);
+
+// use config and parse file
+$parser = new Smalot\PdfParser\Parser([], $config);
+$pdf = $parser->parseFile('document.pdf');
+
+$data = $pdf->getPages()[0]->getDataTm();
+
+Array
+(
+    [0] => Array
+        (
+            [0] => Array
+                (
+                    [0] => 0.999429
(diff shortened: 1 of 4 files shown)


k00ni:
@shtayerc thank you for your contribution.

I like your extension. Can you check https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/Config.php and see if your values can be moved there? It would make this even more flexible. Here is an example how to use Config object: https://github.com/smalot/pdfparser/wiki/Use#config

src/Smalot/PdfParser/Page.php
```diff
@@ -658,6 +665,12 @@ public function getDataTm(array $dataCommands = null): array
          */
         $defaultTl = 0;
 
+        /*
+         *  Set default values for font data
+         */
+        $defaultFontId = -1;
+        $defaultFontSize = 0;
```
k00ni, line 672: Do these values matter later on? If yes, makes it sense to move them to `Config` instance too?

shtayerc: It does not matter later on. I just used some values that could not be valid font id or font size.

shtayerc: @k00ni btw, I wonder why the are separate cases for Tj and TJ here https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/Page.php#L767,L814
Could this be joined? Now there are two cases that do the same thing

k00ni (changes requested):
Thank you for your changes.

The current name **dataTmIncludeFontInfo** could be improved in my opinion, so it outlines its meaning more clearly. 

I suggest **dataTmFontInfoHasToBeIncluded**, which works good with if-clauses.

What do you think?

shtayerc:
Sure. Name is changed

k00ni:
I merged #513, can you add a basic example in https://github.com/smalot/pdfparser/blob/master/doc/Usage.md how to use this new function, please?

doc/Usage.md
```diff
@@ -21,6 +21,130 @@ $text = $pdf->getText();
...
+                (
+                    [0] => 0.999402
+                    [1] => 0
+                    [2] => 0
+                    [3] => 1
+                    [4] => 70.8
+                    [5] => 698.96
+                )
+
+            [1] =>
```
k00ni, line 60:
```suggestion
    [1] => Array
        (
            [0] => Array
                (
                    [0] => 0.999402
                    [1] => 0
                    [2] => 0
                    [3] => 1
                    [4] => 70.8
                    [5] => 698.96
                )

            [1] =>
```

Does it make sense to remove this part, because it is not clear what it shows. Elements of index 0 and 2 are more clear. What do you think?

shtayerc: It is only space which is not shown by print_r. I wil remove it.

doc/Usage.md
```diff
@@ -21,6 +21,98 @@ $text = $pdf->getText();
...
+                    [4] => 70.8
+                    [5] => 673.64
+                )
+
+            [1] => Calibri : Lorem ipsum dolor sit amet, consectetur a
+        )
+)
+```
+
+Additionally font identifier - index for `Document::getFonts()` array (index 2) and font size (index 3) can be added to dataTm with config setting.
```
k00ni, line 65: I would adjust this sentence a bit:

> When activated via Config setting (`Config::setDataTmFontInfoHasToBeIncluded(true)`) font identifier (index 2) and font size (index 3) can be added to dataTm.

What do you think? Please refine it if to unclear/faulty.

shtayerc: Your suggestion is more clear, i changed only the end 


> can be added to dataTm

to

 > are added to dataTm

src/Smalot/PdfParser/Page.php
```diff
@@ -799,6 +823,10 @@ public function getDataTm(array $dataCommands = null): array
                     $extractedData[] = [$Tm, $data[2]]; //Verify
                     break;
 
+                case 'Tf':
+                    list($fontId, $fontSize) = explode(' ', $command['c'], 2);
```
k00ni, line 827: Is a further check here required? Or can we assume that `$command['c']` always has first two elements set?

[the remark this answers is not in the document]
shtayerc: From [PDF 1.0 spec](https://ia902503.us.archive.org/10/items/pdfy-0vt8s-egqFwDl7L2/PDF%20Reference%201.0.pdf) page 106

> fontname size **Tf**  Set font and size
Sets the text font and text size in the graphics state. There is no default value for
either fontname or size; they must be selected using Tf before drawing any text.
fontname is a resource name. size is a number expressed in text space units.

Definition is the same in 1.7 spec. I think it is safe to assume that both values are always set.

k00ni (approved):
Looks good to me. Thank you for your patience and time.

I will leave this open until end of this week. If there are no objections / comments it will be merged next week.

j0k3r (approved):

PR #517, opened 2022-02-25
Added calculateTextWidth function to Font

shtayerc:
Character width is read from font details field 'Widths'.

### DIFF
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -268,6 +268,38 @@ class Font extends PDFObject
     }
 
     /**
+     * Calculate text width with data from header 'Widths'. If width of character is not found then character is added to missing array.
+     */
+    public function calculateTextWidth($text, &$missing = null): ?float
+    {
+        $index_map = array_flip($this->table);
+        $details = $this->getDetails();
+        $widths = $details['Widths'];
+
+        // Widths array is zero indexed but table is not. We must map them based on FirstChar and LastChar
+        $width_map = array_flip(range($details['FirstChar'], $details['LastChar']));
+
+        $width = null;
+        $missing = [];
+        $textLength = mb_strlen($text);
+        for ($i = 0; $i < $textLength; ++$i) {
+            $char = mb_substr($text, $i, 1);
+            if (
+                !\array_key_exists($char, $index_map)
+                || !\array_key_exists($index_map[$char], $width_map)
+                || !\array_key_exists($width_map[$index_map[$char]], $widths)
+            ) {
+                $missing[] = $char;
+                continue;
+            }
+            $width_index = $width_map[$index_map[$char]];
+            $width += $widths[$width_index];
+        }
+
+        return $width;
+    }
+
+    /**
      * Decode hexadecimal encoded string. If $add_braces is true result value would be wrapped by parentheses.
      */
     public static function decodeHexadecimal(string $hexa, bool $add_braces = false): string
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -270,7 +270,7 @@ class Font extends PDFObject
     /**
      * Calculate text width with data from header 'Widths'. If width of character is not found then character is added to missing array.
      */
-    public function calculateTextWidth($text, &$missing = null): ?float
+    public function calculateTextWidth(string $text, array &$missing = null): ?float
     {
         $index_map = array_flip($this->table);
         $details = $this->getDetails();
--- a/doc/Usage.md
+++ b/doc/Usage.md
@@ -57,6 +57,10 @@ Try to calculate text width for given font.
 Characters without width are added to `$missing` array in second parameter.
 
 ```php
-$font = reset($pdf->getFonts()); //get first font
-$width = $font->calculateTextWidth("Some text", $missing);
+$parser = new \Smalot\PdfParser\Parser();
+$pdf = $parser->parseFile('document.pdf');
+// get first font (we assume here there is at least one)
+$font = reset($pdf->getFonts());
+// get width
+$width = $font->calculateTextWidth('Some text', $missing);
 ```
--- a/doc/Usage.md
+++ b/doc/Usage.md
@@ -50,3 +50,13 @@ $pdf = $parser->parseContent(base64_decode($base64PDF));
 $text = $pdf->getText();
 echo $text;
 ```
+
+## Calculate text width
+
+Try to calculate text width for given font.
+Characters without width are added to `$missing` array in second parameter.
+
+```php
+$font = reset($pdf->getFonts()); //get first font
+$width = $font->calculateTextWidth("Some text", $missing);
+```
--- a/tests/Integration/FontTest.php
+++ b/tests/Integration/FontTest.php
@@ -410,4 +410,25 @@ TEXT;
         $font->setTable([]);
         $this->assertEquals('?', $font->translateChar('a'));
     }
+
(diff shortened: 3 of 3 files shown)


src/Smalot/PdfParser/Font.php
```diff
@@ -267,6 +267,33 @@ public function setTable(array $table)
...
+    /**
+     * Calculate text width with data from header 'Widths'. If width of character is not found then character is added to missing array.
+     */
+    public function calculateTextWidth($text, &$missing = null): float
+    {
+        $index_map = array_flip($this->table);
+        $details = $this->getDetails();
+        $widths = $details['Widths'];
+
+        //Widths array is zero indexed but table is not. We must map them based on FirstChar and LastChar
```
k00ni, line 279:
```suggestion
        // Widths array is zero indexed but table is not. We must map them based on FirstChar and LastChar
```

src/Smalot/PdfParser/Font.php
```diff
@@ -267,6 +267,33 @@ public function setTable(array $table)
...
+        $width_map = array_flip(range($details['FirstChar'], $details['LastChar']));
+
+        $width = 0;
+        $missing = [];
+        for ($i = 0; $i < mb_strlen($text); ++$i) {
```
k00ni, line 284:
```suggestion
        $textLength = mb_strlen($text);
        for ($i = 0; $i < $textLength; ++$i) {
```

shtayerc:
@k00ni 
`$missing` parameter is solving problem about user not knowing if text width is correct or not. Now we give user a list of missing characters so he can handle errors by himself. I also wanted to make that parameter optional. I saw similar usage in [preg_match](https://www.php.net/manual/en/function.preg-match.php) function. I am not sure what undefined behavior you mean.

If you prefer to not use parameter by reference, it could be done without reference by returning array, for example:
```php
[
    'width' => 123,
    'missing' => ['a', 'b']
]
```

k00ni:
> I am not sure what undefined behavior you mean.

Can we assume that all calls using an array-index inside the function are succeed (meaning all keys we wanna access are set)? Related variables are `$width_map[$index_map[$char]]`, `$index_map[$char]` and `$widths[$width_index]`.

src/Smalot/PdfParser/Font.php
```diff
@@ -267,6 +267,35 @@ public function setTable(array $table)
...
+        // Widths array is zero indexed but table is not. We must map them based on FirstChar and LastChar
+        $width_map = array_flip(range($details['FirstChar'], $details['LastChar']));
+
+        $width = 0;
+        $missing = [];
+        $textLength = mb_strlen($text);
+        for ($i = 0; $i < $textLength; ++$i) {
+            $char = mb_substr($text, $i, 1);
+            if (!\array_key_exists($char, $index_map) || !\array_key_exists($index_map[$char], $width_map) ||
+                !\array_key_exists($width_map[$index_map[$char]], $widths)) {
```
k00ni, line 288:
```suggestion
            if (
                !\array_key_exists($char, $index_map) 
                || !\array_key_exists($index_map[$char], $width_map)
                || !\array_key_exists($width_map[$index_map[$char]], $widths)
            ) {
```

k00ni:
> I am not sure, both arrays are read from pdf file. I added check for all indexes.

Thanks, this way we can be sure that an index is always set before accessing it.

Thinking about the return value, we should keep float as return value. But is it semantically correct to assume that the width is a float equal or higher than 0? I am not sure here, maybe PDF specification is more clear. What do you think about setting `$width = null;` instead of `$width = 0;` (and also adapt return value of function to `?float`)?

shtayerc:
`$width = null;` is better (character width can be 0. Now user can see if loop was skipped or width is actually 0), changed. According to pdf spec widths are integers, but the array returned by `Font::getDetails()` contains floats. I guess it makes sense to continue returning float or null.

k00ni:
> According to pdf spec widths are integers [...]

We should be in line with the PDF specification here, even though other functions do this differently. Can you paste the part of the specification where you found this information, please? Also change return value from float to integer.

shtayerc:
Both specs were found [here](https://www.pdfa.org/resource/pdf-specification-index/)
In [PDF 1.0 spec](https://ia902503.us.archive.org/10/items/pdfy-0vt8s-egqFwDl7L2/PDF%20Reference%201.0.pdf) chapter 4.3 
> PDF provides two types of numbers, integer and real. Integers may be specified by
signed or unsigned constants. Reals may only be in decimal format. Throughout
this book, number means an object whose type is either integer or real.

In examples of Widths array in the same pdf there are only intergers. Because reals should be written only in decimal format I assumed that Widths are integers.

But in [PDF 1.7 spec](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdf_reference_archive/pdf_reference_1-7.pdf) chapter 3.2.2

> Throughout this book, the term number refers to an object whose type may be
either integer or real. Wherever a real number is expected, an integer may be used
instead and is automatically converted to an equivalent real value. For example, it
is not necessary to write the number 1.0 in real format; the integer 1 is sufficient.

But here it says that reals can be written without decimal part. I only checked spec 1.0 before.
Now I think it makes sense to keep float type. It handles both specs (it does not hurt to have int represented as float).

k00ni:
Thank you! OK, lets keep float.

k00ni:
I merged #513, can you add a basic example in https://github.com/smalot/pdfparser/blob/master/doc/Usage.md how to use this new function, please?

shtayerc:
I added basic example

src/Smalot/PdfParser/Font.php
```diff
@@ -267,6 +267,38 @@ public function setTable(array $table)
         $this->table = $table;
     }
 
+    /**
+     * Calculate text width with data from header 'Widths'. If width of character is not found then character is added to missing array.
+     */
+    public function calculateTextWidth($text, &$missing = null): ?float
```
k00ni, line 273: Can we change second parameter to `array &$missing = null`? This way its type is explicit and more clear.

k00ni:
_I extended your example a bit._

I wanna make sure I understand the meaning of this function: You call it with a text part (which must be in the PDF?) and a variable (in which any characters are stored, which have no width). It tries to find the width for the whole text part?

src/Smalot/PdfParser/Font.php
shtayerc: Done

shtayerc:
Yes, it is close to what you said. Text part must be written in font from which you call the function - otherwise it is missing. Preferably it should be used with `getDataTm()` text part and font identifier (from my other PR) to make sure that widths for all characters are available.

k00ni:
> Preferably it should be used with getDataTm() text part and font identifier (from my other PR) to make sure that widths for all characters are available.

If you think that is important, please add a hint and/or example to the usage passage. Both of your PRs will be merged so it doesn't matter in which you do this. That will help users and avoid confusion.

k00ni:
> If you think that is important, please add a hint and/or example to the usage passage. Both of your PRs will be merged so it doesn't matter in which you do this.

@shtayerc, can you give me a short status info about remaining tasks here? Thanks!

shtayerc:
@k00ni I think everything is done. Example usage with dataTm was added in previous PR (at the end of [this section](https://github.com/smalot/pdfparser/blob/master/doc/Usage.md#extract-text-positions))

k00ni (approved):
Thank you for getting back to me so fast.

PR looks good. I will leave it open until end of this week. If there are no objections or comments, it will be merged next week.

PR #518, opened 2022-03-01
Fix Parsing Header Element with null value

Happyrain67:
`Parser::parseHeader` needs an array at first argument. When $value is null an exception is thrown.

### DIFF
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -263,6 +263,10 @@ class Parser
         switch ($type) {
             case '<<':
             case '>>':
+                
+                if(empty($value))
+                    return null;
+                
                 $header = $this->parseHeader($value, $document);
                 PDFObject::factory($document, $header, null, $this->config);
 


k00ni:
@Happyrain67 thanks for this PR.

Can you please describe a bit how this change helps?
Also add one or two tests to strengthen the code base.

Happyrain67:
Your function "parseHeader" need an Array at first argument.
When $value is null this throw an exception.

What do you want more ?

k00ni:
We value any contribution here. But for each new one its always helpful to get some background information. We don't know each class here in and out (and no, its not my function, I am just helping out here as a maintainer).

> What do you want more ?
There are few checks failing. Can you fix them?

Happyrain67:
Sorry, I understand for testing but I don't know because fails.
I don't have lot of times for look that, 
I'll come back to that later.

k00ni:
We changed some code parts lately, you might wanna merge it latest master-branch.

k00ni:
Please reopen if you want to work on this again.

PR #524, opened 2022-03-24
Add issue template

rubenvanerk:
This should help eliminate some follow-up questions like PHP version and PDF requests.


k00ni (approved):
Cool! I always wanted something like that.

j0k3r (approved):

PR #525, opened 2022-04-01
Handle null Header and falsy header elements

### LINKED ISSUE #514: exception Call to a member function getContent() on bool
Hi,
i try get text from this PDF:
[C1RI6+x3HOS.pdf](https://github.com/smalot/pdfparser/files/8068403/C1RI6%2Bx3HOS.pdf)

php with laravel installation (latest version of pdfparser), php7.4
```

        try {
            $pdf = $parser->parseFile(storage_path() . '/app/mpdf/' . $tempPDFName);
        } catch (\Exception $e) {
            dd($e);
        }
```

Out error:
```
Call to a member function getContent() on bool

  at vendor/smalot/pdfparser/src/Smalot/PdfParser/Document.php:107
    103▕                 }
    104▕ 
    105▕                 $this->dictionary[$type]['all'][$id] = $object;
    106▕ 
  ➜ 107▕                 $subtype = $object->getHeader()->get('Subtype')->getContent();
    108▕                 if (null != $subtype) {
    109▕                     if (!isset($this->dictionary[$type]['subtype'][$subtype])) {
    110▕                         $this->dictionary[$type]['subtype'][$subtype] = [];
    111▕                     }

      +4 vendor frames
```

dsuurlant:
**Document.php**
`$object->getHeader()` can be `null` but that is never checked before calling `->get(...)` methods on it. Fixed by adding a `continue` if `getHeader()` returns `null`.

**Header.php**
Header elements may resolve to have `false` as a value (in `ElementDate::parse`, `ElementString::parse` and `ElementName::parse`). Another solution would be to change those `bool` returns to i.e. `ElementNull` or `ElementMissing`, but that seemed too big of a change, so instead I added a check to `Header->get(...)` that returns `ElementMissing` if the value of the Element is falsy (i.e. `null, false, 0, ''` )

This should fix #514 .

### DIFF
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -78,7 +78,7 @@ class PDFObject
         ?Config $config = null
     ) {
         $this->document = $document;
-        $this->header = null !== $header ? $header : new Header();
+        $this->header = $header ?? new Header();
         $this->content = $content;
         $this->config = $config;
     }
--- a/tests/Integration/HeaderTest.php
+++ b/tests/Integration/HeaderTest.php
@@ -155,6 +155,18 @@ class HeaderTest extends TestCase
         $this->assertTrue($header->get('Font') instanceof Page);
         $this->assertTrue($header->get('Image') instanceof ElementMissing);
         $this->assertTrue($header->get('Resources') instanceof ElementMissing);
+
+        /**
+         * A double forward slash in the header's content results in a falsy element
+         * that should be parsed to ElementMissing instead.
+         *
+         * @see https://github.com/smalot/pdfparser/pull/525
+         */
+        $content = '<</Type/Page/SubType//Text>>foo';
+        $position = 0;
+        $header = Header::parse($content, $document, $position);
+
+        $this->assertTrue($header->get('SubType') instanceof ElementMissing);
     }
 
     public function testResolveXRef(): void
--- a/src/Smalot/PdfParser/Header.php
+++ b/src/Smalot/PdfParser/Header.php
@@ -132,8 +132,8 @@ class Header
      */
     public function get(string $name)
     {
-        if (\array_key_exists($name, $this->elements)) {
-            return $this->resolveXRef($name);
+        if (\array_key_exists($name, $this->elements) && $element = $this->resolveXRef($name)) {
+            return $element;
         }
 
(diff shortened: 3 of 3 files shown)


j0k3r (commented):
Any chance to add some tests to cover your changes?

dsuurlant:
Updated the PR:

* Added test for Header->get when an element may be `false` 
* Removed the `null` check on `getHeader()` since it is actually never null -- it's instantiated in the constructor of `PDFObject`. I also updated the property and getter to reflect this.

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -92,7 +92,7 @@ public function getDocument(): Document
         return $this->document;
     }
 
-    public function getHeader(): ?Header
```
k00ni, line 95: That is an API change and would require us to raise major version of the library. Can you please revert it?

dsuurlant:
Updated the PR:
* Reverted changes to PDFObject API

tests/Integration/HeaderTest.php
```diff
@@ -155,6 +155,12 @@ public function testGet(): void
         $this->assertTrue($header->get('Font') instanceof Page);
         $this->assertTrue($header->get('Image') instanceof ElementMissing);
         $this->assertTrue($header->get('Resources') instanceof ElementMissing);
+
+        $content = '<</Type/Page/SubType//Text>>foo';
+        $position = 0;
+        $header = Header::parse($content, $document, $position);
+
+        $this->assertTrue($header->get('SubType') instanceof ElementMissing);
```
k00ni, line 163: Thank you for your latest changes.

Can you please add a short comment here, which describes what these lines do. I assume you try to exploit the error before but with your patch it works now?

k00ni (changes requested):

tests/Integration/HeaderTest.php
k00ni: Yeah, it looks great, go for it.

k00ni: You might have to use `/*` instead of `/**`.

k00ni:
🥳 Looks good to me. I will leave this open until beginning of next week and will merge it, if there are no objections.

Thanks for your work @dsuurlant!

k00ni (approved):

Issue #542, opened 2022-05-31
Cannot parse read correctly with getTextArray() method


Arnaldo66:
- PHP Version: 7.4
 - PDFParser Version: 2.2.1

### Description:
The parser cannot read information on this PDF. It's postal stamp fedited by french post office.
It contains img and text.

### PDF input

[https://github.com/Arnaldo66/tools_file/blob/main/1653988714_import_stamp.pdf](https://github.com/Arnaldo66/tools_file/blob/main/1653988714_import_stamp.pdf)

### Expected output & actual output

Expected output on the first page for example:
FRANCE
Timbre imprimé sur laposte.fr
Lettre verte
Max 50g
FRANCE
SD:87000639942948A

Actual output on the first page
b"Timbre imprimÃ© sur laposte.fr\x00y\x01\x04\x00\x04\x00Ç\x00:\x00K\x00²\x00( \x06( \x05È\x05Ç\x05À\x05À\x05À\x05Æ\x05Ã\x05É\x05É\x05Ä\x05Â\x05É\x05Ä\x05È\x00\x04\n\x00¸\x01ÉÌ‡Ì‡\x02h\x01É \x02¤\x01É\x02hÌ‡\x01É \n\n\x00Ta\x01X \x01ä\x01ß\x00æ\n\x009\x00t\x00\x04\x00U\x00\x1D\x00'    "

for information, I tried to read why online parser and the result is good:
[https://pdftotext.com/fr/](https://pdftotext.com/fr/)

### Code

$pdf    = $parser->parseFile($filePath);
$pages  = $pdf->getPages();
dump($pages[0]->getTextArray());

GreyWyvern:
`getTextArray()` is returning properly parsed output in the latest release v2.7.0

```
Timbre imprimé sur laposte.frFRANCESD : 87000639942948A
Lettre verte 

Max 50g
FRANCE    
```

PR #544, opened 2022-06-09
Fixes text decoding error when Tf command appears before BT. (#542)

cris-shi:
### Issue Description
The [#Issue 542](https://github.com/smalot/pdfparser/issues/542) caused by the Tf command appears before the text block. And the method PDFObject::getSectionText only extracts command within BT & ET.

```pdf
Q
0.18 Tc
/F1 3.05 Tf
BT
0 0 0 sc
5.66929 7.65354 Td
<007D00F0010900C5012B00D0000300F001090128012B00F0010900D10003012E0137012B0003010300AD0128010F012E013400D0021800E5012B> Tj
ET
```

### Solution
```php
public function getSectionsText(?string $content): array
{
    ...
    if (preg_match_all('/(\sQ)?\s+(.*?)BT[\s|\(|\[]+(.*?)\s*ET(\sq)?/s', $textCleaned, $matches, \PREG_OFFSET_CAPTURE)) {
        foreach ($matches[3] as $pos => $part) {
            ...

            // Add Tx commands which before BT.
            // @see: https://github.com/smalot/pdfparser/issues/542
            if (!empty($matches[2][$pos][0]) && preg_match('/\sTf\s/', $matches[2][$pos][0])) {
                $section = trim($matches[2][$pos][0].$section);
            }

            ...
        }
    }
    ...
}
```

### DIFF
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -191,8 +191,8 @@ class PDFObject
         $textCleaned = $this->cleanContent($content, '_');
 
         // Extract text blocks.
-        if (preg_match_all('/(\sQ)?\s+BT[\s|\(|\[]+(.*?)\s*ET(\sq)?/s', $textCleaned, $matches, \PREG_OFFSET_CAPTURE)) {
-            foreach ($matches[2] as $pos => $part) {
+        if (preg_match_all('/(\sQ)?\s+(.*?)BT[\s|\(|\[]+(.*?)\s*ET(\sq)?/s', $textCleaned, $matches, \PREG_OFFSET_CAPTURE)) {
+            foreach ($matches[3] as $pos => $part) {
                 $text = $part[0];
                 if ('' === $text) {
                     continue;
@@ -203,9 +203,17 @@ class PDFObject
                 // Removes BDC and EMC markup.
                 $section = preg_replace('/(\/[A-Za-z0-9]+\s*<<.*?)(>>\s*BDC)(.*?)(EMC\s+)/s', '${3}', $section.' ');
 
+                // Add Tx commands which before BT.
+                // @see: https://github.com/smalot/pdfparser/issues/542
+                if (!empty($matches[2][$pos][0]) && preg_match('/\sTf\s/', $matches[2][$pos][0])) {
+                    $section = trim($matches[2][$pos][0].$section);
+                }
+
                 // Add Q and q flags if detected around BT/ET.
                 // @see: https://github.com/smalot/pdfparser/issues/387
-                $section = trim((!empty($matches[1][$pos][0]) ? "Q\n" : '').$section).(!empty($matches[3][$pos][0]) ? "\nq" : '');
+                if (empty($matches[2][$pos][0]) || !preg_match('/\sq\s/', $matches[2][$pos][0])) {
+                    $section = trim((!empty($matches[1][$pos][0]) ? "Q\n" : '').$section).(!empty($matches[4][$pos][0]) ? "\nq" : '');
+                }
 
                 $sections[] = $section;
             }
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -191,8 +191,8 @@ class PDFObject
         $textCleaned = $this->cleanContent($content, '_');
 
         // Extract text blocks.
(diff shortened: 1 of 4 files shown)


k00ni (changes requested):
@cris-shi first of all, thank you for your pull request!

* [ ] Can you add 1 or 2 PHPUnit tests which show that your changes work?

Adding these if-clauses might solve the issue for now, but it also adds complexity. Developer now have to take these also into account when working with the code. 
* [ ] An improvement would be to add further explanation/documentation. Why was it changed? What didn't work before? What is possible from now on?
* [ ] Another point is the position of this fixes. Is this the only way to handle these flags?
* [ ] Last point: Please split up these one-liners to increase readability. You are not only making a lot of assumption (which may fail during runtime), but also stack conditions with function calls.

I am looking forward on your feedback and getting this to work.

cris-shi:
And I did a little more research. From what I understand, these text parsing errors are caused by not selecting the correct text font, which contains CMap table.

Outside the text block, there are three commands that will change the current font:
Q: Restore graphics state
q: Save graphics state
Tf: Set text font and size

So I rewrote the code in the function PDFObject::getSectionsText to extract all the commands of Q, q, Tf before the text block, which might affect the font selection of the text block.

> 
> @cris-shi first of all, thank you for your pull request!
> 
> * [ ]  Can you add 1 or 2 PHPUnit tests which show that your changes work?
> 
> Adding these if-clauses might solve the issue for now, but it also adds complexity. Developers now have to take these also into account when working with the code.

I added a PHPUnit test with pdf sample provided by issue 542.

> * [ ]  An improvement would be to add further explanation/documentation. Why was it changed? What didn't work before? What is possible from now on?

In most cases, the font is defined in the text block, the original code work fine with this.
But in this case, the font is defined before the text block.

> * [ ]  Another point is the position of this fixes. Is this the only way to handle these flags? Architecture of PDFParser isn't the best and it grew organically.

Handling these flags worked fine. The problem is extracting these flags.
As I know, this is the only position to extract text blocks, so I think this fixes is enough.

> * [ ]  Last point: Please split up these one-liners to increase readability. You are not only making a lot of assumptions (which may fail during runtime), but also stack conditions with function calls.
> 
> I am looking forward to your feedback and getting this to work.

k00ni (changes requested):
Hey! Thank you for taking the time.

You wrote: 
> And I did a little more research. From what I understand, these text parsing errors are caused by not selecting the correct text font, which contains CMap table.

Can you add a link (or part reference) to your source?

* [ ] I appreciate the new test, but please revert your changes in `testGetSectionText`. Please copy it and make your additions. This way we preserve the meaning of the old test and test PDFParser with similar parameters. Same goes for `testGetSectionText` of PageTest.

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -191,7 +191,7 @@ public function getSectionsText(?string $content): array
         $textCleaned = $this->cleanContent($content, '_');
 
         // Extract text blocks.
-        if (preg_match_all('/(\sQ)?\s+BT[\s|\(|\[]+(.*?)\s*ET(\sq)?/s', $textCleaned, $matches, \PREG_OFFSET_CAPTURE)) {
+        if (preg_match_all('/(.*?)\s+BT[\s|\(|\[]+(.*?)\s*ET(?=\s|$)?/s', $textCleaned, $matches, \PREG_OFFSET_CAPTURE)) {
```
k00ni, line 194:
```suggestion
        $regex = '/(.*?)\s+BT[\s|\(|\[]+(.*?)\s*ET(?=\s|$)?/s';
        if (preg_match_all($regex, $textCleaned, $matches, \PREG_OFFSET_CAPTURE)) {
```

Makes it easier to read.

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -184,14 +184,49 @@ public function cleanContent(string $content, string $char = 'X')
...
+            }
+            $str = substr($matches[1][$pos][0], 0, $qMatches[0][0][1]);
+            if (preg_match('/\sTf(\s|$)/', $str)) {
+                $section = trim($str)."\n".$section;
+            }
+        } elseif (preg_match('/\sTf(\s|$)/', $matches[1][$pos][0])) {
+            $section = trim($matches[1][$pos][0])."\n".$section;
+        }
+
+        return $section;
```
k00ni, line 219: I moved this to a separate function. 

If possible, add the comments you made here also in code. That would bolster code readability and understanding.

k00ni:
@cris-shi What is the status here? Feel free to tell me, if you need assistance.

k00ni:
Closed for now, please reopen when you want to work on it again.

PR #546, opened 2022-06-28
Add test for the Issue#495

Reqrefusion:
I'm adding these tests in case it improves in the future. @likemusic explained how this problem got there in #495.


k00ni:
@Reqrefusion Thank you for your pull request. I would like to merge it right away, but tests fail.

Reqrefusion:
Hello, it should not pass tests anyway because it cannot parser the relevant documents correctly. It will pass this test when you do it correctly. So technically, it shouldn't have passed this test anyway. I designed this test this way, but if you say it should give an error in case of a change, I can understand. But at that time it will throw an error when an output is true. Honestly, it's a bit of a complicated problem. After all, this is not a fix but a forward-looking reminder that a problem cannot be solved. The main question is whether these reminders will be made through unit test. I'm asking you, @k00ni how it should be. I'll edit the PR based on your answer.

k00ni (changes requested):
Hello @Reqrefusion, thank you for your clarification. I follow your argument.

> The main question is whether these reminders will be made through unit test. I'm asking you, @k00ni how it should be. I'll edit the PR based on your answer.

I suggest the following:
* [ ] please split up your new test into 3, each for one PDF file. This way we can assure that there is no mixture later on, when related changes arrive into the code.
* [ ] add `$this->expectExceptionMessage('Some error message');` at the beginning of each test, with an appropriate message to catch the exception/error which is currently thrown (`iconv(): Detected an illegal character in input string`). Hope that is technically possible.
* [ ] add a short explanation why we expect this behavior, plus a reference to this PR and the issue #495

Thanks.

k00ni:
@Reqrefusion Can you tell where we stand here?

k00ni:
Closed for now, please reopen when you want to work on it again.

Reqrefusion:
> @Reqrefusion Can you tell where we stand here?

Oh sorry I didn't see it. I'm busy these days, but I'll be back for sure sometime. Thank you.

PR #559, opened 2022-11-13
consider scaling by fontSize(Tf, Tfs) and text matrix (Tm)

### LINKED ISSUE #532: Error in Y coordinate
There are pages/documents where I don't get the correct Y coordinate.
I believe that using tables (I convert html to PDF) messes with the coordinate.
An example: Im searching for the tag {f{Docente}f} in this [document](https://drive.google.com/file/d/1FFenCAVkAdy7VU16bXJL6M11jBIkLa1Q/view?usp=sharing) in white font at the bottom (next to `Firma:` and under `Juan Carlos de la Lera` just not visible) but coordinates given are x: 251.367 - y: 109.779 instead

oliver681:
closes #532 

Previously, the coordinates in getDataTm() were calculated without taking into accout the scaling that happens by the set font size and text matrix.

In a PDF that would e.g. look like:
- /R9 11.04 Tf (second operand is font size)
- 0.999402 0 0 1 70.8 698.96 Tm (first and fourth operand set horizontal and vertical scaliing)

When the posiiton is changed using the text positioning operators Td or TD, the new coordinates are now calculated with the set scaling.

One line in the test files was  changed to reflect the correct scaling.

This pull request also fixes a sign error in the calculation of the y-coordinate in TD.

### DIFF
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -677,7 +677,11 @@ class Page extends PDFObject
          * Indexes of horizontal/vertical scaling and X,Y-coordinates in the matrix (Tm)
          */
         $hSc = 0; // horizontal scaling
-        $vSc = 3; // vertical scaling
+        /**
+         * index of vertical scaling in the array that encodes the text matrix.
+         * for more information: https://github.com/smalot/pdfparser/pull/559#discussion_r1053415500
+         */
+        $vSc = 3;
         $x = 4;
         $y = 5;
 
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -612,9 +612,7 @@ class Page extends PDFObject
 
                 case 'Tf':
                 case 'TF':
-                    if ($this->config->getDataTmFontInfoHasToBeIncluded()) {
-                        $extractedData[] = $command;
-                    }
+                    $extractedData[] = $command;
                     break;
 
                     /*
@@ -673,20 +671,28 @@ class Page extends PDFObject
          *  Set default values for font data
          */
         $defaultFontId = -1;
-        $defaultFontSize = 0;
+        $defaultFontSize = 1;
 
         /*
-         * Setting where are the X and Y coordinates in the matrix (Tm)
+         * Indexes of horizontal/vertical scaling and X,Y-coordinates in the matrix (Tm)
          */
+        $hSc = 0; // horizontal scaling
+        $vSc = 3; // vertical scaling
         $x = 4;
         $y = 5;
+
+        /*
+         * x,y-coordinates of text space origin in user units
+         *
+         * These will be assigned the value of the currently printed string
+         */
         $Tx = 0;
         $Ty = 0;
 
         $Tm = $defaultTm;
         $Tl = $defaultTl;
         $fontId = $defaultFontId;
-        $fontSize = $defaultFontSize;
+        $fontSize = $defaultFontSize; // reflects fontSize set by Tf or Tfs
 
         $extractedTexts = $this->getTextArray();
         $extractedData = [];
@@ -695,11 +701,11 @@ class Page extends PDFObject
             switch ($command['o']) {
                 /*
                  * BT
-                 * Begin a text object, inicializind the Tm and Tlm to identity matrix
+                 * Begin a text object, initializing the Tm and Tlm to identity matrix
                  */
                 case 'BT':
                     $Tm = $defaultTm;
-                    $Tl = $defaultTl; // review this.
+                    $Tl = $defaultTl;
                     $Tx = 0;
                     $Ty = 0;
                     $fontId = $defaultFontId;
@@ -712,7 +718,7 @@ class Page extends PDFObject
                      */
                 case 'ET':
                     $Tm = $defaultTm;
-                    $Tl = $defaultTl;  // review this
+                    $Tl = $defaultTl;
                     $Tx = 0;
                     $Ty = 0;
                     $fontId = $defaultFontId;
@@ -720,12 +726,13 @@ class Page extends PDFObject
                     break;
 
                     /*
-                     * leading TL
+                     * text leading TL
                      * Set the text leading, Tl, to leading. Tl is used by the T*, ' and " operators.
                      * Initial value: 0
                      */
                 case 'TL':
-                    $Tl = (float) $command['c'];
+                    // scaled text leading
+                    $Tl = (float) $command['c'] * (float) $Tm[$vSc];
                     break;
 
                     /*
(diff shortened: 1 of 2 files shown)


k00ni (changes requested):
@oliver681 Thank you for taking the time and this pull request.

In the following just a few remarks and questions:

> One line in the test files was changed to reflect the correct scaling.

That is alright. Are there any edge cases now we have to take care of? I consider changing tests bad practice in general, but in this case core functional was buggy in the first place. I would welcome at least one additional test.

> This pull request also fixes a sign error in the calculation of the y-coordinate in TD.

What do you mean with "signing" here?

To fix the failing test just run PHP-CS-Fixer locally.

oliver681:
@k00ni 

> I would welcome at least one additional test.

Do you know how I can easily create a PDF manually? I was trying to write an extra test yesterday, but it is necessary to write the test file manually to ensure that Td, TD and next line operators are called to test all possible cases of position translation.

> What do you mean with "signing" here?

See line 765. The y-translation was subtracted instead of added:

`$Ty += (float) $coord[1] * (float) $fontSize * (float) $Tm[$vSc];`

> To fix the failing test just run PHP-CS-Fixer locally.

Ok, i will do that. I wasn't sure if that was right, because when I run that yesterday it changed a lot of code that I haven't touched.

oliver681:
@k00ni What about the errors phpstan findes in RawDataParser?

k00ni:
> Ok, i will do that. I wasn't sure if that was right, because when I run that yesterday it changed a lot of code that I haven't touched.

I thought about it and my proposal is to fix all problems which arise during a pull request (even though it "pollutes" your fix). So thanks for fixing it right away.

>  What about the errors PHPStan finds in RawDataParser?

We use PHPStan always in the latest stable version. A side effect is, that its getting better in finding errors 😆 Can you fix them? If not or you are unsure, I can try to help out.

> Do you know how I can easily create a PDF manually? I was trying to write an extra test yesterday, but it is necessary to write the test file manually to ensure that Td, TD and next line operators are called to test all possible cases of position translation.

Please try to build a basic test around `Page.php::getDataTm()`. You could make a faulty `$dataCommands`-instances to trigger your code part and see how it reacts. If that type of testing is new to you, give me a ping, I will give you an example.

ilconfo:
Hello,
I was working on some other issue and I've found a minus sign issue also under the T* case, I think it should not be:

`$Ty -= $Tl;`

but

`$Ty += $Tl;`

oliver681:
@ilconfo No, in this case the minus is right. The text leading is given as a positive number, but the y-coordinate in a pdf is from bottom to top, so the coordinate must be decreased by Tl to obtain the new y-position.

But the comment above is actually wrong. It should be changed `0 -Tl Td` instead of '0 Tl Td'.

See also here int the PDF specification:

grafik

k00ni:
@oliver681 can you tell me the status here? Any plans to finish this soon? If you need help, let me know.

oliver681:
@k00ni 

> can you tell me the status here? Any plans to finish this soon? If you need help, let me know.

I was busy unfortunately. I hope I can write the test case in the next days.

oliver681:
@k00ni Sorry, I won't have time to work on this until 8th December.

oliver681:
@k00ni 

Finally I can finish this :)

Inside getDataCommands, `TF` and `Tf` (`Tfs` is missing, I will implement that) might be igored by the following setting . This will lead to wrong positions by default (default value = false) if font size is changed by these operators.

Is there a good reason why to exclude these commands? Otherwise I would suggest removing this if-statement at that place (getDataCommands) and only leave it within getDataTm(), thus resulting in the desired behaviour (only including font information in dataTm if set to true, but still returning correct coordinates).

grafik

oliver681:
@k00ni 

The line was change by shtayerc*.

grafik

As far as I can see, removing it won't affect functionality, because it just affects the data commands returned by get DataCommands() but the font info still is ignored in get DataTm().

> Which file do you mean?

Page.php:617

k00ni:
@shtayerc I would really appreciate if you could help us out here. There is a question about code you pushed in the past: https://github.com/smalot/pdfparser/pull/559#issuecomment-1354655993

oliver681:
@k00ni When running phpunit locally, i get the following errors (always the same error):

grafik

Any idea how I can get rid of them?

shtayerc:
I quickly checked the code. If statement is probably there to keep the same behavior as before and to pass tests for getDataCommands (https://github.com/smalot/pdfparser/blob/988efcba8d9e4127691f88aac0245ea90b795607/tests/Integration/PageTest.php#L259). I did not put that much thought in that. The correct way was to remove the if and fix tests.

k00ni:
@shtayerc thank you for your fast response!

> Any idea how I can get rid of them?

Did you have `iconv` PHP-extension installed? https://www.php.net/manual/en/book.iconv.php

k00ni:
If I see it correctly, you force-pushing and remove/replace older commits. Thats fine, but if you only do it so that we have a clean history, dont worry, I will squash all commits when merging.

oliver681:
@k00ni It will be only one commit anyway ;) I needed to force push because I haven't got phpunit to work locally to see if my tests are correct or not. But now I got it to work locally. I'm adding another test case and I think I'll be ready soon ;)

src/Smalot/PdfParser/Page.php
```diff
@@ -673,20 +673,28 @@ public function getDataTm(array $dataCommands = null): array
...
         $defaultFontId = -1;
-        $defaultFontSize = 0;
+        $defaultFontSize = 1;
 
         /*
-         * Setting where are the X and Y coordinates in the matrix (Tm)
+         * Setting where the scaling and X,Y-coordinates in the matrix are (Tm)
          */
+        $hSc = 0; // horizontal scaling
+        $vSc = 3; // vertical scaling
```
k00ni, line 682: Out of curiosity, why using `3` as default value for vertical scaling? Is there a reference in the PDF spec for instance?

k00ni (commented):
@oliver681 thank you for taking the effort! It looks good to me, I just have one question (see below).

src/Smalot/PdfParser/Page.php
oliver681: @k00ni It is not the value of the vertical scaling, but the index of vertical scaling the  in the array that encodes the text matrix.

grafik

So it corresponds to `d` here.

oliver681:
@k00ni I forgot to exclude the if-statement we talked about above. Now its removed and I adapted the corresponding test cases.

src/Smalot/PdfParser/Page.php
```diff
@@ -673,20 +671,28 @@ public function getDataTm(array $dataCommands = null): array
...
         $defaultFontId = -1;
-        $defaultFontSize = 0;
+        $defaultFontSize = 1;
 
         /*
-         * Setting where are the X and Y coordinates in the matrix (Tm)
+         * Indexes of horizontal/vertical scaling and X,Y-coordinates in the matrix (Tm)
          */
+        $hSc = 0; // horizontal scaling
+        $vSc = 3; // vertical scaling
```
k00ni, line 680:
```suggestion
        /**
         * index of vertical scaling in the array that encodes the text matrix.
         * for more information: https://github.com/smalot/pdfparser/pull/559#discussion_r1053415500
         */
        $vSc = 3;
```

k00ni: Good info, I would like to add this explanation to the code.

k00ni (approved):

k00ni:
@oliver681 I appreciate your work here, thanks.

I will merge it probably today/tomorrow and add it to a Christmas release.

Issue #557, opened 2022-10-25
I have a error when i use parseFile function


DogLoc:
- PHP Version: 7.4
 - PDFParser Version: 2.2

### Description:

When i use $parser->parseFile($pathToPdf) I got this :

```
Argument 1 passed to Smalot\PdfParser\Parser::parseHeader() must be of the type array, string given, called in /path/to/project/vendor/smalot/pdfparser/src/Smalot/PdfParser/Parser.php on line 266
```

### Code
```
$parser = new PdfParser();
$pdf = $parser->parseFile($path);
```

k00ni:
Can you provide the PDF you are using?

DogLoc:
I can send you this pdf in private ? Because is not a public document...

DogLoc:
> Can you provide the PDF you are using?

I can send you this pdf in private ? Because is not a public document...

k00ni:
> I can send you this pdf in private ? Because is not a public document...

No. I would appreciate, if you find some time to provide a basic exploit. Can you to get an example string which triggers this problem? Add an echo or var_dump right before the faulty line, then paste the raw string here. This way we can reproduce the error and add it to our test environment. Thanks.

DogLoc:
No problem, here is the result of my dump before the error was triggered:

```
^ ""
^ Smalot\PdfParser\Document^ {#656
  #objects: []
  #dictionary: []
  #trailer: Smalot\PdfParser\Header^ {#640
    #document: Smalot\PdfParser\Document^ {#656}
    #elements: []
  }
  #details: null
}
```

First we have an empty string (cause of the problem) 
Second a `Smalot\PdfParser\Document` object type


dump at line 266 in `parseHeaderElement` function

```
dump($value, $document);
$header = $this->parseHeader($value, $document);
```

The empty string cause this error because the `parseHeader()` function take a `array` type argument for `array $structure` parameter
but I pass a empty string

DogLoc:
> > I can send you this pdf in private ? Because is not a public document...
> 
> No. I would appreciate, if you find some time to provide a basic exploit. Can you to get an example string which triggers this problem? Add an echo or var_dump right before the faulty line, then paste the raw string here. This way we can reproduce the error and add it to our test environment. Thanks.

I asked my client and finally I can provide you the pdf
[20221025_debug_header_parse.pdf](https://github.com/smalot/pdfparser/files/9868222/20221025_debug_header_parse.pdf)

PR #560, opened 2022-11-23
fix: allow to parse null or empty header element

DogLoc:
## Why I open this pull request ?

When I pass a empty string value at the first arguments of ``parseHeader`` function, it throw an exception :   

``` 
Argument 1 passed to Smalot\PdfParser\Parser::parseHeader() must be of the type array, string given, called in /path/to/project/vendor/smalot/pdfparser/src/Smalot/PdfParser/Parser.php on line 266
```

that append when I provide a PDF like this : [20221025_debug_header_parse.pdf](https://github.com/smalot/pdfparser/files/9868222/20221025_debug_header_parse.pdf)

If it helps, I have already opened a case for this problem. : https://github.com/smalot/pdfparser/issues/557

Have a good day ^^


k00ni (changes requested):
@DogLoc thank you for this fix. I have a few things:

* [ ] Can you move the new if-clause before the switch-clause? I can not see why this check is only relevant for the `>>` case.
* [ ] Please provide at least one test. If you need help here, let me know.
* [ ] Please fix failing CS workflow by running PHP-CS-Fixer.
* [x] Out of curiosity: why not return `ElementNull` instead? Why do you use `null`?

DogLoc:
> @DogLoc thank you for this fix. I have a few things:
> 
> * [ ]  Can you move the new if-clause before the switch-clause? I can not see why this check is only relevant for the `>>` case.
> * [ ]  Please provide at least one test. If you need help here, let me know.
> * [ ]  Please fix failing CS workflow by running PHP-CS-Fixer.
> * [ ]  Out of curiosity: why not return `ElementNull` instead? Why do you use `null`?

Finally I use an empty array for input and output (<<, >>) because with the ElementNull class, the code tries to retrieve the header on a null element I don't have strong knowledge of this project, so if you have a better suggestion, I take it! ^^

I have provided a unit test for the parser, let me know if this is sufficient.

DogLoc:
@k00ni master merged in my branch 👍

k00ni (changes requested):
Thank you for your changes! There is only one thing and we are good to go.

src/Smalot/PdfParser/Parser.php
```diff
@@ -262,6 +262,11 @@ protected function parseHeader(array $structure, ?Document $document): Header
      */
     protected function parseHeaderElement(?string $type, $value, ?Document $document)
     {
+        $valueIsEmpty = null == $value || '' == $value || false == $value;
+        if (('<<' === $type || '>>' === $type) && $valueIsEmpty)) {
```
k00ni, line 266:
```suggestion
        if (('<<' === $type || '>>' === $type) && $valueIsEmpty) {
```
My bad, I missed a `)`.

k00ni:
Looks good to me. I will merge it next week if there are no objections.

Thank you @DogLoc!

k00ni (approved):

DogLoc:
Thanks you too  ^^ ! @k00ni

Issue #561, opened 2022-11-23
Only get limited number of pages from PDF


alesrebec:
In Document.php we have this:

```
    public function getText(): string
    {
        $texts = [];
        $pages = $this->getPages();

        foreach ($pages as $index => $page) {
```

Can we change getText() to accept a numeric value and limit the returned string to get only the first X pages (or less if the document has less than X pages).
It's quite a useful feature, especially when you try to detect the language of the document and use this library. The documents can be hundreds of pages long but you only need the first page or first 5 to determine the language.

I can write a PR if this would be something that can be accepted by the maintainers.

k00ni:
Hello @alesrebec, one of the maintainers here. The feature you are describing sounds quite useful indeed.

> I can write a PR if this would be something that can be accepted by the maintainers.

I suggest you do that and we discuss details there. IMHO there is nothing which speaks against this, but I wanna make sure it fits in our current code base.

PR #562, opened 2022-11-24
#561 Added optional getText() argument to return limited number of document pages if set

alesrebec:
Solving #561 
Tested locally and working fine.


src/Smalot/PdfParser/Document.php
```diff
@@ -264,11 +264,16 @@ public function getPages()
         throw new \Exception('Missing catalog.');
     }
 
-    public function getText(): string
+    public function getText($limit = NULL): string
```
k00ni, line 267:
```suggestion
    public function getText(int|null $pageLimit = null): string
```

src/Smalot/PdfParser/Document.php
```diff
@@ -264,11 +264,16 @@ public function getPages()
...
     {
         $texts = [];
         $pages = $this->getPages();
 
+        // Only use the first X number of pages if $limit is set and numeric.
+        if (!empty($limit) && is_numeric($limit)) {
```
k00ni, line 273:
```suggestion
        if (is_int($pageLimit) && 0 < $pageLimit) {
```

k00ni (changes requested):
@alesrebec thank you for this nice addition.

Please provide at least one test which covers your code.

k00ni:
@alesrebec I took the liberty and finalized the PR. Any remarks?

I will merge it in a couple of days if there are no objections.

alesrebec:
> @alesrebec I took the liberty and finalized the PR. Any remarks?
> 
> I will merge it in a couple of days if there are no objections.

Thanks for finalizing the PR.
I had some issues running the tests locally when I was about to write the tests and it just slipped from my todo list.
All good with the merge.

k00ni (approved):

PR #566, opened 2022-12-13
Extend Github workflow file to run tests in Windows environment

k00ni:
We can run almost all tests in Windows, but a few fail because of different line endings (`\n` in Linux **vs.** `\r\n` in Windows). I marked them with `@group linux-only` so they can be skipped in Windows.

Current state is pretty good for now in my opinion and helps us to catch more (potential) problems beforehand.

What do you guys think?

CC @j0k3r


j0k3r (approved):
Good job, looks good to me 👍🏼

PR #582, opened 2023-03-08
Reduce excessive memory allocations

se-ti:
Hi!
I ran into issue, that extracting text from ~20Mb pdfs (~300K text, lots of images), fails with memory allocation error at 128M limit.
I looked through the code and found several unlimited substring allocations and fixed them.
I've replaced couple of regexps with a bit more effective code also.

Hope, You'd accept my PR.

You can test my fixes with such a pdf https://westra.ru/reports/kavkaz/danilina-1el2-arhyz-2021.pdf and this code:

```php
$lim = '128M';

$text = $pdf->getText();

echo $text;
```
Given file required ~90Mb RAM for the first launch and ~45Mb for the best subsequent ones.
After the fixes it requires 45Mb for the first launch and just ~25Mb for the best subsequent ones.


k00ni:
@se-ti Thank you for your pull request.

I am especially interested in optimization-related pull requests. Could you elaborate at bit what your changes do?

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -762,17 +755,18 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
                     // start stream object
                     $objtype = 'stream';
                     $offset += 6;
-                    if (1 == preg_match('/^([\r]?[\n])/isU', substr($pdfData, $offset), $matches)) {
+                    if (1 == preg_match('/^([\r]?[\n])/isU', substr($pdfData, $offset, 4), $matches)) {
                         $offset += \strlen($matches[0]);
```
se-ti, line 759: if this stream object appears in the beginning of 20M pdf this unlimited substr results in almost 20M RAM allocation.

While this regexp can match only 1 or 2 character string

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -762,17 +755,18 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
...
                         $pregResult = preg_match(
                             '/(endstream)[\x09\x0a\x0c\x0d\x20]/isU',
-                            substr($pdfData, $offset),
+                            $pdfData,
                             $matches,
```
se-ti, line 763: Same case here. Instead of getting potentially huge substring just to feed it to regexp we can just use preg_match's offset parameter.

But using offset parameter changes offset arithmetics. So, we gonna repair it the nex few lines.

k00ni:
Thank you for your fast response. I didn't check your code in detail, but was wondering in which instances your adaptions change code semantics (and implications for existing programs using it).

Are there any edge cases we have to keep in mind here?

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -888,7 +882,7 @@ public function parseData(string $data): array
         }
 
         // get PDF content string
-        $pdfData = substr($data, $trimpos);
+        $pdfData = $trimpos > 0 ? substr($data, $trimpos) : $data;
```
se-ti, line 886: Be reasonably lazy: in most cases $trimpos would be zero. If you should do nothing -- do nothing.

Probably, It would be better to use $offset parameter of the getXrefData($pdfData); instead of substring. But I'm afraid of this comment:
* @param int   $offset xref offset (if known)
and strange if in the protected function getXrefData
```php
if (0 == $offset) {
            // find last startxref
```
Seems, zero offset is really a special case.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -723,15 +719,12 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
...
                     ++$offset;
-                    $pregResult = preg_match(
-                        '/^([0-9A-Fa-f\x09\x0a\x0c\x0d\x20]+)>/iU',
-                        substr($pdfData, $offset),
-                        $matches
-                    );
-                    if (('<' == $char) && 1 == $pregResult) {
+
+                    $span = strspn($pdfData, "0123456789abcdefABCDEF\x09\x0a\x0c\x0d\x20", $offset);
+                    if (('<' == $char) && $span > 0 && @$pdfData[$offset+$span] == '>') {
```
k00ni, line 724: Why using the `@` character here? I know it being used with functions to suppress PHP warnings/errors.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -635,14 +635,10 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
                 // name object
                 $objtype = $char;
                 ++$offset;
-                $pregResult = preg_match(
-                    '/^([^\x00\x09\x0a\x0c\x0d\x20\s\x28\x29\x3c\x3e\x5b\x5d\x7b\x7d\x2f\x25]+)/',
```
k00ni, line 639: If I understand this part right, you restrict the range from "all string" to "sub string with offset $offset and 256 length". Is that correct?

src/Smalot/PdfParser/RawData/RawDataParser.php
se-ti, line 724: Common case: unlimited substring (and a regexp) just to test, that string consists of 20 allowed chars followed by >.
That's right what strspn with $offset parameter do, and one small if do.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -635,14 +635,10 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
...
-                    substr($pdfData, $offset, 256),
-                    $matches
-                );
-                if (1 == $pregResult) {
-                    $objval = $matches[1]; // unescaped value
-                    $offset += \strlen($objval);
+                $span = strcspn($pdfData, "\x00\x09\x0a\x0c\x0d\x20\n\t\r\v\f\x28\x29\x3c\x3e\x5b\x5d\x7b\x7d\x2f\x25", $offset, 256);
+                if ($span > 0) {
```
se-ti, line 639: That's first in the code, but it was the last case i discovered.
We get not a tiny substring (and allocate memory) just to feed it to regexp.
And we use regexp just to get longest string, withous prohibited ones.

That's what strcspn with offset param do! (and max len, if you really need it)

Note, that 0x20\s fragment in regexp has transformed to a list of prohibited chars: \x20\n\t\r\v\f

se-ti:
> Thank you for your fast response. I didn't check your code in detail, but was wondering in which instances your adaptions change code semantics (and implications for existing programs using it).
> 
> Are there any edge cases we have to keep in mind here?
I didn't change semantics of any method.
Therefore i suppose my changes are well-isolated and won't change behaviour of existsing programs.
Except for reduced memory consumption ^)

k00ni:
> I didn't change semantics of any method. Therefore i suppose my changes are well-isolated and won't change behaviour of existsing programs.

Thank you for these detailled and quick answers! It will help to better understand these changes. There are over 1000 projects using this library, thats why I wanted to check these points.
I am currently busy, but will try to find some time to get back to you. If someone else wants to bring this PR forward, feel free though!

k00ni:
Can you please merge in master branch. Latest changes should fix all the failing tests.

se-ti:
Is this problem PR problem, or tests problem?

It doesn't like non-yoda-style condition, no space around +, @, or what?
$span > 0 && @$pdfData[$offset+$span] == '>'

k00ni:
> [...] is current problem PR problem, or tests problem?

To my knowledge, PHPUnit 10.0 "wrecked" some tests which also impedes this PR. It was not your fault.

> It doesn't like non-yoda-style condition, no space around +, @, or what? $span > 0 && @$pdfData[$offset+$span] == '>'

I have to check that. Usually it is enough to run our PHP-CS-Fixer instance and it fixes all. Did you run `make run-php-cs-fixer`?

> P.S. Is it reliable to use object header's field /Length when parsing stream-endstream section?

I can't answer that. Can you provide more context?

se-ti:
> I have to check that. Usually it is enough to run our PHP-CS-Fixer instance and it fixes all. Did you run `make run-php-cs-fixer`?

No, i didn't.
> > P.S. Is it reliable to use object header's field /Length when parsing stream-endstream section?
> 
> I can't answer that. Can you provide more context?

I've already found its usages in existing code :), so i'll use it too.

https://github.com/smalot/pdfparser/blob/3ef8bc505304f94c7c7075fe2644ebed29f8ff18/src/Smalot/PdfParser/RawData/RawDataParser.php#L562
https://github.com/smalot/pdfparser/blob/3ef8bc505304f94c7c7075fe2644ebed29f8ff18/src/Smalot/PdfParser/RawData/RawDataParser.php#L100 

Waiting for this PR to be merged in master to suggest another one.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -723,15 +719,12 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
...
                     ++$offset;
-                    $pregResult = preg_match(
-                        '/^([0-9A-Fa-f\x09\x0a\x0c\x0d\x20]+)>/iU',
-                        substr($pdfData, $offset),
-                        $matches
-                    );
-                    if (('<' == $char) && 1 == $pregResult) {
+
+                    $span = strspn($pdfData, "0123456789abcdefABCDEF\x09\x0a\x0c\x0d\x20", $offset);
+                    if ('<' == $char && $span > 0 && '>' == @$pdfData[$offset + $span]) {
```
k00ni, line 724: Your code is fine, but I would like to make it a bit more explicit. This way one is less likely to overlook the `@` in the if clause in my opinion.

```suggestion
                    $dataToCheck = $pdfData[$offset + $span] ?? null;
                    if ('<' == $char && $span > 0 && '>' == $dataToCheck) {
```

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -762,17 +755,18 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
                     // start stream object
                     $objtype = 'stream';
                     $offset += 6;
-                    if (1 == preg_match('/^([\r]?[\n])/isU', substr($pdfData, $offset), $matches)) {
+                    if (1 == preg_match('/^([\r]?[\n])/isU', substr($pdfData, $offset, 4), $matches)) {
```
k00ni, line 758: Can you give me a short hint, why you added `4` (= length of substring) here?

k00ni (commented):
Looks good to me, only have a few remarks.

Any feedback @j0k3r @smalot?

k00ni:
Just curios, did you also try with https://github.com/smalot/pdfparser/blob/master/doc/CustomConfig.md#option-setdecodememorylimit--setretainimagecontent-manage-memory-usage?

src/Smalot/PdfParser/RawData/RawDataParser.php
se-ti: I'm not strong with inner structure of php string-like objects, their byteness, whether they are null-terminated or not (looks like they are not).

Since this regexp can match up to 2 character, 4 is 2 char * up to 2 bytes/char.
Probably it should be 2 insead of 4.

k00ni: > I'm not strong with inner structure of php string-like objects, their byteness, whether they are null-terminated or not (looks like they are not).
Since this regexp can match up to 2 character, 4 is 2 char * up to 2 bytes/char.
Probably it should be 2 insead of 4.

I am not sure if I can follow your argument. The additional `4` is part of this pull request, why was it added? I don't want to split hairs here, just to make sure there are no drive-by changes were not meant to be part of the PR in the end.

k00ni:
> What should i do next? What's the profit?

We introduce these optimizations with #476 and #441, whereas #441 is about the images. As far as I understood these PRs, these options allow to restrict certain (PHP-) functions and use less ressources.

src/Smalot/PdfParser/RawData/RawDataParser.php
se-ti: Sorry, i was sure i have answered this question earlier:

Imagine this stream object appears in the beginning of 20M pdf. If so, this unlimited substr results in almost 20M RAM allocation.

And this regexp can match only 1 or 2 first characters of the string. 
That's why we can take no more than 4 bytes of $pdfData and match regexp against them.

k00ni: My bad, I overlooked it. Thanks for the clarification.

j0k3r (commented):
AFAIK the code was already too complex for me to understand it properly.
If tests are green and OP fixed a memory leak, I'm 👍🏼

k00ni (approved):

k00ni:
Thank you for your work here @se-ti.

Issue #581, opened 2023-02-23
Question: Group text by Paragraphs?


inpresif:
Hi, I'm trying to group each block/paragraph in a PDF as a oneliner in a textfile. Is this possible?

For some reason I get a funky result now on a test PDF, where text inside a paragraph block is exported as newlines, or even broken down in the middle (as with the URL). 
Also, it is it possible to Skip parts? I'd love to not have the talke of contents  as I want to use the scanned docs like in a Python vector seach. Hopefully there's some tips that arent in the docs (or maybe a link to extended docs if/where available?).

See image below:
![image](https://user-images.githubusercontent.com/12870992/220909677-756f54fb-5024-46ba-82c0-ad778bc0f909.png)
![image](https://user-images.githubusercontent.com/12870992/220910228-cc2306ce-2779-4b9f-b70b-f94018e96c50.png)

I used this code:

```
            $parser        = new Parser();
            $pdf           = $parser->parseFile($file_info);
            $pdf_contents = $pdf->getText();
            $file_contents = preg_replace(array("/(\r\n|\r|\n)/", "/\s+/"), array(" ", " "), $pdf_contents);
            $fileAsArray   = array($file_contents);
```

PR #583, opened 2023-03-09
Fixes failing tests; reorganized test files (PHPUnit 10.x related)

k00ni:
Since version 10, PHPUnit complains about:

1) Your XML configuration validates against a deprecated schema. Migrate your XML configuration using "--migrate-configuration"!

2) Class AltAutoloadTest cannot be found in
D:\a\pdfparser\pdfparser\tests\AltAutoloadTest.php

3) Class Tests\Smalot\PdfParser\Performance\Test\AbstractPerformanceTest cannot be found in
D:\a\pdfparser\pdfparser\tests\Performance\Test\AbstractPerformanceTest.php

4) Class
Tests\Smalot\PdfParser\Performance\Test\DocumentDictionaryCacheTest cannot be found in
D:\a\pdfparser\pdfparser\tests\Performance\Test\DocumentDictionaryCacheTest.php

The first one is easy to solve.

But the other 3 are confusing, because PHPUnit 9 did not complain at all. PHPUnit v10 doesn't like abstract Test classes or files (containing Test in the name), which contain no class at all. Source: https://github.com/sebastianbergmann/phpunit/issues/4979

To mitigate these warnings and to make maintenance easier in the future, I reorganized our test files and put them in different folders. PHPUnit tests are separated from our other tests.

#581 and #582 will be processed after this one is merged.


j0k3r (commented):
Maybe we should lock the PHPUnit version, see https://thephp.cc/articles/the-death-star-version-constraint

k00ni:
> Maybe we should lock the PHPUnit version, see https://thephp.cc/articles/the-death-star-version-constraint

I would advise against that, because it only delays required changes. I prefer seeing problems as soon as possible. Maybe there is some compromise here?

j0k3r:
Well, it's a way to do it.
Keeping unlocked version means when PHPUnit 11 will be released someone will have to quickly fix issues (if any) otherwise all PRs will fail.
Keeping a locked version avoid that problem. And then, we can smoothly update PHPUnit later without hurry

k00ni:
Good points, I implemented a lock.

What doyou think about the rest?

dev-tools/composer.json
```diff
@@ -4,6 +4,6 @@
         "friendsofphp/php-cs-fixer": "^3",
         "phpstan/phpstan": "^1",
         "phpstan/phpstan-phpunit": "^1",
-        "phpunit/phpunit": ">=7.5"
+        "phpunit/phpunit": ">=7.5 <10.0"
```
j0k3r, line 7: Shouldn't this be like that?
```suggestion
        "phpunit/phpunit": ">=7.5 <11.0"
```

j0k3r (commented):
The rest looks ok 👍🏼

PR #590, opened 2023-03-27
Improve performance by not extracting compressed image data if retainImageContent was set to false

se-ti:
One of previous pull requests allowed not to decompress image data, if the user has set retainImageContent flag to false.
This pull request suggests even not to extract compressed image data, if this flag was set to false.

Minor optimization: extracting stream data tries to use header information about stream's length to speed preg_match up.

This PR improves memory allocation, achieved with [pull 582](https://github.com/smalot/pdfparser/pull/582),
from 45M for the first launch, 25M for the best subsequent one 
to 43M for the first launch, 18M for the best subsequent.

https://westra.ru/reports/kavkaz/danilina-1el2-arhyz-2021.pdf
```php
$lim = '128M';

$text = $pdf->getText();

echo $text;
```

### DIFF
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -805,20 +805,25 @@ class RawDataParser
     }
 
     /**
+     * Get value of an object header's section
+     *
      * @param array|null $header obj's header, parsed by getRawObject
-     * @param string $key Header's section
+     * @param string $key header's section name
      * @param string $type type of the section (ie 'numeric', '/', '<<', etc.)
      * @param string|null $default default value for header's section
-     * @return string|null value of obj header's section, or default value if none found
+     *
+     * returns string|array|null value of obj header's section, or default value if none found, or its type doesn't match $type param
      */
-    protected function getHeaderValue(?array $headerDic, string $key, string $type, ?string $default = ''): ?string
+    protected function getHeaderValue(?array $headerDic, string $key, string $type, $default = '')
     {
-        if (!is_array($headerDic))
+        if (!is_array($headerDic)) {
             return $default;
+        }
 
         foreach ($headerDic as $i => $val) {
-            if (is_array($val) && 3 == count($val) && '/' == $val[0] && $val[1] == $key && isset($headerDic[$i + 1]))
-                return is_array($headerDic[$i + 1]) && 1 < count($headerDic[$i + 1]) && $type == $headerDic[$i+1][0] ? $headerDic[$i + 1][1] : $default;
+            if (is_array($val) && 3 == count($val) && '/' == $val[0] && $val[1] == $key && isset($headerDic[$i + 1])) {
+                return is_array($headerDic[$i + 1]) && 1 < count($headerDic[$i + 1]) && $type == $headerDic[$i + 1][0] ? $headerDic[$i + 1][1] : $default;
+            }
         }
 
         return $default;
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -557,7 +557,7 @@ class RawDataParser
         do {
             $oldOffset = $offset;
             // get element
-            $element = $this->getRawObject($pdfData, $offset, $header[1]);
+            $element = $this->getRawObject($pdfData, $offset, $header != null ? $header[1] : null);
             $offset = $element[2];
             // decode stream using stream's dictionary information
             if ($decoding && ('stream' === $element[0]) && $header != null) {
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -763,7 +763,7 @@ class RawDataParser
                         $offset += \strlen($matches[0]);
 
                         $streamLen = (int) $this->getHeaderValue($headerDic, 'Length', 'numeric', 0);
-                        $skip = !$this->config->getRetainImageContent() && 'XObject' == $this->getHeaderValue($headerDic, 'Type', '/') && 'Image' == $this->getHeaderValue($headerDic, 'Subtype', '/');
+                        $skip = false === $this->config->getRetainImageContent() && 'XObject' == $this->getHeaderValue($headerDic, 'Type', '/') && 'Image' == $this->getHeaderValue($headerDic, 'Subtype', '/');
 
                         $pregResult = preg_match(
                             '/(endstream)[\x09\x0a\x0c\x0d\x20]/isU',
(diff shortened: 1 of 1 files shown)


se-ti:
Sorry, i don't have PHPStan installed and set up.
What is the problem with it?

How can i interpret its output?
```
Loaded config default from "/home/runner/work/pdfparser/pdfparser/.php-cs-fixer.php".
................F................................................ 65 / 72 ( 90%)
.......                                                           72 / 72 (100%)
Legend: .-no changes, F-fixed, S-skipped (cached or empty file), I-invalid file syntax (file ignored), E-error
   1) src/Smalot/PdfParser/RawData/RawDataParser.php (modernize_types_casting, method_argument_space, no_unneeded_control_parentheses, no_superfluous_phpdoc_tags, native_function_invocation, yoda_style, phpdoc_align)
```

k00ni:
> How can i interpret its output?

PHP-CS-Fixer complains the file. Give me a day or two, if you don't want to install PHP-CS-Fixer locally (to fix it).

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -758,15 +761,20 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
                     $offset += 6;
                     if (1 == preg_match('/^([\r]?[\n])/isU', substr($pdfData, $offset, 4), $matches)) {
                         $offset += \strlen($matches[0]);
+
+                        $streamLen = (int) $this->getHeaderValue($headerDic, 'Length', 'numeric', 0);
```
k00ni, line 765: Please add one or more sentences about this new code.

k00ni (changes requested):
_Sorry, I didn't had the time to answer properly. Now I have._

First of all, thank you again for your work @se-ti.

I pushed a few changes fixing coding style issues (the ones you mentioned) and made the new method `getRawObject` private.

Further remarks in code.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -758,15 +761,20 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
...
+                        $skip = !$this->config->getRetainImageContent() && 'XObject' == $this->getHeaderValue($headerDic, 'Type', '/') && 'Image' == $this->getHeaderValue($headerDic, 'Subtype', '/');
```
k00ni, line 766: Replace the first part (`!$this->...`) with something like `false === `. It is more readable and will also improve maintenance in the future.

src/Smalot/PdfParser/RawData/RawDataParser.php
se-ti: If you look through PDFs, you'll find that stream blocks (stream-endstream) are anticipated with header section (<< >>) with a field Length, that contains length of this stream.

This fact is already used in protected function decodeStream
https://github.com/smalot/pdfparser/blob/65bdad55246ec230a055fbda501c6fed592c1942/src/Smalot/PdfParser/RawData/RawDataParser.php#L100

So we can extract this length, add it to preg_match's offset and:
-- help preg_match to test less data, i.e. work faster
-- prevent (not very probable) false match, when compressed data contains 'endstream ' by itself

src/Smalot/PdfParser/RawData/RawDataParser.php
se-ti: done

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -796,6 +804,38 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
...
+                \is_array($val)
+                && 3 == \count($val)
+                && '/' == $val[0]
+                && $val[1] == $key
+                && isset($headerDic[$i + 1])
+            ) {
+                return \is_array($headerDic[$i + 1]) && 1 < \count($headerDic[$i + 1]) && $type == $headerDic[$i + 1][0]
+                    ? $headerDic[$i + 1][1]
+                    : $default;
+            }
```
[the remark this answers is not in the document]
se-ti: This is a utility function, that provides access to object header's fields (obj << YYY >> part ), like Header::get('...')->getContent() do.
The only difference -- it can be used during the parsing process, when no Smalot\PdfParser\Header objects are created yet.

It recieves dictionary of header fields, as it is returned by RawDataParser::getRawObject,
iterates over it, searching for section of type '/' whith requested key.
If such a section is found, it tries to receive it's value (next object in dictionary), returning it, if it matches requested type, or default value otherwise.

This function can simplify index arithmetic here: 
https://github.com/smalot/pdfparser/blob/65bdad55246ec230a055fbda501c6fed592c1942/src/Smalot/PdfParser/RawData/RawDataParser.php#L262
It's just $valid_crs = 'XRef' == $this->getHeaderValue($sdik, 'Type', '/');
and several ifs below this line;

Here:
https://github.com/smalot/pdfparser/blob/65bdad55246ec230a055fbda501c6fed592c1942/src/Smalot/PdfParser/RawData/RawDataParser.php#L100
And there:
https://github.com/smalot/pdfparser/blob/65bdad55246ec230a055fbda501c6fed592c1942/src/Smalot/PdfParser/RawData/RawDataParser.php#L107

Answering your questions:

1. I don't see any more edge cases, we have to test, since i'm testing correctness of inner structure of an object, returned by RawDataParser::getRawObject (those is_array, count() >= X). Even this may be a bit excessive, but let it be :)
 
2. It's just a separation of responsibility, nothing more. if ensures, that desired field is found and it has some value, ternary operator ensures that value object is correct and has desired type.
It could be reduced to one if, if you insist.

se-ti:
>  I pushed a few changes fixing coding style issues...
Thanks a lot, @k00ni !

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -796,6 +804,39 @@ protected function getRawObject(string $pdfData, int $offset = 0): array
...
+     *
+     * @param string            $key     header's section name
+     * @param string            $type    type of the section (i.e. 'numeric', '/', '<<', etc.)
+     * @param string|array|null $default default value for header's section
+     *
+     * @return string|array|null value of obj header's section, or default value if none found, or its type doesn't match $type param
+     */
+    private function getHeaderValue(?array $headerDic, string $key, string $type, $default = '')
+    {
+        if (!\is_array($headerDic)) {
```
k00ni, line 818:
```suggestion
        if (false === \is_array($headerDic)) {
```

Does it make sense to you to expand this by including a check that `$headerDic` is not empty?

se-ti: I don't see any sense to add this empty check. 
If it's empty, foreach will just do nothing and function will return $default at the last line.

P.S. A comment in [doc](https://www.php.net/manual/en/language.operators.logical.php#90386):
> Evaluation of logical expressions is stopped as soon as the result is known.

It's too powerful way of optimization to ignore it

se-ti:
I think, I have added all the requested changes.
Should i do anything else?
Any other requests?
@k00ni

se-ti:
Аny news, @k00ni?

k00ni (approved):
@se-ti I am very busy currently, but I will try to squeeze in a few minutes here and there. Thank you for bearing with me.

I added a few comments to the code based on your latest feedback. My knowledge in this topic is limited so I will have to trust you on this one. It looks clean and results speak for themselves.

I will leave this open for a few days to give others a chance to object.

se-ti:
Thank You!
Waiting for this PR to be merged, to suggest another one.

PR #595, opened 2023-04-14
Reduce excessive substring allocation, using preg_match's offset parameter

se-ti:
This PR proposes to use preg_match's offset parameter instead of preparing special string for it.
This change requires minor changes of regexps: "match start of string" meta-character ^ should be replaced with \G escape sequence https://www.php.net/manual/en/regexp.reference.escape.php

This PR slightly improves memory allocation, achieved with pulls https://github.com/smalot/pdfparser/pull/582, https://github.com/smalot/pdfparser/pull/590, but would be useful for all PDF's and not only image-rich ones.

This PR shows 
42M for the first launch, 18M for the best subsequent vs
43M for the first launch, 18M for the best subsequent for pull 590

https://westra.ru/reports/kavkaz/danilina-1el2-arhyz-2021.pdf
```php
$lim = '128M';

$text = $pdf->getText();

echo $text;
```

### DIFF
--- a/src/Smalot/PdfParser/Element.php
+++ b/src/Smalot/PdfParser/Element.php
@@ -107,7 +107,7 @@ class Element
             $old_position = $position;
 
             if (!$only_values) {
-                if (!preg_match('/\G\s*(?P<name>\/[A-Z0-9\._]+)(?P<value>.*)/si', $content, $match, 0, $position)) {
+                if (false === preg_match('/\G\s*(?P<name>\/[A-Z0-9\._]+)(?P<value>.*)/si', $content, $match, 0, $position)) {
                     break;
                 } else {
                     $name = ltrim($match['name'], '/');
--- a/src/Smalot/PdfParser/Element.php
+++ b/src/Smalot/PdfParser/Element.php
@@ -107,7 +107,7 @@ class Element
             $old_position = $position;
 
             if (!$only_values) {
-                if (!preg_match('/^\s*(?P<name>\/[A-Z0-9\._]+)(?P<value>.*)/si', substr($content, $position), $match)) {
+                if (!preg_match('/\G\s*(?P<name>\/[A-Z0-9\._]+)(?P<value>.*)/si', $content, $match, 0, $position)) {
                     break;
                 } else {
                     $name = ltrim($match['name'], '/');
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -261,8 +261,6 @@ class PDFObject
 
         self::$recursionStack[] = $this->getUniqueId();
 
-        $parts = [];
-
         foreach ($sections as $section) {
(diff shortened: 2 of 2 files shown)


k00ni:
Wow another PR already. Thanks for investing the time. I will try to get back to you asap!

k00ni (commented):
@se-ti In the following [example](https://github.com/smalot/pdfparser/pull/595/files#diff-749ce186ec0973f9f43da3714976412f5a08cf27334e141bac6ef160ea6f334cR110) you replaced the part `begin of the string` (`^`) with `first appearance` (`\G`). 

```
/^\s*(?P<name>\/[A-Z0-9\._]+)(?P<value>.*)/si
/\G\s*(?P<name>\/[A-Z0-9\._]+)(?P<value>.*)/si
```

You altered the scope of the regex, because they can produce different results. As of now, it always checks directly at the beginning of the string and if nothing is to be found, the if-clause is dismissed.

My question is, is this kind of information you check with the regex always at the beginning of the string anyway? Can there multiple instances of the this kind of information in the string at question?

se-ti:
Sorry, i don't understand You. :(
`As of now, it always checks directly at the beginning of the string...`
It was a `/^\s*(?P<name>\/[A-Z0-9\._]+)(?P<value>.*)/si` 
It DID check (see ^) just very beginning of the string.

I have replaced it with \G, which is "first matching position in subject". And matching position is changed because of the $offset parameter.

`The \G assertion is true only when the current matching position is at the start point of the match, as specified by the offset argument of preg_match. It differs from \A when the value of offset is non-zero.`
https://www.php.net/manual/en/regexp.reference.escape.php

Multiple instances are not the issue, because of regexps greedyness -- see .* in the end.
And this code definitly works not with the whole document, but with it's small part, substr-ed at higher level.

k00ni (approved):
I just wanna make sure who don't change parser behavior by accident.

My question about `\G` was due to the fact that it behaves differently if `$offset` is set. Quote:

> The \G assertion is true only when the current matching position is at the start point of the match, as specified by the offset argument of [preg_match()](https://www.php.net/manual/en/function.preg-match.php). **It differs from \A when the value of offset is non-zero.**

And about `\A` it says:

> start of subject (independent of multiline mode)

After reading the docs again I am for merging this.

---

_Offtopic:_ We have a few "performance" tests in https://github.com/smalot/pdfparser/tree/master/tests/Performance. Their purpose is to check time or RAM usage when processing some more-or-less heavy data. You posted that you did some performance tests yourself. Do you think we should integrate them?

se-ti:
My performance tests were pretty naive: just launching code from above with different memory limits, testing whether it would fail or not.

se-ti:
Looked at existing performance tests. They are about time performance.
I'd think, but right now i do not see reliable way to test memory performance.

One of reasons are these results "X Mb for the first launch, X/2.5 for the best subsequent one". 
I hope, that was OpCache warm up, but won't bet it. :(

Issue #599, opened 2023-05-02
Incorrect handling of "<~" in ASCII85 encoding


Seb35:
- PHP Version: 7.4
 - PDFParser Version: 2.4.0 (43a1c145b91d83f6b9c6fb4e5c2e6b274d2a0a17)

### Description:

I cannot share the PDF since it is one of my bank statement, but for one specific page the text was not extracted at all (the first page in a 2-page PDF).

After diving in the PDF and in PDFParser, I understood the [ASCII85 encoding](https://en.wikipedia.org/wiki/Ascii85#Adobe_version) has an optional `<~` header and an optional `~>` footer. The stream of this specific page was ASCII85Decode+FlateDecode, it had no `<~` header but the _**3 last characters were `<~>`**_ (since `<` is one of the ASCII85 character set).

### Solution:

In the ASCII85 decoder of PDFParser, there is a [search of the two characters `<~`](https://github.com/smalot/pdfparser/blob/9094d7777aeaa7d8a0a845b352f7dfe4a8baced3/src/Smalot/PdfParser/RawData/FilterHelper.php#L153)  and, if found, the two first characters of the string are removed. Here these two characters are found (at the end because of chance), but the two first characters must not be removed since there is no header.

Replacing the condition `false !== strpos($data, '<~')` by `0 === strpos($data, '<~')` fixes the issue.

Seb35:
PS: **_many many thanks_** for this library, I was able to extract data from the PDFs that my bank provides me.

k00ni:
Thank you for the feedback, would you mind opening a pull request with the solution you proposed?

PR #600, opened 2023-05-02
Fix ASCII85 decoding in a specific case

Seb35:
When the start or the end of the string is `<~>`. In this case we must be sure of the exact position of this substring: e.g. when there is no `<~` header but `<~>` at the end, we must not remove the first two characters even if `<~` is found somewhere.

Issue: #599

### DIFF
--- a/src/Smalot/PdfParser/RawData/FilterHelper.php
+++ b/src/Smalot/PdfParser/RawData/FilterHelper.php
@@ -150,13 +150,13 @@ class FilterHelper
         // all white-space characters shall be ignored
         $data = preg_replace('/[\s]/', '', $data);
         // remove start sequence 2-character sequence <~ (3Ch)(7Eh)
-        if (false !== strpos($data, '<~')) {
+        if (0 === strpos($data, '<~')) {
             // remove EOD and extra data (if any)
             $data = substr($data, 2);
         }
         // check for EOD: 2-character sequence ~> (7Eh)(3Eh)
         $eod = strpos($data, '~>');
-        if (false !== $eod) {
+        if (\strlen($data) - 2 === $eod) {
             // remove EOD and extra data (if any)
             $data = substr($data, 0, $eod);
         }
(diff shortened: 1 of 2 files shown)


Seb35:
It will be complicated to create a PDF whose the gzencode then ASCII85-encoding has the problematic ending, but I can add a test similarly to the [existing ones](https://github.com/smalot/pdfparser/blob/master/tests/PHPUnit/Integration/RawData/FilterHelperTest.php#L54) whose the end of the decoded string is (hex) `215B 33C0` (=`![3`+0xC0) which is ASCII85-encoded as `+^6b<~>` – I found this in my original PDF.

k00ni:
@Seb35 thank you for this pull request.

 > but I can add a test

That is sufficient, if it demonstrates the change. Providing a whole PDF is not required, most of the time it is only a bad trade in comparison to a tailored unit test.

k00ni:
@Seb35 what is the status here?

Seb35:
Sorry, I didn’t take the time to work on it, I do it soon (i.e. add a test).

Seb35:
I added 3 tests:
* the optional initial sequence `<~`
* the optional final sequence `~>`
* the specific end sequence seemingly mixing initial and end sequences but legal since `<` and `>` are in the target characters (althought not `~` only authorised in the initial and final sequences)

k00ni:
Good work!

Coding style rules changed with latest release of PHP-CS-Fixer. It should be sufficient to jut run it once to solve these errors (e.g. by using `make run-php-cs-fixer`).

> PS: I was a bit surprised not to find PHPUnit in the root composer.json, but found the [dev doc](https://github.com/smalot/pdfparser/blob/master/doc/Developer.md) which is very clear, and I have to say I like this way of separating dev tools from ordinary dependencies.

Yeah, it was a pain in the ass to handle PHP version differences as well as drag PHPUnit along. Separate dev tools from library did us a good service ever since.

Seb35:
> Coding style rules changed with latest release of PHP-CS-Fixer. It should be sufficient to jut run it once to solve those errors (e.g. by using `make run-php-cs-fixer`).
I submitted https://github.com/smalot/pdfparser/pull/602 for that and will rebase this PR just after, to separate the two tasks.

Seb35:
Sorry for not answering before on the other PR, thanks for finalising it. I’ve done a simple rebase (no conflicts) on this one.

k00ni (approved):

PR #601, opened 2023-05-19
Update Font.php: Remove unnecessary blank replacement

radieter:
Remove unnecessery blank replacement; caused incorrect pdf decoding

### DIFF
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -434,8 +434,8 @@ class Font extends PDFObject
 
             // replace escaped chars
             $text = str_replace(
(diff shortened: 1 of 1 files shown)


k00ni:
You wrote in #597:

> I just submitted a very similar pull request. I had issues with lost characters. Same like you I cannot provide a demo by now for the same reasony. Using xdebug I discovered that the str_replace removed a "\" too much...  also: why is this replacement hard-coded?

What program generated the PDF? Is the PDF valid? 

Please also describe what you mean with lost characters. What happened after you deployed this fix?

I get your point, but just because something was removed to often doesn't mean the library is faulty in the first place. Maybe there is a note or other hint (in the PDF specification) about that?

k00ni:
Please be aware, I will close this PR if there is no feedback given.

PR #606, opened 2023-06-22
Read XMP Metadata and add it to data returned by getDetails()

GreyWyvern:
Try to get XMP Metadata from the PDF document, and if successful, prefer it over any decoded header "details" we may find.

As an example, here's a PDF where the FlateDecode'd header data results in Title and Creator "details" values that contain invalid UTF-8 characters: https://evertz.com/resources/press/Studer-Vista-X-Mixbus-At-Eurovision.pdf

The PDF however also contains an XMP Metadata block where the Title and Creator values are encoded properly. If it exists, we should prefer that over gzuncompress()'ed data which may potentially have invalid UTF-8 characters.


src/Smalot/PdfParser/Parser.php
```diff
@@ -214,6 +214,12 @@ protected function parseObject(string $id, array $structure, ?Document $document
                         // It is not necessary to store this content.
 
                         return;
+
+                    } else if ($header->get('Type')->equals('Metadata')) {
+
+                        // Attempt to parse XMP XML Metadata
+                        $document->getXMPMetadata($content);
```
k00ni, line 221: Just curios: can we be sure that `$content` is valid XML at this point?

src/Smalot/PdfParser/Document.php
```diff
@@ -144,9 +149,119 @@ protected function buildDetails()
...
 
+        $details = array_merge($details, $this->metadata);
+
         $this->details = $details;
     }
 
+    /**
+     * Get XMP Metadata
+     */
+    public function getXMPMetadata(string $content)
```
k00ni, line 160: A get function always returns something in my opinion, but this function extracts meta data. Therefore I propose renaming it to `extractXMPMetadata` instead.

k00ni (changes requested):
@GreyWyvern Thank you for this pull request!

I made some remarks. Also, please add one or two tests which demonstrate that your code is working properly.

k00ni:
There should also be a remark in our documentation (maybe in https://github.com/smalot/pdfparser/blob/master/doc/Usage.md?) about these new meta data and how to get them.

src/Smalot/PdfParser/Parser.php
GreyWyvern: No, at this point it's just sending the content of the whole block as a string. The check for valid XML is in getXMPMetadata at the xml_parse_into_struct() statement. I supoose there could be a few more checks to see if this is a valid XML block of *type* XMP. It's just assuming any valid XML is an XMP block.

GreyWyvern:
> @GreyWyvern Thank you for this pull request!
> 
> I made some remarks. Also, please add one or two tests which demonstrate that your code is working properly. The coding style issues can be fixed by running PHP-CS-Fixer locally (https://github.com/smalot/pdfparser/blob/master/doc/Developer.md#php-cs-fixer).

I'm sorry, I'm having a rough time running the tests with PHPUnit as I'm developing on Windows. I keep getting the error:

PHP Fatal error:  Uncaught Error: Class "PHPUnitTests\TestCase" not found in  C:\Users\...\Documents\GitHub\pdfparser\tests\PHPUnit\Integration\ConfigTest.php:41

I've made a small example PDF that demonstrates the XMP-reading vs. no XMP-reading difference very well, so it should not be too difficult.

GreyWyvern:
> There should also be a remark in our documentation (maybe in https://github.com/smalot/pdfparser/blob/master/doc/Usage.md#extract-metadata?) about these new meta data and how to get them.

This metadata is gotten in the same way as before, by using getDetails(). If any XMP data was found, it will over-write (via array_merge) any header values with the same array keys obtained by regular extraction or gzuncompress().

k00ni:
> I'm sorry, I'm having a rough time running the tests with PHPUnit as I'm developing on Windows. I keep getting the error:
> 
> PHP Fatal error: Uncaught Error: Class "PHPUnitTests\TestCase" not found in C:\Users...\Documents\GitHub\pdfparser\tests\PHPUnit\Integration\ConfigTest.php:41

It looks like your auto loading isn't working. You ran `composer update`? Please check https://github.com/smalot/pdfparser/blob/master/.github/workflows/continuous-integration.yml#L203 (and below). These are our Windows related tests.

GreyWyvern:
I have added the Unit Test. One thing I noticed is that while the XMP Metadata is encoded in UTF-8, the xref data is encoded in CP1252 which means the Registered Sign (U+00AE) glyph from the XMP_Metadata.pdf **is** decodable with mb_convert_encoding($str, 'UTF-8', 'CP1252') or displaying the content on a page with charset CP1252, however high-level UTF-8 glyphs are **not** encodable in CP1252 and cannot be decoded properly.

Might be something to note for xref decoding problems you might have in the Issues list?

k00ni:
> Might be something to note for xref decoding problems you might have in the Issues list?

Good catch! Could you make a small reference in (some of) these issues?

GreyWyvern:
> Good catch! Could you make a small reference in (some of) these issues?

Sure. Where would you like me to do that?

I've done some more studying about this and now I'm not so precisely sure what character encoding Adobe is using to store encoded Properties values, only that the characters it destroys are all in the 32 character range where ISO-8859-1 differs from CP1252. CP1252 actually has a code point for Right Single Quotation mark at 0x92, but Adobe saves it as 0x90 which is an empty code point in CP1252. I'm not sure if this is a bug, or if one of Adobe's special encodings map to this. No other character set seems to: https://www.fileformat.info/info/unicode/char/2019/charset_support.htm

FWIW, WinAnsiEncoding seems to map to CP1252.

Luckily the XMP Metadata is **always** encoded in UTF-8 as per standard, so it can be relied upon.

> The pull request looks good now, after remaining coding style issues (importing a class which is not used) are fixed, we can merge it.

Oh, yes, haha. I noticed PHP-CS-Fixer flagged that, but my edits didn't touch Encoding.php so I didn't commit it.

GreyWyvern:
Hey hey! Outside the scope of this PR, but I believe I found it: https://github.com/maxwell-bland/pdf-latin-text-encodings

Apparently PdfParser is missing an encoding **PDFDocEncoding**, which the github readme file above describes it as: 
> Encoding for text strings in a PDF document outside the document's content streams.

Adobe has a class for it, so it's a real thing: https://developer.adobe.com/experience-manager/reference-materials/cloud-service/javadoc/com/adobe/internal/pdftoolkit/core/util/PDFDocEncoding.html

The github above has a JSON file with all the characters from the set: https://github.com/maxwell-bland/pdf-latin-text-encodings/blob/main/pdf-encoding.json Maybe we can use this to create a new Encoding file specifically for Document Properties.

k00ni (approved):
I took the liberty to fix remaining coding style and PHPStan issues.

> Sure. Where would you like me to do that?

I assumed that you had particular issues in mind. We have a lot of encoding and white space related issues, so every bit of help there is appreciated.

Lets move the discussion about PDFDocEncoding to #609, so we can keep this pull request clean.

Thank you for your work! I will leave this PR open for a few days. If there are no objections/feedback I will merge it.

src/Smalot/PdfParser/Document.php
```diff
@@ -144,9 +149,116 @@ protected function buildDetails()
...
+                    case 'DC:TITLE':
+                        $detail = ('open' == $val['type']) ? 'Title' : '';
+                        break;
+
+                    case 'DC:SUBJECT':
+                        $detail = ('open' == $val['type']) ? 'Subject' : '';
+                        break;
+
+                    case 'RDF:LI':
+                        if ('' !== $detail && 'complete' == $val['type'] && isset($val['value'])) {
```
k00ni, line 187: I replaced `$detail` with `'' !== $detail` in this if-clause. Now its more clear what this part means. The former one was ambiguous, because PHP uses magic sometimes to decide what values related to `true` when using `if($var)`.

GreyWyvern:
One caveat that I've just discovered, I don't think it's a blocker: When saving a PDF with properties, Adobe uses the contents of the **Subject** field to populate the <dc:description> XMP metadata element, even though there is a <dc:subject> element in the XMP specifications. This means that $details['Subject'] will retain the original decoded Subject, while the XMP 'Subject' will be in $details['Description'].

I think in the future I could refine this so that XMP values are represented in their own namespace within the getDetails() response. I assumed it was a 1:1 relationship, it is *almost* but not quite.

GreyWyvern:
@k00ni, if possible, I would like to submit some new code for this PR before you merge it. More and more I feel like overwriting the regular decoded getDetails() values is the wrong idea, and XMP data should be distinguished from it. In some cases, XMP data can even conflict with itself! Such as when there are pdf:creator and dc:creator elements in the same XMP block. You can't just assign them all to $details['Creator'].

I propose adding XMP data to the getDetails array using the lowercased and namespaced element name. So it would look like this:

```php
Array
(
    [Creator] => Adobe Acrobat
    [Producer] => Adobe Acrobat
    [CreatedOn] => 2022-01-28T16:36:11+00:00
    [Pages] => 35
    [dc:creator] => My Name
    [pdf:producer] => Adobe Acrobat
    [dc:title] => My Document Title
    ...
)
```

Let me know and I'll push the commit I've created.

k00ni:
Sure, take the time you need to polish the pull request. No rush.

src/Smalot/PdfParser/Document.php
```diff
@@ -163,100 +163,77 @@ public function extractXMPMetadata(string $content): void
...
-                        if ('complete' == $val['type'] && isset($val['value'])) {
-                            $this->metadata['InstanceUUID'] = $val['value'];
-                        }
-                        break;
                 }
             }
+
+            // According to the XMP specifications: 'Conflict resolution
+            // for separate packets that describe the same resource is
+            // beyond the scope of this document.' - Section 6.1
```
k00ni: Please insert URL to the XMP specificiation.

src/Smalot/PdfParser/Document.php
```diff
@@ -144,9 +149,97 @@ protected function buildDetails()
...
+                            }
+                        }
+                        break;
+
+                    case 'close':
+                        // If the value of this property is a single-
+                        // element array where the element is of type
+                        // string, use the value of the first list item
+                        // as the value for this property
+                        if (is_array($metadata) && isset($metadata[0]) && count($metadata) == 1 && is_string($metadata[0])) {
```
k00ni, line 217:
```suggestion
                        if (is_array($metadata) && isset($metadata[0]) && 1 == count($metadata) && is_string($metadata[0])) {
```

src/Smalot/PdfParser/Document.php
```diff
@@ -144,9 +149,97 @@ protected function buildDetails()
...
+            $metadata = [];
+            $stack = [];
+            foreach ($values as $val) {
+
+                // Standardize to lowercase
+                $val['tag'] = strtolower($val['tag']);
+
+                // Ignore structural x: and rdf: XML elements
+                if (strpos($val['tag'], 'x:') === 0) continue;
+                if (strpos($val['tag'], 'rdf:') === 0 && 'rdf:li' != $val['tag']) continue;
```
k00ni, line 176:
```suggestion
                if (0 === strpos($val['tag'], 'x:')) continue;
                if (0 === strpos($val['tag'], 'rdf:') && 'rdf:li' != $val['tag']) continue;
```

src/Smalot/PdfParser/Document.php
```diff
@@ -144,9 +149,97 @@ protected function buildDetails()
...
 
+    /**
+     * Extract XMP Metadata
+     */
+    public function extractXMPMetadata(string $content): void
+    {
+        $xml = xml_parser_create();
+        xml_parser_set_option($xml, \XML_OPTION_SKIP_WHITE, 1);
+
+        if (xml_parse_into_struct($xml, $content, $values, $index)) {
```
k00ni, line 165:
```suggestion
        if (1 === xml_parse_into_struct($xml, $content, $values, $index)) {
```

src/Smalot/PdfParser/Document.php
```diff
@@ -144,9 +149,97 @@ protected function buildDetails()
...
+  
+                switch ($val['type']) {
+                    case 'open':
+                        // Create an array of list items
+                        if ('rdf:li' == $val['tag']) {
+                            $metadata[] = [];
+
+                            // Move up one level in the stack
+                            $stack[count($stack)] = &$metadata;
+                            $metadata = &$metadata[count($metadata) - 1];
```
k00ni, line 186: Can you explain this and the next lines when using `&$metadata` with one or two sentences? 

I'd rather use a counter instead of `count` above, but its just a minor thing. The usage of `&$metadata` is a bit strange.

GreyWyvern: Sure. The output of xml_parse_into_struct is a single dimensional array, and the $stack is a last-on, first-off array of pointers to positions in $metadata, while iterating through it, that potentially turn the results into a more intuitive multi-dimensional array. When an "open" XML tag is encountered, we save the current $metadata context in the $stack, then create a child array of $metadata and make that the *current* $metadata context. When a "close" XML tag is encountered, the operations are reversed: the most recently added $metadata context from $stack (IOW, the *parent* of the current element) is set as the current $metadata context.

k00ni:
I've added your explanation to Document.php and fixed remaining coding style issues. Declined PHP-CS-Fixer to use rule `modernize_strpos`, because it replaces strpos with str_starts_with, which is only available since PHP 8.0 ([ref](https://cs.symfony.com/doc/rules/alias/modernize_strpos.html)).

GreyWyvern:
Looks good, thanks! :)

After this, I'll have another PR that decodes the regular metadata with PDFDocEncoding, which is why I removed some tests related to that from DocumentTest.php.

k00ni:
Can I merge it now?

GreyWyvern:
> Can I merge it now?

Sure, go ahead!

PR #610, opened 2023-07-04
PHP 8.3 support

k00ni:
PHP 8.3 is still in alpha, so we don't have to hurry here. I wanted to have it integrated already so we can observe changes and how PDFParser behaves.


k00ni:
No additional changes were needed to support PHP 8.3.

Issue #366, opened 2020-11-03
Weird characters and missing text in multiline metadata text


gnojus:
Hello.
I have an issue with the way this library extracts metadata from a PDF file. It comes out corrupted, and I'm not sure why.
It looks to me that for some reason `\` are prefixed before newlines, but I may be wrong.
Code used:
```php
$parser = new \Smalot\PdfParser\Parser();
$pdf    = $parser->parseFile('./file.pdf');
$details  = $pdf->getDetails();
echo $details['Subject'];
```
I attach the pdf's tested with metadata descriptions used, inserted with _Adobe Acrobat_.
[pdf_1.pdf](https://github.com/smalot/pdfparser/files/5483330/pdf_1.pdf)
[pdf_2.pdf](https://github.com/smalot/pdfparser/files/5483331/pdf_2.pdf)
[pdf_3.pdf](https://github.com/smalot/pdfparser/files/5483332/pdf_3.pdf)
[subject_in_1.txt](https://github.com/smalot/pdfparser/files/5483333/subject_in_1.txt)
[subject_in_2.txt](https://github.com/smalot/pdfparser/files/5483334/subject_in_2.txt)
[subject_in_3.txt](https://github.com/smalot/pdfparser/files/5483335/subject_in_3.txt)
[subject_out_1.txt](https://github.com/smalot/pdfparser/files/5483339/subject_out_1.txt)
[subject_out_2.txt](https://github.com/smalot/pdfparser/files/5483341/subject_out_2.txt)
[subject_out_3.txt](https://github.com/smalot/pdfparser/files/5483342/subject_out_3.txt)

Also I actually came here because of another library https://github.com/pauln/tcpdi. However, that one didn't seem very active and both of these libraries seemed to have problems with metadata. So there comes my other question - what is the relationship between `tcpdi_parser` and`tcpdf_parser` could `tcpdi_parser` be replaced by `tcpdf_parser`, this way having more update version?

k00ni:
> Also I actually came here because of another library https://github.com/pauln/tcpdi. However, that one didn't seem very active and both of these libraries seemed to have problems with metadata. So there comes my other question - what is the relationship between tcpdi_parser andtcpdf_parser could tcpdi_parser be replaced by tcpdf_parser, this way having more update version?

[...]

Issue #609, opened 2023-06-27
Missing PDFDocEncoding


GreyWyvern:
The Adobe PDF Reference defines a special encoding which is an extension of Latin1 such that:

> Informational or content strings can be represented in Unicode. These strings include text annotations, bookmark names, article names, document information, date strings, etc. In PDF 1.1 these strings are stored in PDFDocEncoding, which is a superset of ISOLatin1.
PDF Reference 1.2 - https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.2.pdf

See also: https://ia801001.us.archive.org/1/items/pdf1.7/pdf_reference_1-7.pdf

As this includes "document information" such as titles, authors and other details, PdfParser should use PDFDocEncoding to translate these strings.

Here is a proposed 'PDFDocEncoding.php' file I quickly mocked up but haven't tested yet. You can give it a shot; I will also see if I can create a branch where this works and submit a PR.

```
<?php

/**
 * @file
 *          This file is part of the PdfParser library.
 *
 * @author  Sébastien MALOT <sebastien@malot.fr>
 *
 * @date    2017-01-03
 *
 * @license LGPLv3
 *
 * @url     <https://github.com/smalot/pdfparser>
 *
 *  PdfParser is a pdf library written in PHP, extraction oriented.
 *  Copyright (C) 2017 - Sébastien MALOT <sebastien@malot.fr>
 *
 *  This program is free software: you can redistribute it and/or modify
 *  it under the terms of the GNU Lesser General Public License as published by
 *  the Free Software Foundation, either version 3 of the License, or
 *  (at your option) any later version.
 *
 *  This program is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *  GNU Lesser General Public License for more details.
 *
 *  You should have received a copy of the GNU Lesser General Public License
 *  along with this program.
 *  If not, see <http://www.pdfparser.org/sites/default/LICENSE.txt>.
 */

// Source : https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.2.pdf
// Source : https://ia801001.us.archive.org/1/items/pdf1.7/pdf_reference_1-7.pdf

namespace Smalot\PdfParser\Encoding;

/**
 * Class PDFDocEncoding
 */
[...]

Issue #79, opened 2015-08-26
Problem with multiple authors


sinnbeck:
If a document has multiple authors then only the first one i retrieved.
The author field is setup as ; seperated (with a space as well).
Example
Jane Doe; John Doe

This will only retrieve Jane Doe.

I have not been able to find the error myself.

René Sinnbeck

Resin01:
I have done some research and it would seem that PDF Parser checks the /Author() tag in the file, where as my files uses Dublin Core so store the metadata.
http://www.pdfa.org/2011/08/pdfa-metadata-xmp-rdf-dublin-core/

Here is an example.
    <dc:creator>
    <rdf:Seq>
    <rdf:li>Chad Cook</rdf:li>
    <rdf:li>Kenneth Learman</rdf:li>
    <rdf:li>Chris Showalter</rdf:li>
    <rdf:li>Vincent Kabbaz</rdf:li>
    <rdf:li>Bryan O'Halloran</rdf:li>
    </rdf:Seq>
    </dc:creator>

Is there some way to get the DC metadata?

k00ni:
@sinnbeck I know this issue is super old, but can you please test if #611 fixes your problem? That would be helpful.

sinnbeck:
I sadly dont work with the system from back then anymore, so I have no way of testing sadly. But I will close this thread as I am sure it is solved :)

PR #611, opened 2023-07-04
Enable PDFDocEncoding support for metadata

GreyWyvern:
Regular PDF metadata (outside of XMP), depending on the characters it includes, can be encoded in UTF-8 escaped (or binary) bytes, or using a proprietary Adobe encoding PDFDocEncoding which is similar to, but not exactly like CP1252 (WinAnsiEncoding).

For more information on the PDFDocEncoding character set, see: https://ia801001.us.archive.org/1/items/pdf1.7/pdf_reference_1-7.pdf

Another issue is that regardless of the storage encoding used, Adobe Acrobat will attempt to add a slash-linefeed (\r) to metadata text to avoid long line-lengths (~127 bytes) in the saved PDF data. Unfortunately, the method to do this does not seem binary-safe, resulting in UTF-8 saved bytes that are damaged and must be repaired.

This commit enables decoding PDF metadata using PDFDocEncoding, and also repairs added line-feeds in both PDFDocEncoding and UTF-8.

It also adds a sample file "Issue609.pdf" containing both UTF-8 and PDFDocEncoding encoded metadata fields for testing. The name of the file references PDFParser issue #609

_**Edit**: My apologies for the failing tests. I don't think I'm running PHPStan properly since I'm using the phar instead of composer. My env doesn't have direct/full access to the internet, so I can't really use composer._


GreyWyvern:
I just did a quick, glance-thru, but I can confirm that this change resolves issues #366 and #79

src/Smalot/PdfParser/Document.php
```diff
@@ -149,6 +151,43 @@ protected function buildDetails()
...
+
+                    $value = str_replace("\x5c\x0d", '', $value);
+
+                    while (preg_match("/\x5c\x5c\xe0([\xb4-\xb8])(.)/", $value, $match)) {
+                        $diff = (\ord($match[1]) - 182) * 64;
+                        $newbyte = PDFDocEncoding::convertPDFDoc2UTF8(\chr(\ord($match[2]) + $diff));
+                        $value = preg_replace("/\x5c\x5c\xe0".$match[1].$match[2]."/", $newbyte, $value);
+                    }
+
+                    while (preg_match("/(.)\x9c\xe0([\xb3-\xb7])/", $value, $match)) {
```
k00ni, line 172: Please add a short 1-line comment for each while-loop in the code about what they do.

src/Smalot/PdfParser/Encoding/PDFDocEncoding.php
```diff
@@ -0,0 +1,184 @@
...
+
+// Source : https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.2.pdf
+// Source : https://ia801001.us.archive.org/1/items/pdf1.7/pdf_reference_1-7.pdf
+
+namespace Smalot\PdfParser\Encoding;
+
+/**
+ * Class PDFDocEncoding
+ */
+abstract class PDFDocEncoding extends AbstractEncoding
```
k00ni, line 40: I dont really like classes which have only 1 function and are also abstract :laughing: Anyway, I suggest we provide access to the long array separately, because it might be helpful to check array content itself sometimes. Also, a method to change its content would be helpful. What do you think?

src/Smalot/PdfParser/Encoding/PDFDocEncoding.php
```diff
@@ -0,0 +1,184 @@
...
+{
+    public static function convertPDFDoc2UTF8(string $content): string
+    {
+        return strtr($content, array(
```
k00ni, line 44:
```suggestion
        return strtr($content, [
```

src/Smalot/PdfParser/Encoding/PDFDocEncoding.php
```diff
@@ -0,0 +1,184 @@
...
+            "\xf7" => "\u{00f7}", // divide
+            "\xf8" => "\u{00f8}", // oslash
+            "\xf9" => "\u{00f9}", // ugrave
+            "\xfa" => "\u{00fa}", // uacute
+            "\xfb" => "\u{00fb}", // ucircumflex
+            "\xfc" => "\u{00fc}", // udieresis
+            "\xfd" => "\u{00fd}", // yacute
+            "\xfe" => "\u{00fe}", // thorn
+            "\xff" => "\u{00ff}"  // ydieresis
+        ));
```
k00ni, line 182:
```suggestion
        ]);
```

k00ni (changes requested):
@GreyWyvern Thank you again for taking the time improving this library.

I have a few remarks, but nothing big.

src/Smalot/PdfParser/Encoding/PDFDocEncoding.php
GreyWyvern: Haha, yeah. I suppose I don't really need to do that. :)

I can add a function that returns the conversion array, but I'm not sure how PdfParser would make use of an additional method to modify it. PDFDocEncoding is a standard encoding documented by Adobe and pretty much set in stone. How would PdfParser benefit from being able to modify it on the fly?

k00ni: > How would PdfParser benefit from being able to modify it on the fly?

The main reason would be flexibility. PDFParser has many issues and there are also many invalid PDFs out there. As a user it might be helpful, if I can tweak around these obstacles by using a custom configuration/setup for a given PDF. I wouldn't go as far as our [CustomConfig](https://github.com/smalot/pdfparser/blob/master/doc/CustomConfig.md) approach here, but it points in the same direction. I am still torn.

Maybe @j0k3r can add something to this as well?

src/Smalot/PdfParser/Document.php
k00ni: Helpful comments now, thanks.

src/Smalot/PdfParser/Encoding/PDFDocEncoding.php
GreyWyvern: Personally speaking, I believe dynamically modifying PDFDocEncoding is not something that is needed, and the main reason is that it only applies to document metadata and other non-document streams. I looked through the entire Issues list, and there are many unresolved encoding problems, but they all have to do with the document stream, and PDFDocEncoding doesn't apply there.

OTOH, adding a function that lets you dynamically modify the conversion table is pretty trivial to implement, so I'll leave it up to your consensus.

k00ni (approved):
Looks good to me.

PR #614, opened 2023-07-12
Add font fallback + Support for font IDs containing hyphens

### LINKED ISSUE #145: Cannot extract text from PDF (internal font naming issue)
Hi. 
While trying to extract the text from PDF
[test.pdf](https://github.com/smalot/pdfparser/files/867641/test.pdf)
 I was only able to extract "\n"s.
My investigation results:

While trying to parse "commandsText" (PdfParser::getCommandsText), there is an intresting issue:
Given a line "/f-0-0 1 Tf" it tries to parse it with two regular expressions assuming that the font "id" will not contain "-" symbols while it actually contains.

As a result, it stops parsing without any error message and the output is incorrect.

My fix is (one more regexp):
```php
elseif (preg_match(
                        '/^\/([A-Z0-9\._,\+\-]+\s+[0-9]+)\s+([A-Z]+)\s*/si',
                        substr($text_part, $offset),
                        $matches
                    )
                    ) {
                        $operator = $matches[2];
                        $command  = $matches[1];
                        $offset += strlen($matches[0]);
                    }
```

Pull request coming soon. Or maybe some remarks?

### LINKED ISSUE #586: text encoding breaks in the middle of the line
- PHP Version: 7.3.33
 - PDFParser Version: 2.3.0 

### Description:
In the middle of the last line text encoding changes to a wierd one with squares instead of characters.
Acrobat Reader and Chrome pdf-viewer shows this file correct;

### PDF input

[page91.pdf](https://github.com/smalot/pdfparser/files/11008055/page91.pdf)

actual: "Тем  не  менее,  стаявший  ледник	 \hdjm]� gmgZlZdZ�� km^y� ih� \k_fm�� k^_eZe� i_j_\Ze� keh`g__� b�"
expected: "Тем не менее, стаявший ледник вокруг нунатака, судя по всему, сделал перевал сложнее и"

### Code

```php
use Smalot\PdfParser\Parser;

$path = 'page91.pdf';

echo $pdf->getText();
```

GreyWyvern:
If part of a text stream is positioned after an incorrect font command, then undecoded, jumbled bytes will appear in the `getText()` output. This change adds code to check this output for UTF-8 control characters (\x00-\x1f + \x7f) and if they appear, loop through all available fonts to see if we can find one that decodes this output properly. If none is found, the original string is used. Resolves #586.

Also add support for font IDs containing hyphens. Previously these were ignored as invalid. Resolves #145.


k00ni (changes requested):
Its good to see you are still with us!

Just a few remarks/questions.

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -246,6 +246,36 @@ private function getDefaultFont(Page $page = null): Font
         return new Font($this->document, null, null, $this->config);
     }
 
+    private function getTJUsingFontFallback(Font $font, array $command, Page $page = null): string
```
k00ni, line 249: Third parameter seems invalid. If `$page` can be `null`, than it has to be `?Page $page = null`.

Also, how does array `$command` look like? Please have a look at https://phpstan.org/writing-php-code/phpdoc-types#general-arrays

If you are not sure, we can use the following for now:

 ```suggestion
    /*
     * @param array<mixed> $command
     */
    private function getTJUsingFontFallback(Font $font, array $command, ?Page $page = null): string
```

Comment (GreyWyvern) [reply src/Smalot/PdfParser/PDFObject.php:249]: $command looks like this. It's an array of parsed elements from a single line of stream content.

```
array(31) {
  [0]=>
  array(3) {
    ["t"]=>
    string(1) "<"
    ["o"]=>
    string(0) ""
    ["c"]=>
    string(4) "023E"
  }
  [1]=>
  array(3) {
    ["t"]=>
    string(1) "n"
    ["o"]=>
    string(0) ""
    ["c"]=>
    string(3) "1.7"
  }
  [2]=>
  array(3) {
    ["t"]=>
    string(1) "<"
    ["o"]=>
    string(0) ""
    ["c"]=>
    string(4) "025F"
  }
  [3]=>
  array(3) {
    ["t"]=>
    string(1) "n"
    ["o"]=>
    string(0) ""
    ["c"]=>
    string(3) "2.9"
  }
  ...
```

So how would you write that? Maybe `list<array>`?

Comment (k00ni) [reply src/Smalot/PdfParser/PDFObject.php:249]: Based on the given data it is `array<int,array<string,string>>` in my opinion. If values can be `null`, it must be `array<int,array<string,string|null>>` instead.

Comment (GreyWyvern) [reply src/Smalot/PdfParser/PDFObject.php:249]: I think it might be `array<int,array<string,mixed>>`. If you look at `getCommandsText()` it initializes the three values that end up in "t", "o" and "c" as `string`, `string`, `bool` respectively.

```php
            $operator = '';
            $type = '';
            $command = false;
```

If I change it to `$command = '';` the output is broken so the `bool` is required here somehow.

k00ni: I am not familiar with the content of those variables. Its just a minor thing, but `array<int,array<string,string|bool>>` would be most precise. But either way, use one that fits the best in your opinion.

samples/FontIDHyphen.pdf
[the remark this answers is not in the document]
GreyWyvern: I'm afraid I've not been successful installing WeasyPrint 0.31. I also tried generating a PDF with a dash in a font name and was successful, but for the bug to appear the text in the content stream needs to be encoded in a CIDMap which I couldn't get FPDF to do. It just puts the text as-is in the stream, so PdfParser doesn't need to know anything about the font to extract it.

So... should I just remove the PDF and the associated test?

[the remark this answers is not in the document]
k00ni: PDF file could be removed, if we are able to setup a test which uses mocks/stubs instead. Could you try it?

Maybe something like:

```php
/*
 *  setup $document and maybe parameter of getText in a way to trigger new functions/functionality 
 * and check its result
 */
$document = $this->createMock(Document::class);
// $document->method(...)->willReturn(...);

$pdfObject = new PDFObject($document);
$pdfObject->getText(/* .. */);
```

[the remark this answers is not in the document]
k00ni: In case you need a hand here, give me a hint.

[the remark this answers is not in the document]
GreyWyvern: I did change the test to just test for a correctly parsed Font command. If you'd like me to get more elaborate with mock documents, I could do that if you like. :)

k00ni:
Are you done here @GreyWyvern?

k00ni (approved):

GreyWyvern:
> Are you done here @GreyWyvern?

Yes, sorry. I was on vacation this week. :)

k00ni:
> > Are you done here @GreyWyvern?
> 
> Yes, sorry. I was on vacation this week. :)

All good, hope you had a good one.

Issue #591, opened 2023-03-27
Fix failing Scrutinizer CI tests


k00ni:
Since a few months, Scrutinizer CI keeps failing. Here are latest results: https://scrutinizer-ci.com/g/smalot/pdfparser/inspections

It seems to complain about the PHPUnit config file we use. I don't really have time to look into that further, but if anyone wants to bring the issue forward (in form of a pull request), I am happy to assist.

PR #615, opened 2023-07-16
Fix failing Scrutinizer CI tests

jzohrab:
Fix Scrutinizer 'Incorrect whitelist config' error.

Fixes https://github.com/smalot/pdfparser/issues/591.

Note: I tested this out on a branch in my fork, setting up scrutinizer to run on a test branch in my repo.  Initially the scrutinize run failed with the same error as you received in your runs on master.  Then I added the "filter" section to the php config as suggested by others in google/stackoverflow and re-ran it, and it passed in scrutinize:

image

Link to the issues: https://scrutinizer-ci.com/g/jzohrab/pdfparser/issues

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -18,4 +18,9 @@
       <directory>tests/PHPUnit</directory>
     </testsuite>
   </testsuites>
+  <filter>
+    <whitelist processUncoveredFilesFromWhitelist="true">
+      <directory suffix=".php">src</directory>
+    </whitelist>
+  </filter>
 </phpunit>


k00ni:
@jzohrab Thank you for the work!

It seems to be problematic in the Linux tests, https://github.com/smalot/pdfparser/actions/runs/5564829310/jobs/10177967216?pr=615#step:6:24

jzohrab:
Hi @k00ni , interesting, maybe that version of the software (whatever it is) uses a different schema, and the xml needs to be sed'ed for that configuration.

My recommendation would be to merge this PR, as it's a good step in the right direction, and then the "failing linux" thing is a different issue.  Thoughts?

jzohrab:
Interesting, I find it odd that phpunit is validating the config file with its own XSD and passing/failing the CI run based on that, given that all tests pass.  Ah well, that's code.  At least with my local setup, I get the same validation error:

```
$ xmllint --noout --schema dev-tools/vendor/phpunit/phpunit/phpunit.xsd phpunit.xml 
phpunit.xml:5: element include: Schemas validity error : Element 'include': This element is not expected.
phpunit.xml:21: element filter: Schemas validity error : Element 'filter': This element is not expected.
phpunit.xml fails to validate
```

While I haven't checked, it appears that the php xsd changes across different versions, so the pipeline will fail with this xsd violation for some configurations.

k00ni:
`filter`-element was available only until PHPUnit 8.5 ([doc](https://docs.phpunit.de/en/8.5/configuration.html#the-filter-element)). `<coverage>` is to be used since PHPUnit 9.0 ([doc](https://docs.phpunit.de/en/9.6/configuration.html#the-coverage-element)). It has a sub element `include` which should cover our case here: https://docs.phpunit.de/en/9.6/configuration.html#the-include-element

Interesting observation: Windows tests also complain about it, but finish successfully, like https://github.com/smalot/pdfparser/actions/runs/5564829310/jobs/10177968574?pr=615#step:6:17

> My recommendation would be to merge this PR, as it's a good step in the right direction, and then the "failing linux" thing is a different issue. Thoughts?

That would leave the library with a broken continuous pipeline. I don't see any reason why it is not fixable inside a pull request.

jzohrab:
Are the different versions of PHP used in the various matrix pipelines not also using different versions of phpunit?

I'll look further into this when I get time, thanks!

jzohrab:
Hi @k00ni - some follow up questions as to how to best handle this.  The composer.json has `"phpunit/phpunit": ">=7.5 <11.0"`, so anything in this range should work.  On my dev box (Mac), I have phpunit 10.2.5, which has a very different xsd (ref https://docs.phpunit.de/en/10.2/configuration.html#the-coverage-element) from phpunit 9.x.

On the github ci jobs, there are the following different phpunit versions, each with different possible configs:

```
lowest deps - PHPUnit 9.6.10
PHP 7.1 - PHPUnit 7.5.20
win 7.1 - PHPUnit 7.5.20 
PHP 7.4 - PHPUnit 9.6.10
PHP 8.1 - PHPUnit 10.2.5
win 8.1 - PHPUnit 10.2.5
```

So, a few questions:

* is it necessary to generate coverage for all different versions?  Would it suffice to limit coverage to just one version (say, the latest for win and non-win)?  I've always aimed to increase coverage, but perhaps chasing it across multiple versions doesn't have much real payoff; and just the pass-fail would suffice.
* if you do wish to have coverage for each possible version, then perhaps it would be necessary to commit different versions of the configs (phpunit-7.x.xml, phpunit-9.x.xml), and then either switch-case to cp that file to `phpunit.xml`, or add the `--configuration` flag to the run.

Cheers, jz

k00ni:
> * is it necessary to generate coverage for all different versions?  Would it suffice to limit coverage to just one version (say, the latest for win and non-win)?  I've always aimed to increase coverage, but perhaps chasing it across multiple versions doesn't have much real payoff; and just the pass-fail would suffice.

You are right, coverage for one PHP version is more than enough. In case it is not possible to achieve with one phpunit.xml for all cases, create a dedicated `phpunit.xml` solely used for Scrutinizer.

jzohrab:
Will do, cheers!

k00ni:
@jzohrab What is the status here? Do you need any help?

jzohrab:
Hi there, I have been travelling with limited hacking time, will get to it
when I can.  Hopefully in the next week or so.  Cheers!

k00ni:
Closing this for now due to (almost) no activity. Don't worry, just work on it when you have the time.

@GreyWyvern is working on many parts of the library currently. I want only active PR's for the moment so management stays practical. @jzohrab I am looking forward to future contributions from you.

PR #622, opened 2023-07-31
Allow for CRC23 checksum gzuncompress()

### LINKED ISSUE #592: parsed incorectly, wrong symbols
- PHP Version: 8.0
 - PDFParser Version: Latest: 2.5.5

EXPECTED:
Buyer Receipt
Receipt # 19509334
Receipt Date 10/26/2022
Received By AC/ASAP S...... (and so on as in pdf file)

ACTUAL:
%X\HU5HFHLSW 5HFHLSW  5HFHLSW'DWH  5HFHLYHG%\ $&$6$36 6ROG$W%UDQFK 'UHD

### Code

include "vendor/autoload.php";
$pdf = $parser->parseFile("inv.pdf");
$text = $pdf->getText();
echo $text;

GreyWyvern:
A zlib compressed stream may have a CRC32 checksum instead of Adler-32 which the PHP gzuncompress() function expects. Add a second zlib decompression attempt if the first one fails. See: https://www.php.net/manual/en/function.gzuncompress.php#79042

Partially resolves #592 by correctly decoding the document's default font stream. Includes `inv.pdf` from that issue by permission from the issue author.


k00ni (commented):
Good extension!

Your code looks a bit clunky, because of a try-catch inside a catch. Also, an `@` in front of a function may absorb errors which otherwise would have came to the developers attention. Some changes are just opinionated to improve readability. Here is my suggestion for the **whole** try-catch part (untested):

```php
// initialize string to return
try {
    $decoded = gzuncompress($data, $decodeMemoryLimit);
    if (false === $decoded) {
        // If gzuncompress() failed try again allowing for a CRC32 
        // checksum instead of Adler-32.
        // See: https://www.php.net/manual/en/function.gzuncompress.php#79042
        // Issue: https://github.com/smalot/pdfparser/issues/592
        $crc32 = tempnam('/tmp', 'gz_fix');
        if (false != $crc32) {
            file_put_contents($crc32, "\x1f\x8b\x08\x00\x00\x00\x00\x00".$data);
            $decoded = file_get_contents('compress.zlib://'.$crc32);
            unlink($crc32);
        }

        if (is_string($decoded) && 0 < strlen($decoded)) {
            // all good
        } else {
           // If the decoded string is empty, that means decoding failed.
           throw new \Exception('TODO: meaningful error message');
        }
} catch (\Exception $e) {
    throw $e;
} finally {
    // Restore old handler just in case it was customized outside of PDFParser.
    restore_error_handler();
}
```

What do you think?

GreyWyvern:
> Good extension!
> 
> Your code looks a bit clunky, because of a try-catch inside a catch. Also, an `@` in front of a function may absorb errors which otherwise would have came to the developers attention. Some changes are just opinionated to improve readability. Here is my suggestion for the **whole** try-catch part (untested):
> 
> ```
> [snip]
> ```
> 
> What do you think?

So I tried this and it turns out that the original `gzuncompress()` must fail and emit an E_WARNING in the case that the data is alternately coded. This means that because of the `set_error_handler()` above, that E_WARNING is upgraded to an E_ERROR and execution is knocked out of the single `try { ... } catch () { ... }` loop. The `if (false === $decoded) {` line is never even reached.

So either there **must** be a try/catch within a try/catch, or else the `set_error_handler()` should be removed, and the `gzuncompress()` should have it's E_WARNING suppressed with @. We can add comments in the code to explain why this is necessary.

```php
    protected function decodeFilterFlateDecode(string $data, int $decodeMemoryLimit): ?string
    {
        /* REMOVE THIS
         * gzuncompress may throw a not catchable E_WARNING in case of an error (like $data is empty)
         * the following set_error_handler changes an E_WARNING to an E_ERROR, which is catchable.
         */
        /* REMOVE THIS
        set_error_handler(function ($errNo, $errStr) {
            if (\E_WARNING === $errNo) {
                throw new \Exception($errStr);
            } else {
                // fallback to default php error handler
                return false;
            }
        }); */

        $decoded = null;

        // initialize string to return
        try {
            // Uncatchable E_WARNING for "data error" is @ suppressed
            // so execution may proceed with an alternate decompression
            // method.
            $decoded = @gzuncompress($data, $decodeMemoryLimit);
            if (false === $decoded) {
                // If gzuncompress() failed try again using the compress.zlib://
                // wrapper in case it might be something gzuncompress() doesn't
                // understand.
                // See: https://www.php.net/manual/en/function.gzuncompress.php#79042
                // Issue: https://github.com/smalot/pdfparser/issues/592
                $ztmp = tempnam('/tmp', 'gz_fix');
                if (false != $ztmp) {
                    file_put_contents($ztmp, "\x1f\x8b\x08\x00\x00\x00\x00\x00".$data);
                    $decoded = file_get_contents('compress.zlib://'.$ztmp);
                    unlink($ztmp);
                }

                if (is_string($decoded) && 0 < strlen($decoded)) {
                    // all good
                } else {
                   // If the decoded string is empty, that means decoding failed.
                   throw new \Exception('decodeFilterFlateDecode: invalid data');
                }
            }
        } catch (\Exception $e) {
            throw $e;
        } /* REMOVE THIS finally {
            // Restore old handler just in case it was customized outside of PDFParser.
            restore_error_handler();
        } */

        return $decoded;
    }
```

This change will require changing some tests in **FilterHelperTest.php** as it will no longer be returning several different exceptions. Just the one 'invalid data' one.

k00ni:
Good point. So the final code would look like the following (including some minor adaptions)?

```php
protected function decodeFilterFlateDecode(string $data, int $decodeMemoryLimit): ?string
{
    // Uncatchable E_WARNING for "data error" is @ suppressed
    // so execution may proceed with an alternate decompression
    // method.
    $decoded = @gzuncompress($data, $decodeMemoryLimit);

    if (false === $decoded) {
        // If gzuncompress() failed, try again using the compress.zlib://
        // wrapper to decode it in a file-based context.
        // See: https://www.php.net/manual/en/function.gzuncompress.php#79042
        // Issue: https://github.com/smalot/pdfparser/issues/592
        $ztmp = tempnam('/tmp', 'gz_fix');
        if (false != $ztmp) {
            file_put_contents($ztmp, "\x1f\x8b\x08\x00\x00\x00\x00\x00".$data);
            $decoded = file_get_contents('compress.zlib://'.$ztmp);
            unlink($ztmp);
        }
    }

    if (false === is_string($decoded) || 0 === strlen($decoded)) {
       // If the decoded string is empty, that means decoding failed.
       throw new \Exception('decodeFilterFlateDecode: invalid data');
    }

    return $decoded;
}
```

k00ni:
> Yes. We should even account for the $decodeMemoryLimit by limiting the size of the "file" file_get_contents() will load:

Sounds reasonable, please do.

k00ni:
Ready to merge?

Issue #617, opened 2023-07-21
getDetails() returning single element arrays instead of strings


GreyWyvern:
In the [sample PDF](https://github.com/smalot/pdfparser/blob/master/samples/bugs/Issue391.pdf) for #391 `/samples/bugs/Issue391.pdf` the results of `getDetails()` returns a set of values that are single element arrays instead of strings:

**Expected:**
```
array(6) {
  ["Title"]=>
  string(49) "Microsoft Word - GVTW70SPAH1R0_20101210_REV_A.doc"
  ["Author"]=>
  string(8) "aronzhao"
  ["Producer"]=>
  string(33) "Mac OS X 10.7.2 Quartz PDFContext"
  ["Creator"]=>
  string(26) "PScript5.dll Version 5.2.2"
  ["CreationDate"]=>
  string(25) "2011-12-28T21:12:01+00:00"
  ["ModDate"]=>
  string(25) "2011-12-28T21:12:01+00:00"
}
```

**Actual:**
```
array(6) {
  ["Title"]=>
  array(1) {
    [0]=>
    string(49) "Microsoft Word - GVTW70SPAH1R0_20101210_REV_A.doc"
  }
  ["Author"]=>
  array(1) {
    [0]=>
    string(8) "aronzhao"
  }
  ["Producer"]=>
  array(1) {
    [0]=>
    string(33) "Mac OS X 10.7.2 Quartz PDFContext"
  }
  ["Creator"]=>
  array(1) {
    [0]=>
    string(26) "PScript5.dll Version 5.2.2"
  }
  ["CreationDate"]=>
  array(1) {
    [0]=>
    string(25) "2011-12-28T21:12:01+00:00"
  }
  ["ModDate"]=>
  array(1) {
    [0]=>
    string(25) "2011-12-28T21:12:01+00:00"
  }
}
```

I knew arrays of XMP details values were possible, but not so for the regular details. I'll have to look at the PDF reference again, but they are probably allowed and PdfParser is parsing them properly.

My PR #606 automatically converts any XMP property that's a single element array into the value of that single element. Should we do that in this case for the regular details?

k00ni:
Please have a look first to confirm how it has to look. I'd rather prefer one way all the time, so in this case keep the arrays and don't convert to a single value, if the value only contains one element. This way it is easier for data consumers to check data, because they can rely on an array structure rather than checking for string, array ... . What do you think?

GreyWyvern:
[...]

PR #624, opened 2023-08-02
Account for empty arrays in XMP data

GreyWyvern:
The previous version of Document::extractXMPMetadata() did not account for empty arrays as values. When encountered, use the empty string as a value for this property.

There is no test, because in the future this behaviour might change based on the outcome of #617.

Sorry, I am trying to commit all my small changes before any big ones. :)

### DIFF
--- a/src/Smalot/PdfParser/Document.php
+++ b/src/Smalot/PdfParser/Document.php
@@ -271,7 +271,6 @@ class Document
                             // value for this property
                             if (1 == \count($metadata) && isset($metadata[0]) && \is_string($metadata[0])) {
                                 $metadata = $metadata[0];
-
                             } elseif (0 == \count($metadata)) {
                                 // if the value is an empty array, set
                                 // the value of this property to the empty
--- a/src/Smalot/PdfParser/Document.php
+++ b/src/Smalot/PdfParser/Document.php
@@ -272,10 +272,10 @@ class Document
                             if (1 == \count($metadata) && isset($metadata[0]) && \is_string($metadata[0])) {
                                 $metadata = $metadata[0];
 
-                                // else if the value is an empty array, set
+                            } elseif (0 == \count($metadata)) {
+                                // if the value is an empty array, set
                                 // the value of this property to the empty
                                 // string
-                            } elseif (0 == \count($metadata)) {
                                 $metadata = '';
                             }
                         }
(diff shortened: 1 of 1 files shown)


k00ni:
> Sorry, I am trying to commit all my small changes before any big ones. :)

A bit off topic: If you plan any bigger changes, we could have a short chat or audio call to coordinate. I really appreciate that you taking the time to improve PFDParser.

src/Smalot/PdfParser/Document.php
```diff
@@ -263,12 +263,21 @@ public function extractXMPMetadata(string $content): void
...
+                            // the value of the first list item as the
+                            // value for this property
+                            if (1 == \count($metadata) && isset($metadata[0]) && \is_string($metadata[0])) {
+                                $metadata = $metadata[0];
+
+                                // else if the value is an empty array, set
+                                // the value of this property to the empty
+                                // string
+                            } elseif (0 == \count($metadata)) {
+                                $metadata = '';
```
k00ni, line 279:
```suggestion
                            } elseif (0 == \count($metadata)) {
                                // if the value is an empty array, set
                                // the value of this property to the empty
                                // string
                                $metadata = '';
```

k00ni: The only option left is that `$metadata` has more than 1 element. Why is that not possible here?

GreyWyvern: > The only option left is that `$metadata` has more than 1 element. Why is that not possible here?

If `$metadata` has more than one element, it is not modified and remains an array. If it has zero or one elements (where the one element is of type `string`) then the value will be collapsed to a string.

For example, the following XMP XML:

```xml
<dc:Title>
  <rdf:Seq>
    <rdf:li></rdf:li>
  </rdf:Seq>
</dc:Title>
```

... would result in a value for `dc:title` of an empty string. However, for the following:

```xml
<dc:Title>
  <rdf:Seq>
  </rdf:Seq>
</dc:Title>
```

... the result was an empty array which is non-intuitive considering the behaviour of the first example.

k00ni:
> There is no test, because in the future this behaviour might change based on the outcome of https://github.com/smalot/pdfparser/issues/617.

The change is comprehensive. Also, I am OK for making an exception in this case, because you have proven to be reliable.

Issue #629, opened 2023-08-07
Fail to identify fonts (name and size)


mbideau-atreal:
- PHP Version: 7.4.5 (built: Apr 19 2020 08:43:53)
 - PDFParser Version: c974994 2023-08-03

### Description:
The fonts names and size are not found (always returns `fontid=-1` and `fontsize=1`).

### PDF input
See attached PDF sample
[20230803-160138-lettretype-arrete.pdf](https://github.com/smalot/pdfparser/files/12275119/20230803-160138-lettretype-arrete.pdf)

### Expected output & actual output
Expected output:
```
Fonts:
 - [3_0] Helvetica (Type1)
 - [4_0] Courier-Bold (Type1)
 - [5_0] Times-Roman (Type1)
 - [6_0] Times-Bold (Type1)
 - [7_0] AAAAAF+ArialMT (Type0)
 - [9_0] Helvetica-Oblique (Type1)
 - [14_0] AAAAAF+ArialMT (CIDFontType2)
Items:
- font: Courier-Bold (Type1) size: 40.000000  'IL - DOCUMENT DE TRAVAIL - DOCUMENT'
- font: Courier-Bold (Type1) size: 40.000000  'DE TRAVAIL - DOCUMENT DE TRAVAIL - DOCUMENT DE'
- font: Times-Bold (Type1) size: 10.000000  'Déclaration préalable'
- font: Times-Bold (Type1) size: 10.000000  'ARRETE'
- font: Times-Bold (Type1) size: 10.000000  '______________________________________________________________________'
- font: Times-Bold (Type1) size: 10.000000  'Dossier numéro         DP 020001 23 00001P0  déposé le 21/06/2023'
- font: Times-Bold (Type1) size: 10.000000  'par                              Monsieur a'
- font: Times-Bold (Type1) size: 10.000000  'Correspondant :        Monsieur a'
- font: Times-Bold (Type1) size: 10.000000  '                                     '
- font: Times-Bold (Type1) size: 10.000000  '                                      '
- font: Times-Bold (Type1) size: 10.000000  '                                       France'
- font: Times-Bold (Type1) size: 10.000000  'sur le terrain                    '
- font: Times-Bold (Type1) size: 10.000000  'arrondissement          '
- font: Times-Bold (Type1) size: 10.000000  '______________________________________________________________________'
- font: Times-Bold (Type1) size: 10.000000  'Dossier suivi par   -  -  - '
[...]

Issue #570, opened 2023-01-23
calculateTextWidth throws an error for some fonts


benlongstaff:
Not all fonts have the widths data in the font header e.g. $font->getDetails() returns.

```
Array
(
    [Name] => WPJPNX+E13BP
    [Type] => Type0
    [Encoding] => Identity-H
    [BaseFont] => WPJPNX+E13BP
    [DescendantFonts] => Array
        (
            [0] => Array
                (
                    [Name] => WPJPNX+E13BP
                    [Type] => CIDFontType2
                    [Encoding] => Ansi
                    [BaseFont] => WPJPNX+E13BP
                    [CIDToGIDMap] => Identity
                    [DW] => 1000
                    [Subtype] => CIDFontType2
                )

        )

    [Subtype] => Type0
    [ToUnicode] => Array
        (
            [Filter] => FlateDecode
            [Length1] => 1024
            [Length] => 303
        )

)
```

vs the expected

```
Array
(
    [Name] => MyriadPro-Regular
    [Type] => Type1
    [Encoding] => WinAnsiEncoding
    [BaseFont] => MyriadPro-Regular
    [FirstChar] => 1
    [FontDescriptor] => Array
        (
            [Ascent] => 750
            [CapHeight] => 674
            [Descent] => -250
            [Flags] => 32
            [FontName] => MyriadPro-Regular
            [ItalicAngle] => 0
            [StemV] => 80
            [Type] => FontDescriptor
        )

    [LastChar] => 255
    [Subtype] => Type1
    [Widths] => Array
        (
            [0] => 0
            ...
            [254] => 471
        )

)
```

k00ni:
Can you provide us the PDF?

benlongstaff:
Unfortunately the files are Bank Statements, I will need to find a way to remove the elements with sensitive information. 

Is there other information about the font I could provide in the meantime?

k00ni:
The most helpful would be PHP exploit code which triggers the error. In the following (untested) a rough example.

```php
/*
 * $elements must contain faulty data to trigger the error.
 * $header->getDetails() is used inside "calculateTextWidth".
 * If it doesnt return an array with key "Widths", the error occur.
 *
 * You can build $elements yourself or you place var_dump near
 * https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/Font.php#L278
 * and use that.
 */
$elements = [
    'Name' => ''...',
    'Type' => '...',
    'Encoding' => '...',
[...]

PR #630, opened 2023-08-07
fix: Fail to identify fonts (name and size) #629

mbideau-atreal:
In order to solve #629 I identified that _BT_ commands were resetting the font name and size previously defined by a _Tf_ command... all the time.
So I just removed the font (over)initialization from the _BT_ code, and all the font names and size were showing up.
The (first) initialization of `$fontId` and `$fontSize` is already done before the loop starts, BTW.

PS: I don't know anything about PDF standard, so please take my fix with caution.

Thanks for this great piece of code.
Best regards.


k00ni:
CC @GreyWyvern because it related to fonts, which was the topic of one of recent your pull requests.

GreyWyvern:
My own PR that I'm working on removes these same lines from **Page.php** as unnecessary as well. If font can be set outside `BT ... ET` blocks then there is no need to reset it each time we enter a new one.

k00ni:
@mbideau-atreal I need a simple unit-test to prove the change is working and to avoid a regression in the future. Can you do that? If not, I can suggest one.

GreyWyvern:
@mbideau-atreal May we use your sample PDF [20230803-160138-lettretype-arrete.pdf](https://github.com/smalot/pdfparser/files/12275119/20230803-160138-lettretype-arrete.pdf) in the PdfParser test suite? Then we could add to the bottom of PageTest::testGetDataTm():

```php
        // Check that BT and ET do not reset the font
        $config = new Config();
        $config->setDataTmFontInfoHasToBeIncluded(true);

        $filename = $this->rootDir.'/samples/bugs/Issue629.pdf';
        $parser = $this->getParserInstance($config);
        $document = $parser->parseFile($filename);
        $page = end($pages);
        $dataTm = $page->getDataTm();

        $this->assertCount(4, $dataTm[0]);
        $this->assertEquals('F2', $dataTm[0][2]);
```

The first test should always be true if `setDataTmFontInfoHasToBeIncluded` is true. The second test is only true if `$fontId` and `$fontSize` lines are removed from **Page.php**.

I would also recommend removing the same lines from further down under `case 'ET':` as well.

k00ni (approved):
Thank you @GreyWyvern. I took your suggestion and added two tests. PR is ready to merge. @mbideau-atreal is there anything left to do?

mbideau-atreal:
Sorry for my delay.
Great work :+1: 
And yes, you can use my PDF in your test case... thank you very much for asking.

mbideau-atreal:
Nothing left to do here, on my part :wink: 

After the merge, if you can, I would really need some feedback on the issue #570 regarding the same PDF sample document :pray: 

Thanks again.

Issue #488, opened 2021-11-19
Getting false "Secured pdf file are currently not supported."


parijke:
I have a PDF that, when trying to parse the parser says
Secured pdf file are currently not supported.

![image](https://user-images.githubusercontent.com/7671998/142625431-6cb7e113-5f3b-4fae-9957-78f9580d6dbe.png)

Sounds like a false positive to me?

k00ni:
Can you provide us the PDF? We would like to include it in our test environment, therefore it must be free of charge and without any obligations.

parijke:
I've looked into it. If I comment out line 102 - 104 in the Parser, the file get parsed. The problem is, that the trailer HAS a reference to /Encrypt:

here's the trailer:
```
trailer

<<
/Encrypt 17 0 R
/Info 18 0 R
/Root 16 0 R
/Size 19
/ID [<8f6e73e43f6fc457c6d42783e4005ec4><624b2ff302bbf94e2e0f53c90933b5c1>]
>>
startxref
28499
%%EOF
```

/Encrypt [dictionary]: Specifies the document’s encryption dictionary.

But is does not use it. The file is not encrypted. 

So testing a pdf on the value encrypt is not fail safe to really test for secured files.

parijke:
[test.pdf](https://github.com/smalot/pdfparser/files/7576214/test.pdf)

Here's an example file. I created it using LibreOffice. While exporting to PDF, I set the WRITE password, but not the READ password.

So it is not encrypted for reading, which is the case in the file I encountered this at first as well.

Now you have a test file

k00ni:
Thank you for your detailed feedback. I hope someone can pick up the ball and writes a fix.

parijke:
I am looking into it. Looks like the answer could be found here:
https://github.com/empira/PDFsharp/blob/3205bd933b464d150c0f42e8bcdff3314b6c6164/src/PdfSharp/Pdf.Security/enums/PdfUserAccessPermission.cs

parijke:
Additional explanation
https://www.cs.cmu.edu/~dst/Adobe/Gallery/anon21jul01-pdf-encryption.txt

EthosLuke:
Any progress on this?

k00ni:
Not to my knowledge, sorry.

k00ni:
#653 should provide a valid work around until there is a better solution/fix.

Issue #320, opened 2020-07-17
Encrypted but not really _encrypted_ PDF isn't parsed and detected as "secured pdf"


rameezrami:
This is a very good package with excellent and easy to use API but it seems not production-ready yet. I'm having errors with different PDFs. 

also, I'm getting an error saying pdf in encrypted. I double-checked if the PDF is protected, it wasn't.

maybe a version issue? like in another issue posted in this repository.

I have created a screencast for better understanding and these pdf is not getting parsed in your demo page too.

Please find the files in the attachment.

[pdf and screencasts.zip](https://github.com/smalot/pdfparser/files/4936225/pdf.and.screencasts.zip)

cesardmoro:
any advice on this  ?

k00ni:
> any advice on this ?

Sorry, I can't help you on this one.

j0k3r:
I've tried locally and on my Mac the PDF information are saying the PDF is password encrypted.

<img width="255" src="https://user-images.githubusercontent.com/62333/90859012-a2d24200-e387-11ea-9907-c30164059563.png">

Also tried using an online tool and it says the same: https://www.metadata2go.com/result/cffe1bdf-d6be-45d1-b810-f4db8ae846af

![image](https://user-images.githubusercontent.com/62333/90859109-d745fe00-e387-11ea-86fe-7c44599067c4.png)

rameezrami:
@j0k3r but you were able to open the pdf without any issue right? no passwords asked. I have a couple of other pdfs(can't share here, confidential information) too which doesn't work the same way.

j0k3r:
Of course, but the label says the PDF is encrypted. So we don't try to read it at all.
See https://github.com/smalot/pdfparser/blob/master/src/Smalot/PdfParser/Parser.php#L97-L99

Maybe we should improve that and try to read the PDF?

Sagar1219:
Hey I have encountered same issue, is this fixed? Also, meanwhile do we have any alternate solution for this ?

j0k3r:
No alternate solution, but contributions are welcome.

unixnut:
Fix in _unixnut/pdfparser_ fork, 'decryption' branch.  I was going to create a draft pull request but I already have an unrelated PR open for one of the commits on that branch so I can't make another until that's merged.  @k00ni can you please reopen this issue?

matkozikowski:
Hi,
[...]

PR #632, opened 2023-08-14
Add ability to ignore PDF encryption check

DivineOmega:
In some cases PDF files may be internally marked as encrypted even though the content is not encrypted and can be read.

This MR provides a config option to inform the PDF parser to ignore the encryption and attempt to read the PDF anyway.

This therefore provides a work around for the following issues:

* https://github.com/smalot/pdfparser/issues/320
* https://github.com/smalot/pdfparser/issues/488

### DIFF
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -101,7 +101,7 @@ class Parser
         // Create structure from raw data.
         list($xref, $data) = $this->rawDataParser->parseData($content);
 
-        if (isset($xref['trailer']['encrypt'])) {
+        if (isset($xref['trailer']['encrypt']) && $this->config->getIgnoreEncryption() === false) {
             throw new \Exception('Secured pdf file are currently not supported.');
         }
 
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -101,7 +101,7 @@ class Parser
         // Create structure from raw data.
         list($xref, $data) = $this->rawDataParser->parseData($content);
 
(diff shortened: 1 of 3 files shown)


src/Smalot/PdfParser/Parser.php
```diff
@@ -101,7 +101,7 @@ public function parseContent(string $content): Document
         // Create structure from raw data.
         list($xref, $data) = $this->rawDataParser->parseData($content);
 
-        if (isset($xref['trailer']['encrypt'])) {
+        if (isset($xref['trailer']['encrypt']) && !$this->config->getIgnoreEncryption()) {
```
k00ni, line 104: The following change makes the code more readable in my opinion.

```suggestion
        if (isset($xref['trailer']['encrypt']) && false === $this->config->getIgnoreEncryption()) {
```

GreyWyvern:
This is a good addition, but as the OP says, this is a *workaround*. Eventually in the future the simple check in `Parser::parseContent()` should be modified to check if the document *actually* cannot be read.

```php
        if (isset($xref['trailer']['encrypt'])) {
            throw new \Exception('Secured pdf file are currently not supported.');
        }
```

It should be taken into account that a future fix for this would obsolete the use of the config option being added here. That's probably the only thing I *don't* like about this change.

k00ni:
@DivineOmega Are you still with us here?

DivineOmega:
Hi. Sorry for the delayed response. Things have been busy recently.

I didn't end up actually using this functionality myself. I found that a majority of the PDFs I ignored the encryption check for would actually be parsed as containing no text or limited useful text. I'm not sure why this is and so my workaround here ended up not being useful for my use case.

This library still provides some of the best parsing I've found. My solution was to use an alternative parser if this one detected an encrypted PDF.

Issue #458, opened 2021-09-03
Library crash when I tried to parse an 1.3 PDF file


andres28ramirez:
I'm trying different methods but it gaves me this result as error:
count(): Parameter must be an array or an object that implements Countable 
This happend in the line (261) of the Parser:

\vendor\\smalot\\pdfparser\\src\\Smalot\\PdfParser\\Parser.php(261)

[12249.pdf](https://github.com/smalot/pdfparser/files/7108200/12249.pdf)

GreyWyvern:
The sample file [12249.pdf](https://github.com/smalot/pdfparser/files/7108200/12249.pdf) extracts without error in the latest v2.7.0 release.

PR #634, opened 2023-08-18
Major Update to PDFObject.php + Ancillary

### LINKED ISSUE #219: Text differ a lot from Adobe Acrobat reader
[F1.pdf](https://github.com/smalot/pdfparser/files/2683550/F1.pdf)

Hi! I am trying to read a cyrillic file with pdfparser, but text is quite different than what you see, when you copy from Adobe Acrobat. 
Why is that? What can I do to get better result with pdfparser?

Thanks in advance!

### LINKED ISSUE #353: getText() adds the letter "j" to a lot of words in the PDF content
When I parse the attached PDF, the letter "j" is added to random words when I use getText(). For instance, in some cases the term "BrabantWonen" is changed to "BjrabantWonen" and "aanbrengen" becomes "aanbrengejn". Is this a known issue?

[4c011e062aafcc305834aa245734eafc6945c1dc.pdf](https://github.com/smalot/pdfparser/files/5343195/4c011e062aafcc305834aa245734eafc6945c1dc.pdf)

### LINKED ISSUE #398: Hebrew text displayed in reverse
Hello, for first thank you for the great parser.
The issue that when i parsing documents in Hebrew, all the text displayed reverse.

Thank you.

### LINKED ISSUE #464: Duplicating numbers and problem with whitespaces
Hi, i have some issues with parsing pdf in Russia. First of all it's some problem with witespaces and also for some values of numeric i can see dupicates in parsed text. Attaching pdf and some screenshots, Thanks
Screenshot 2021-10-06 at 03-36-46 Demo PDF Parser
[obj39453_pd32-000296.pdf](https://github.com/smalot/pdfparser/files/7290042/obj39453_pd32-000296.pdf)

### LINKED ISSUE #474: PDF only being partly read.

### LINKED ISSUE #508: Unnecessary line breaks inserted
When I parse [this](https://www.livehealthyiowakids.org/wp-content/uploads/sites/3/2021/06/2021-State-Meet-Results-Field.pdf) file, many line breaks are inserted:
```text
                                Marshalltown High SchoolHy-Tek's MEET MANAGER  1:04 PM  6/21/2021  Page 1
2021 Live Healthy Iowa Kids State Track Meet - 6/19/2021
Marshalltown High School	
Results	
 	Long Jump STANDING LJ 8 & Under Division Girls Age	
   Team	Name Finals
Finals Sibley
  7
Sandersfeld, Aubree
1 J5-00.00
DeWitt
  8
Borgen, Gracie
2 J5-00.00
Forest City
  8
```

This can be fixed by disabling this check: https://github.com/smalot/pdfparser/blob/ddb14aedf47c0e9848dc6c183d3bb6f3cdb7f9d5/src/Smalot/PdfParser/PDFObject.php#L274

The output then becomes:
```text
                                Marshalltown High SchoolHy-Tek's MEET MANAGER  1:04 PM  6/21/2021  Page 1
2021 Live Healthy Iowa Kids State Track Meet - 6/19/2021
Marshalltown High School	
Results	
 	Long Jump STANDING LJ 8 & Under Division Girls Age	
   Team	Name Finals
Finals Sibley  7Sandersfeld, Aubree 1 J5-00.00
DeWitt   8Borgen, Gracie 2 J5-00.00
Forest City   8Jenkins, Cleo 3 J5-00.00
```

I implemented a possible solution which allows someone to disable the check with a configuration value: https://github.com/rubenvanerk/pdfparser/commit/27b9a268d741743550e88ddd62533c08a12b5fdd

I could create a PR for this but:
- I don't know what to call this configuration value
- I don't know if this is the best solution
- It does seem like a very specific problem, no other issues seem to mention this or can be fixed with the proposed solution

### LINKED ISSUE #528: Unwanted spaces in Content
- PHP Version: 7.2
 - PDFParser Version: ~~2.1~~ 2.2.0

### Description:
Hey guys, thanks for the hard work to keep this great library up to date. 
Unfortunately, we are having one strange issue parsing a file, we tried adjusting the config like this as @rubenvanerk mentioned in a few issues (in fact the FontSpaceLimit variable solves multiple issues with the char separation). 

```
$config->setHorizontalOffset('');
$config->setFontSpaceLimit(-136);

$pdf = $parser->parseFile("../storage/app/IL-Field-Guide-final-online.pdf");
```

But we cannot get rid of some unwanted empty spaces for example with: 

> "as opposed to persons i n vehicles" should be "as opposed to persons in vehicles" 

And more importantly in the PDF section titles we have: 

> "Consensual Encounters - Tier 1 in Seizure M odel" should be "Consensual Encounters - Tier 1 in Seizure Model"

Additionally, this one is not the expected output too: 

> "Your Powers and the Suspe ct's R ights during a TERRY Stop" needs to be "Your Powers and the Suspect's Rights during a TERRY Stop"

### PDF input
[IL-Field-Guide-final-online.pdf](https://github.com/smalot/pdfparser/files/8514566/IL-Field-Guide-final-online.pdf)

This may be not necessarily an issue, but we are suspecting that for some reason the PDF has a space within the conflicting phrases/sentences. 
In any case, we are starting to use the library, we modified a few things on the vendors folder trying to fix the issue, but we are going out of ideas now.

### LINKED ISSUE #537: getText() returns only line breakes
- PHP Version:  1.4
 - PDFParser Version:  2.1

### Description:
If I use the getText() function of the parsed pdf file, the content is not empty, but it haves no text, only line breaks (\n, \t) inside.

"""
\n
\t\n
\t\n
\t\n
\t\n
\t\t\n
\n
\n
\t\n
\n
\n
\n
\t\n
\t\n
\t\n
\t\n
 
"""

### LINKED ISSUE #564: Does not parse text from pdf file
```
$content = Storage::get('iaa/receipt1.pdf');
$parser = new \Smalot\PdfParser\Parser;
$data = $parser->parseContent($content);
dd($data->getText());
```
Output:
b"%\t\n\x00X\x00\\x00H\x00U\x00\x03\x005\x00H\x00F\x00H\x00L\x00S\x00W\t\n\x005\t\n\n\x00H\x00F\x00H\x00L\x00S\x00W\x00\x03\x00\x06\t\n\x00\x14\t\n \x00\x1C\x00\x19\x00\x17\x00\x1A\x00\x19\x00\x17\x00\x13\t\n\x005\t\n\n\x00H\x00F\x00H\x00L\x00S\x00W\x00\x03\x00'\x00D\x00W\x00H\t\n\x00\x14\t\n \x00\x14\x00\x12\x00\x15\x00\x16\x00\x12\x00\x15\x00\x13\x00\x15\x00\x15\t\n\x005\t\n\n\x00H\x00F\x00H\x00L\x00Y\x00H\x00G\x00\x03\x00%\x00\\t\n\x00$\t\n \x006\x00$\x003\x00\x03\x006\t\n\x006\t\n\n\x00R\x00O\x00G\x00\x03\x00$\x00W\x00\x03\x00%\x00U\x00D\x00Q\x00F\x00K\t\n\x00\x19\t\n 

[receipt1.pdf](https://github.com/smalot/pdfparser/files/10123883/receipt1.pdf)
IMHO parser does not parse font correctly and does not load translate tables.

### LINKED ISSUE #568: Spaces not respected when dealing with numbers in a borderless table
- PHP Version: 7.4.3
 - PDFParser Version: 2.3
[35803_2-21-29-12-2022-R.pdf](https://github.com/smalot/pdfparser/files/10331850/35803_2-21-29-12-2022-R.pdf)
 

### Description:
The PDF has a borderless table with numeric data. The parser does not respect spaces between numbers

### PDF input
PDF Attached

Expected Output
`Patient report Bio-Rad DATE: 12/29/2022 D-10 TIME: 01:45 PM S/N: #DJ0G520325 Software version: 4.30-2 Sample ID: 35803 Injection date 12/29/2022 12:21 PM Injection #: 21 Method: HbA2/F Rack #: --- Rack position: 2 Peak table - ID: 35803 Peak R.time Height Area Area % A1a 0.20 2955 16154 1.2 A1b 0.28 2911 13155 0.9 F 0.46 530 6306 < 0.8 * LA1c/CHb-1 0.64 964 7119 0.5 LA1c/CHb-2 0.73 1442 9759 0.7 A1c 0.88 5093 50721 5.5 P3 1.53 8531 67945 4.9 A0 1.76 259075 1181768 85.2 A2 3.38 2307 34548 2.8 Total Area: 1387477 Concentration: %F < 0.8 *A1c 5.5 A2 2.8`

Actual Output
`Patient report Bio-Rad DATE: 12/29/2022 D-10 TIME: 01:45 PM S/N: #DJ0G520325 Software version: 4.30-2 Sample ID: 35803 Injection date 12/29/2022 12:21 PM Injection #: 21 Method: HbA2/F Rack #: --- Rack position: 2Peak table - ID: 35803 Peak R.time Height Area Area % A1a 0.202955161541.2 A1b 0.282911131550.9 F 0.465306306< 0.8 * LA1c/CHb-1 0.6496471190.5 LA1c/CHb-2 0.73144297590.7 A1c 0.885093507215.5 P3 1.538531679454.9 A0 1.76259075118176885.2 A2 3.382307345482.8 Total Area: 1387477 Concentration: %F < 0.8 *A1c 5.5A2 2.8`

### Code
```php
$PDFfile = $fileEntity->getFileUri();
$PDF = $parser->parseFile($PDFfile);
$PDFContent = $PDF->getText();
echo $PDFContent;
```

### LINKED ISSUE #575: ErrorException     Undefined array key 1
- PHP Version: 8.1
 - PDFParser Version:  "smalot/pdfparser": "^2.3"

I'm getting an error when trying to run getDataTm() on a pdf. I have tried this on a native LEMP stack and inside a Laravel Sail Docker Container. Both the same error message. Any advice would be greatly appreciated. The code is copied from the docs. The getText() function works as expected, but this one doesn't seem to.

### Description:
   ErrorException 

  Undefined array key 1

  at vendor/smalot/pdfparser/src/Smalot/PdfParser/Page.php:852
    848▕                      *
    849▕                      * Source: https://ia902503.us.archive.org/10/items/pdfy-0vt8s-egqFwDl7L2/PDF%20Reference%201.0.pdf
    850▕                      * Introduced with https://github.com/smalot/pdfparser/pull/516
    851▕                      */
   852▕                     list($fontId, $fontSize) = explode(' ', $command['c'], 2);
    853▕                     break;
    854▕ 
    855▕                     /*
    856▕                      * array TJ

### Code
$data = $pdf->getPages()[0]->getDataTm();

### LINKED ISSUE #576: Undefined array key when calling getDataTm()
- PHP Version: 8.0.27
 - PDFParser Version: 2.3

### Description:
When I call `getDataTm` it shows `Undefined array key 71`. This error occurs at `Smalot/PdfParser/Page.php:704`.
The PDF that I use is a PDF generated by mPDF.

### PDF input

[63ed26d07e3d1-erkk-pdf.pdf](https://github.com/smalot/pdfparser/files/10746292/63ed26d07e3d1-erkk-pdf.pdf)

I'm expecting output something like this:
image

However the actual output I get is like this:
image

### Code

```
$parser = new Parser();
$pdf = $parser->parseFile(storage_path("app/$tmpPath"));
$pages = $pdf->getPages();

$dataTm = [];
foreach ($pages as $page) {
    $dataTm[] = $page->getDataTm();
}

return var_dump($dataTm);
```

### LINKED ISSUE #585: getText() result contains annacceptable unicode chars
- PHP Version: 7.3.33
 - PDFParser Version: 2.3.0

### Description:
text, returned by getText() is contaminated with unicode control characters (0x97 hex, 151 dec -- https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C1_controls )

### PDF input
[page92.pdf](https://github.com/smalot/pdfparser/files/11008038/page92.pdf)

### Expected 
actual: "слева	 по ходу	 � веревка"
expected: "слева по ходу — веревка"

Same text with correct em-dash symbol.

### Code

```php
use Smalot\PdfParser\Parser;

$path = 'page92.pdf';

$text = $pdf->getText();
$sub = substr($text, 1658, 43);              // just to extract essential

$json = json_encode($sub);
// json === false, errorcode 5 : Malformed UTF-8 characters, possibly incorrectly encoded
```

### LINKED ISSUE #608: Line breaks being lost between headers and paragraphs
- PHP Version: 8.1
 - PDFParser Version: v2.5.0

### Description:
Line breaks between headings and paragraph text are being lost. Also, the subsequent line breaks after each paragraph are being lost. The PDF was created using macOS Pages' export function, using the following settings:
Screenshot 2023-06-23 at 14 32 03

### PDF input
[Example PDF.pdf](https://github.com/smalot/pdfparser/files/11848963/Example.PDF.pdf)

[expected.txt](https://github.com/smalot/pdfparser/files/11848977/expected.txt)
[actual.txt](https://github.com/smalot/pdfparser/files/11848978/actual.txt)

### Code
```
$parser = new PdfParser\Parser();
$pdf = $parser->parseFile($filePath);
echo $pdf->getText();
```

### LINKED ISSUE #628: PdfParser does not consider the entire document stream
https://github.com/smalot/pdfparser/blob/2608ac3c0db64802ffbe0ce648de7dd0825f2b5d/src/Smalot/PdfParser/PDFObject.php#L195-L201

Above is the code in **PDFObject.php** that extracts lines from a document stream to determine what to display. It only considers content between `BT` and `ET` commands (and maybe a `Q` or `q` on either side) to be valid commands. However, many valid commands such as `cm` (graphics position affecting initial position of `BT`) and `Tf` (font changes) can and do occur outside of `BT ... ET` blocks. Even `q` and `Q` occur regularly in streams and not just adjacent to `BT ... ET`. In order to more correctly display the content of a PDF, the entire stream must be used, with mainly graphics-related commands able to be ignored.

As well, `q` and `Q` are currently handled in a two state manner. If a `q` is encountered, the state is saved; if a `Q` is encountered, the saved state is restored. This does not account for the fact that multiple states can be saved and restored in a stack in a push/pop manner. Both fonts (`Tf`) and graphics positions (`cm`) should be stored in this fashion.

https://github.com/smalot/pdfparser/blob/2608ac3c0db64802ffbe0ce648de7dd0825f2b5d/src/Smalot/PdfParser/PDFObject.php#L387-L390

### Affect on Positioning

In addition to ignoring `cm` positioning commands, PdfParser's treatment of `Tm` (set text matrix) and `Td`/`TD` (set text current point) does not take into account the full matrix position of 6 values. In the following example stream commands:

```
0.8 0 0 0.8 100 100 Tm
200 200 Td
(Hello World)Tj
```

... PdfParser only considers the `100 100` from the `Tm` command and sets that as the current text position. Then it sees the `200 200` from the `Td` and _overwrites_ the current text position so it is now `200 200`. The correct positioning interpretation is the following:

- Set the current text position to `100 100`.
- **Also** set the current text size ratio to `0.8 0.8`.
- Take the values from the `Td` command, multiply them by the text size ratio, then **add** them to the current text position: 200 x 0.8 + 100 = 260
- Therefore the current text position to display "Hello World" is actually `260 260` and not `200 200`.

Fortunately we can ignore the graphics size ratios from the `cm` commands as they only affect graphics commands. :)

I'm preparing a PR that will essentially completely re-write the `cleanContent()`, `getSectionsText()`, `getText()`, and `getCommandsText()` methods from **PDFObject.php** (as well as a couple minor changes in **Font.php** and **Page.php**) to switch to this new way of interpreting the document stream. It is an extensive change which I hope gets a lot of scrutiny! Already in my test environment it is passing all unit tests except one, and resolves a large number of open issues.

Opening this issue for discussion purposes, and I may start tagging issues here that will (hopefully) be resolved by the change.

### LINKED ISSUE #637: Cannot Extract Bengali text
I have a PDF file that contains Bengali text. Texts are not extracted correctly using the getText function.

GreyWyvern:
Okay, here it is! I fully expect this will take quite some time to review and merge and likely require many more commits before it's ready to go; please mark it as a draft if that fits better.

It fully passes all unit tests here (PHP 8.2.7), even a couple that were marked "linux only" from which I removed that criteria. Several existing unit test assertions have been altered simply because of the way the update now parses document stream data differently, and thus generates arrays of commands differently. As examples: `BT` commands (as well as several others) are now stored instead of discarded, and outside of `(strings)` and `<<dictionaries>>` whitespace is normalized.

However, since it parses document stream data in a completely new way, what I'm most interested in is whether or not it causes new errors in PDFs that **aren't** in the test suite. So I hope quite a few people decide to test drive this.

---

## PHPObject.php

This is a major update to the **PHPObject.php** file. Where previously PdfParser would attempt to gather document stream data using a series of multiline regular expressions focusing on `BT ... ET` blocks, this update ~~changes the behaviour of `cleanContent()`~~ adds a new function `formatContent()` that considers the entire document stream. It takes the following steps:
- Hide all `(strings)`
- Remove all newlines and carriage-returns
- Hide all `<<dictionaries>>`
- Perform a MIME-type check for 'text/plain'[^1]
- Normalize all whitespace
- Using a list of valid Operators from the PDF Reference[^2], add \r\n back to the remaining text so that there is a single valid PDF command on every line
- Restore the `<<dictionaries>>` 
- Encode newlines and carriage returns in `(strings)` as `\n` and `\r` and restore them as well

By using this system, it is then much easier to examine and parse the document stream in a line-by-line manner, instead of multiline PCRE extraction. `getSectionsText()` has been updated to do just this, stepping through the output of ~~`cleanContent()`~~ `formatContent()` line-by-line and returning an array of only the relevant commands needed to position and display text.

The guts of `getText()` have been moved to `getTextArray()` to reduce code duplication. `getTextArray()` now takes into account both the current graphics position `cm` and the scaling factors of the text matrix `Tm` when adding \n, \t and space " " whitespace for positioning. Positioning is only taken into account at the point of inserting text, rather than whenever a `Tm` or `Td`/`TD` is found.

`getTextArray()` now also treats `Q` and `q` as a stack of stored states rather than a single stored state. Both font `Tf` and graphics positioning `cm` are stored.

The presence of `ActualText` `BDC` commands is also taken into account and the contents of the `ActualText` will replace the formerly output text in both content and position. This requires the new `parseDictionary()` method to reliably extract such commands as well as any others PdfParser may take into account in the future.

## Font.php

`decodeText()` in **Font.php** now takes into account the current ~~text matrix~~ font size and scale when considering whether or not to define strings of text as "words" that require spaces between them.

In `decodeContent()` in **Font.php** add a check to see if the string to decode has the UTF-16BE BOM and decode it directly as Unicode if so.

## Page.php

In **Page.php** remove the addition of a "fake" BT command as the content stream now records them.

Add a check to see if there are remaining texts to use from `PDFObject::getTextArray()` before proceeding in `getDataTm()` which prevents "undefined array key" PHP errors.

Also prevent `ET` commands from resetting the font `Tf` as PR #629 did for `BT` commands.

## Issues affected

Resolves #219.
Resolves #353.
Resolves #398. Current v2.7.0 fixes text direction, but this PR fixes all spacing issues.
Resolves #464. Removes duplicated text by examining `ActualText` commands.
Resolves #474.
Resolves #508.
Resolves #528. Fixes spacing issues.
Resolves #537.
Resolves #564. Current v2.7.0 fixes text extraction, but this PR fixes spacing issues.
Resolves #568. Fixes spacing issues.
Resolves #575.
Resolves #576.
Resolves #585.
Resolves #608. Fixes headings by taking into account the graphics position `cm`.
Resolves #628.
Resolves #637.

[^1]: This prevents non-document-stream data from being passed to ~~`cleanContent()`~~ `formatContent()` such as JPEG data in file '12249.pdf' from https://github.com/smalot/pdfparser/issues/458
[^2]: https://ia801001.us.archive.org/1/items/pdf1.7/pdf_reference_1-7.pdf - Appendix A; https://archive.org/download/pdf320002008/PDF32000_2008.pdf - Annex A


GreyWyvern:
Hrms. It seems all the Windows PHP checks are failing because the `Fileinfo` functions haven't been installed. They're installed by default on Linux, but the Installation manual says: 'Windows users must include the bundled `php_fileinfo.dll` DLL file in `php.ini` to enable this extension.' So Windows users have the DLL file, but I guess it's just not enabled in a "default" install?

For "PHPUnit coverage (PHP 7.4)" it is failing an assertion where it tests to see if PDFParser's memory usage is higher than a checked value, but with my edits, the memory usage is actually **lower** than it expects, so that's probably actually a *win*? :D

src/Smalot/PdfParser/Font.php
```diff
@@ -414,18 +414,25 @@ protected function getFontSpaceLimit(): int
     }
 
     /**
-     * Decode text by commands array.
+     * Decode text by commands array. decodeText now inserts the proper
+     * whitespace by examining the scaling of the current text matrix.
+     * Supply the identity matrix '1 0 0 1' by default.
```
k00ni, line 419: Please add a comment into the code about why using 1 0 0 1 here.

src/Smalot/PdfParser/Font.php
```diff
@@ -414,18 +414,25 @@ protected function getFontSpaceLimit(): int
...
-    {
+    public function decodeText(
+        array $commands,
+        array $textMatrix = ['a' => 1, 'b' => 0, 'i' => 0, 'j' => 1]
+    ): string {
         $word_position = 0;
         $words = [];
+
         $font_space = $this->getFontSpaceLimit();
+        $font_space = $font_space * $textMatrix['a'] + $font_space * $textMatrix['b'];
```
k00ni, line 429: Can we be sure that key `a` and `b` are always set? Please add the explanation above the line.

src/Smalot/PdfParser/Font.php
```diff
@@ -468,6 +475,12 @@ public function decodeText(array $commands): string
      */
     public function decodeContent(string $text, bool &$unicode = null): string
     {
+        // If this string begins with a UTF-16BE BOM, then decode it
+        // directly as Unicode
+        if (preg_match('/^\xFE\xFF/i', $text)) {
```
k00ni, line 480: Just a question: Why not use `substr` to get first two characters and check them instead? Using `preg_match` might be more "expensive".

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -127,100 +132,195 @@ public function getContent(): ?string
         return $this->content;
     }
 
-    public function cleanContent(string $content, string $char = 'X')
+    public function cleanContent(?string $content): string
```
k00ni, line 135: That is a backward compatibility break, because you removed a parameter from a public function. Can you tell me, how it was used until then?

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -127,100 +132,195 @@ public function getContent(): ?string
...
     {
-        $char = $char[0];
-        $content = str_replace(['\\\\', '\\)', '\\('], $char.$char, $content);
-
-        // Remove image bloc with binary content
-        preg_match_all('/\s(BI\s.*?(\sID\s).*?(\sEI))\s/s', $content, $matches, \PREG_OFFSET_CAPTURE);
-        foreach ($matches[0] as $part) {
-            $content = substr_replace($content, str_repeat($char, \strlen($part[0])), $part[1], \strlen($part[0]));
+        if (null == $content) {
```
k00ni, line 137:
```suggestion
        if (null === $content) {
```

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -127,100 +132,195 @@ public function getContent(): ?string
...
+            return '';
         }
 
-        // Clean content in square brackets [.....]
-        preg_match_all('/\[((\(.*?\)|[0-9\.\-\s]*)*)\]/s', $content, $matches, \PREG_OFFSET_CAPTURE);
-        foreach ($matches[1] as $part) {
-            $content = substr_replace($content, str_repeat($char, \strlen($part[0])), $part[1], \strlen($part[0]));
+        // Find all strings () and replace them so they aren't affected
```
k00ni, line 141: What do you mean with strings here? These brackets? `()`

k00ni:
@GreyWyvern Thank you very much for all your work. I will see how I manage to give more feedback soon, but I hope that our community has time to comment too. Surprisingly I can't mark this PR as draft.

src/Smalot/PdfParser/PDFObject.php
GreyWyvern: `cleanContent()` was (and still is) only called once by `getSectionsText()` in the entire of PdfParser. Previously, `cleanContent()` would behave... similarly to how it does in this PR. It would hide strings `()` and dictionaries `<<>>`, but also unnecessarily hide arrays `[]` as well. It would do this by replacing the entire item by a string of the same length made up of the `$char` character. So:

```
(Modificatio[ns] au \\(14\\) septembre 2009 ET 2010)Tj
```

... would become:

```
(________________________________________________)Tj
```

It would also unnecessarily hide `BDC` and `EMC` commands. But the hiding was not done for the same purpose as it is in this PR.

The current v2.7.0 version of `getSectionsText()` would accept the full, uncleaned content stream, then make a `cleanContent()` duplicate of it. It would then apply regexps to match "text blocks" and "'do' commands" from the 'cleaned' content. The offset values of those matches were then used to `substr()` the desired text from the *actual* content string.

However, during this process, the only time `q` and `Q` were acknowledged were if they were directly adjacent to the `BT` or `ET` of a text block. Likewise, the only time `BDC` and `EMC` commands were acknowledged were when they were *inside* a `BT ... ET` text block.

---

This PR changes the mission of `cleanContent()` from being a function that just makes a copy of the content with some items blanked out to make matching easier, to being a workhorse that **cleans** the content that actually gets worked on. It hides strings `()` and dictionaries `<<>>` like the old `cleanContent()` did, however once the rest of the document stream is correctly formatted, it *restores* these items so the content stream is complete. It ensures there is one and only one PDF operator command on each line.

`getSectionsText()` then, has been modified from its previous function of grabbing `BT ... ET` blocks by regexp, to stepping through the entire content stream, command by command, and *discarding* the commands we don't need.

Conclusion: I believe that the previous incarnation of `cleanContent()` didn't need to be a `public` function at all. It was only ever called once, and then only to make an incomplete copy for `getSectionsText()` to find offset values from. The output from it isn't used otherwise.

**However**: Now that `cleanContent()` returns a precisely formatted version of the *complete* content stream with one command per line, it now probably *does* have value as a public function, as users can call it to format arbitrary content streams for whatever reason they like.

src/Smalot/PdfParser/PDFObject.php
GreyWyvern: In a document-stream context, anything delimited by parentheses are strings.

> The following are valid literal strings:
> 
>   (This is a string)
>   (Strings may contain newlines
>   and such.)
>   (Strings may contain balanced parentheses () and
>   special characters (*!&}^% and so on).)
>   (The following is an empty string.)
>   ()
>   (It has zero (0) length.)

- PDF Reference 1.7, Section 3.2.3 String Objects

src/Smalot/PdfParser/Font.php
GreyWyvern: I used `preg_match()` only because the existing `decodeUnicode()` already uses the same check. I can certainly change both. :)

k00ni: Only change if you see fit. It is up to you.

k00ni:
> However, since it parses document stream data in a completely new way, what I'm most interested in is whether or not it causes new errors in PDFs that aren't in the test suite. So I hope quite a few people decide to test drive this.

Can you elaborate a bit about which kind of PDFs are not there yet? We could "reproduce" some missing things by using unit tests instead.

src/Smalot/PdfParser/PDFObject.php
k00ni: My only issue here is that changing parameters/return values of public functions is considered a backward compatibility break according to https://semver.org/#spec-item-8. But, it seems that the function in question is not really part of our public API. PDFObject is a mother class for other sub classes and not directly accessible be users (for most of the time).

On the other hand, developers might directly use these functions to tune their software, because they are public and are not marked or commented in any way to decline that form of usage.

Whatever we decide, please add `@internal For internal use only, not part of the public API` to the function header to signal developers to not rely on it in the future (About **internal** in PHP Doc: https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.internal.pkg.html).

I tend to go with your suggestion and keep your changes. However, I will add a short info about this change in the release notes. A release candidate first, before a stable reason might also be a good idea.

GreyWyvern: Since `getSectionsText()` was the only function using `cleanContent()` perhaps I could *restore* the old function and then rename the one I've created to something new? It'll leave some old, unused code in the source, but it wouldn't be a backwards-incompatible change anymore. What do you think? I could call it something like... `formatContent()`?

k00ni:
> I just meant unknown PDFs "in-the-wild" in general. I run it on my own org's collection of PDFs (~400) for searching and it works fine. Almost all of them worked in v2.7.0 too, but those are PDFs all generated by one entity. There's just no telling, with a change this large, if there might be a PDF out there that works in v2.7.0 but not with this PR.

I created a few PDFs using the following tools:

* Chromium browser (export web page as PDF)
* LibreOffice writer 
* Inkscape (PDF v1.4 and v1.5)

I found in the Chromium one that there are too many white spaces added in the result of `getText`. I marked it: https://github.com/smalot/pdfparser/pull/634/files#diff-50498d428341deceb2ce9d29159c86631d951c9260817d46c4f333fe7feecfabR311-R312

Does this help? I am not sure if that approach gonna work, but we could generate more PDFs using other PDF generators and check their results.

GreyWyvern:
> Does this help?

Oh yeah, this helps a lot! Thanks! Definitely some of the numbers need tweaking. I'll see if I can generate a couple myself.

src/Smalot/PdfParser/PDFObject.php
GreyWyvern: I restored `cleanContent()` to its previous version and renamed my function `formatContent()`. `cleanContent()` is now no longer called anywhere in PdfParser, but if any user relied on it, it's still there to use. It should probably be deprecated and then removed in future major versions.

GreyWyvern:
I feel like this is getting close, but we should still keep throwing more PDFs at it for a while.

src/Smalot/PdfParser/PDFObject.php
GreyWyvern: Is this satisfactorily resolved, @k00ni ?

k00ni:
What other PDF generators are worth using to generate PDFs?

Reqrefusion:
Wow this is truly 2.0. If you are looking for PDFs created in different ways, I recommend public documents (Laws, regulations, announcements, etc.). Since I use PDFparser for these, I can guarantee that all of them are created in completely different ways that you cannot imagine. Especially public announcements are a complete mess from different sources.

k00ni:
@Reqrefusion thank you for the hint and its good to see you around.

I made a few adaptions and additions:
* reorganized samples folder and grouped all new PDFs into `grouped-by-generator`
* split DocumentTest class into three test classes, so it is easier to maintain and extend in the future:
  * one for general tests
  * one for issue related tests
  * one for generator related tests
* Added further PDFs, generated by generators we might not cover yet. Also added tests for each of these PDFs. 
  * Two tests will fail because of images, which were not found (like https://github.com/smalot/pdfparser/pull/634/files#diff-4a0191f73e0276217a40bb93e3ce61b37d440b108b90f6d6e61f5bf53d00cb08R140-R143)
  * One test will fail because of some decoding/encoding issue (I assume) - https://github.com/smalot/pdfparser/pull/634/files#diff-4a0191f73e0276217a40bb93e3ce61b37d440b108b90f6d6e61f5bf53d00cb08R227

k00ni:
@GreyWyvern I took another look at some of your new functions. Can you mark all functions `@internal` which are only meant for internal use?

GreyWyvern:
> * Added further PDFs, generated by generators we might not cover yet. Also added tests for each of these PDFs. I let a few tests fail to outline certain areas in question (images + encoding).
>   
>   * Two tests will fail because of images, which were not found (like https://github.com/smalot/pdfparser/pull/634/files#diff-4a0191f73e0276217a40bb93e3ce61b37d440b108b90f6d6e61f5bf53d00cb08R140-R143)

I am reasonably certain I haven't touched any of the image extraction code. And yes, these two tests also fail (expected: 2, actual: 0) in the currently released v2.7.0. They can probably be removed or worked into another image-focused PR.

>   * One test will fail because of some decoding/encoding issue (I assume) - https://github.com/smalot/pdfparser/pull/634/files#diff-4a0191f73e0276217a40bb93e3ce61b37d440b108b90f6d6e61f5bf53d00cb08R227

Ah! This is due to an oversight in my previous PR #627 where it does not check far enough ahead for escaped backslashes. The PDF string in the document at this point is '`\000\\\000`'.

**Current:**
```php
    /**
     * Decode string with octal-decoded chunks.
     */
    public static function decodeOctal(string $text): string
    {
        $parts = preg_split('/(?<!\\\\)(\\\\[0-7]{1,3})/s', $text, -1, \PREG_SPLIT_NO_EMPTY | \PREG_SPLIT_DELIM_CAPTURE);
        $text = '';
```

**Fixed:**
```php
    /**
     * Decode string with octal-decoded chunks.
     */
    public static function decodeOctal(string $text): string
    {
        $parts = preg_split('/(?<![^\\\\]\\\\)(\\\\[0-7]{1,3})/s', $text, -1, \PREG_SPLIT_NO_EMPTY | \PREG_SPLIT_DELIM_CAPTURE);
        $text = '';
```

I can bundle it into this PR, or I can make a new PR for it if you like.

> Hope it helps.

Sure, thanks! I will take a look at your other notes (unnecessary tabs and such) and see if they're fixable.

> @GreyWyvern I took another look at some of your new functions. Can you mark all functions `@internal` which are only meant for internal use?

I created two major new functions for this PR: `formatContent()` and `parseDictionary()`. I actually think both could be useful for end users to know about, but OTOH, you **do** need the raw PDF document stream output in order to use either of them, so any use would be pretty _niche_. I'll rely on your judgement @k00ni, if you think they should be marked as `@internal` I'll do that.

k00ni:
@GreyWyvern

> I'll rely on your judgement @k00ni, if you think they should be marked as `@internal` I'll do that.

Please only mark the functions you added in this PR with `@internal`. If you are unsure, leave it as it is. I would like to keep our public API as small as possible for now or at least not let it expand it uncontrolled.

**About the images**: Just remove them then, if you didn't change any image related code. 

---
I thought about the size of this PR and would like to know, if there is a reasonable way to split this PR up, so we can start to integrate it iteratively? There is no rush here, but on the other hand there is only feedback if people can try it out in the wild. It is up to you @GreyWyvern, its just a suggestion.

GreyWyvern:
> Please only mark the functions you added in this PR with `@internal`. If you are unsure, leave it as it is. I would like to keep our public API as small as possible for now or at least not let it expand it uncontrolled.

Will do! These functions are not really something a regular user would use, so `@internal` makes enough sense.

> **About the images**: Just remove test-code then, if you didn't change any image related code.

Removed.

> I thought about the size of this PR and would like to know, if there is a reasonable way to split this PR up, so we can start to integrate it iteratively? There is no rush here, but on the other hand there is only feedback if people can try it out in the wild (using release candidates). It is up to you @GreyWyvern, its just a suggestion.

Well... the main change of function is in the two functions `formatContent()` and `getSectionsText()`. Especially with the latter, this is an all-or-nothing switch. Without the new way the document stream is formatted by these two functions, the rest of the changes (especially `getTextArray()`) won't really work.

The BOM-related changes in **Font.php**, and all the changes in **Page.php** (_except_ removing lines 403-404) are really the only changes that are essentially entirely unrelated to the meat of this PR. I will revert my change to `Font::decodeOctal()` in prep for the other PR I've made.

I think at most I would be able to split this into two PRs, the first of which would be all the additions that don't actually change execution flow (like adding `formatContent()` and `parseDictionary()` but not calling them anywhere). And then the ones that **do** affect execution flow second. Personally, that just feels like unnecessary extra paperwork to me. I would rather it be merged all at once, and then we do release candidates instead of a full release.

I'm not in a hurry. I'm happy to allow you all the time you need to review :)

k00ni:
> Personally, that just feels like unnecessary extra paperwork to me. I would rather it be merged all at once, and then we do release candidates instead of a full release.

Alright. If there are no further objections, we can merge it all at once. 

How is your timetable for this PR? 

CC @j0k3r @smalot

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -133,14 +133,13 @@ public function getContent(): ?string
...
-     * formatContent() instead, this function is no longer used, and
-     * could be deleted in a future version of PDFParser.
+     * This function is no longer used, and could be deleted in a
+     * future version of PDFParser.
      *
-     * @internal For internal use only, not part of the public API
+     * @internal Creates a duplicate of the document stream with
+     * strings and other items replaced by $char. Formerly
+     * getSectionsText() used this output to more easily gather offset
+     * values to extract text from the *actual* document stream.
```
k00ni, line 142:
```suggestion
     * values to extract text from the *actual* document stream.
     *
     * @deprecated Function is no longer used and will be removed in a future release.
```

k00ni, line 142:
```suggestion
     * Creates a duplicate of the document stream with
     * strings and other items replaced by $char. Formerly
     * getSectionsText() used this output to more easily gather offset
     * values to extract text from the *actual* document stream.
     *
     * @internal
     * @deprecated
```

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -202,10 +201,11 @@ public function cleanContent(string $content, string $char = 'X')
...
     /**
-     * Takes a string of PDF document stream text and formats it into
-     * a multi-line string with one PDF command on each line, separated
-     * by \r\n. If the given string is null, or binary data is detected
-     * instead of a document stream then return an empty string.
+     * @internal Takes a string of PDF document stream text and formats
+     * it into a multi-line string with one PDF command on each line,
+     * separated by \r\n. If the given string is null, or binary data
+     * is detected instead of a document stream then return an empty
+     * string.
```
k00ni, line 208:
```suggestion
     * Takes a string of PDF document stream text and formats
     * it into a multi-line string with one PDF command on each line,
     * separated by \r\n. If the given string is null, or binary data
     * is detected instead of a document stream then return an empty
     * string.
     *
     * @internal
```

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -325,10 +325,11 @@ public function formatContent(?string $content): string
...
     /**
-     * getSectionsText() now takes an entire, unformatted document
-     * stream as a string, cleans it, then filters out commands that
-     * aren't needed for text positioning/extraction. It returns an
-     * array of unprocessed PDF commands, one command per element.
+     * @internal getSectionsText() now takes an entire, unformatted
+     * document stream as a string, cleans it, then filters out
+     * commands that aren't needed for text positioning/extraction. It
+     * returns an array of unprocessed PDF commands, one command per
+     * element.
```
k00ni, line 332:
```suggestion
     * getSectionsText() now takes an entire, unformatted
     * document stream as a string, cleans it, then filters out
     * commands that aren't needed for text positioning/extraction. It
     * returns an array of unprocessed PDF commands, one command per
     * element.
     *
     * @internal
```

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -421,10 +422,9 @@ private function getDefaultFont(Page $page = null): Font
     }
 
     /**
-     * Decode a '[]TJ' command and attempt to use alternate fonts if
-     * the current font results in output that contains Unicode control
-     * characters. See Font::decodeText for a full description of
-     * $textMatrix
+     * @internal decode a '[]TJ' command and attempt to use alternate
+     * fonts if the current font results in output that contains
+     * Unicode control characters
```
k00ni, line 427:
```suggestion
     * Decode a '[]TJ' command and attempt to use alternate
     * fonts if the current font results in output that contains
     * Unicode control characters.
     *
     * @internal
```

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -459,8 +459,8 @@ private function getTJUsingFontFallback(Font $font, array $command, Page $page =
     }
 
     /**
-     * Expects a string that is a full PDF dictionary object, including
-     * the outer enclosing << >> angle brackets.
+     * @internal expects a string that is a full PDF dictionary object,
+     * including the outer enclosing << >> angle brackets
```
k00ni, line 463:
```suggestion
     * Expects a string that is a full PDF dictionary object,
     * including the outer enclosing << >> angle brackets
     *
     * @internal
```

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -556,9 +556,13 @@ public function parseDictionary(string $dictionary): array
...
-     * getText() leverages getTextArray() to get the content of the
-     * document, setting the addPositionWhitespace flag to true so
-     * whitespace is inserted in a logical way for reading by humans.
+     * Returns the text content of a PDF as a string. Attempts to add
+     * whitespace for spacing and line-breaks where appropriate.
+     *
+     * @internal getText() leverages getTextArray() to get the content
+     * of the document, setting the addPositionWhitespace flag to true
+     * so whitespace is inserted in a logical way for reading by
+     * humans
```
k00ni, line 565:
```suggestion
     * getText() leverages getTextArray() to get the content
     * of the document, setting the addPositionWhitespace flag to true
     * so whitespace is inserted in a logical way for reading by
     * humans
     *
     * @internal
```

k00ni (changes requested):
I changed some function headers. As far as I understand the [doc](https://docs.phpdoc.org/2.9/references/phpdoc/tags/internal.html), `@internal` should reflect why a function is for internal use only. It was accompanied with general information here in some cases.

src/Smalot/PdfParser/PDFObject.php
GreyWyvern: This function should not be @internal. I was just using it to mark text as only appearing in the internal documentation, at least that's the impression I got from the phpdoc site I was looking at. I'll switch to the tagging method instead.

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -186,41 +201,206 @@ public function cleanContent(string $content, string $char = 'X')
...
+    /**
+     * Takes a string of PDF document stream text and formats
+     * it into a multi-line string with one PDF command on each line,
+     * separated by \r\n. If the given string is null, or binary data
+     * is detected instead of a document stream then return an empty
+     * string.
+     *
+     * @internal
+     */
+    public function formatContent(?string $content): string
```
j0k3r, line 213: Why don't you define it as `private` instead? It'll avoid adding the `@internal`

j0k3r (commented):
This is massive rewrite and it's hard to follow up 😅

src/Smalot/PdfParser/PDFObject.php
GreyWyvern: Well, it's called explicitly in **PDFObjectTest.php**, but I can work around that with `ReflectionMethod`. Initially I thought it would be useful to be able to use this method publicly to format any old document stream, but that's probably not necessary.

j0k3r: Maybe this can be applied to other `@internal` method?

k00ni: The less public methods we have the better, because `@internal` is just a label and can't be enforced.

_About the testing of private methods:_ I support the view that private functionality is not to test, because it is the objects obligation. One should only test public methods. In practice this might lead to complicated situations in which it is hard to cover a certain case sometimes.

> Initially I thought it would be useful to be able to use this method publicly to format any old document stream, but that's probably not necessary.

I was thinking that a method like that should be extracted from PDFObject and moved to a utility class or something. It is useful and might be handy outside of PDFObject context. But we should finalize this one first and then see. Because it is private now, we could extract it from PDFObject and make it available later on, if we want.

GreyWyvern: > I was thinking that a method like that should be extracted from PDFObject and moved to a utility class or something. It is useful and might be handy outside of PDFObject context.

Definitely. For instance, I feel like my added function `parseDictionary()` duplicates much of the `protected` parsing functionality from **Parser.php** that breaks apart PDF 'dictionary' structures, which aren't only used in document streams, but trailer info, *etc*. In the future, there should probably be one global class or function that parses dictionaries (a fundamental PDF data structure) for all situations.

k00ni:
@GreyWyvern A quick follow up: Are you planning any updates on this one in the near future? I won't have that much time in the next weeks/months most likely. I remember your suggestion to "throw further PDFs" at the code.

I suggest the following next steps:
* You finalize your work
* You mark the PR as ready to review
* I will merge it (if there are no objections)
* I will create a release candidate, so our community can start to test it

WDYT?

GreyWyvern:
> WDYT?

I think this is a good plan. I **have** actually found a couple more PDFs that PdfParser has trouble with since I last visited, but related to embedded images rather than text-extraction.

Putting out a release candidate will certainly help get the new code more testing. The most important thing to find are PDFs that v2.7.0 parses "correctly" while this new version does not. That's where the most tweaking information will come from. :)

I will mark this PR as Ready for review.

k00ni:
@j0k3r any objections? I would merge it right away and proceed as [suggested](https://github.com/smalot/pdfparser/pull/634#issuecomment-1788528712).

k00ni:
Big thanks to @GreyWyvern and all commentators.

PR #635, opened 2023-08-18
Check for wrong line-endings when getting xref

GreyWyvern:
If we didn't find the `xref` command at the offset specified, then replace Windows `\r\n` line endings with Unix style `\n` and try again. If it succeeds, then edit the line-endings and proceed as normal. Otherwise continue on to the `decodeXrefStream()` method.

Fixes parsing of existing test suite file **/samples/bugs/Issue95_ANSI.pdf** the unit test for which would normally be passed over because of the `@group linux-only` flag. Remove this flag, as all assertions in the `testDecodeText()` function now resolve as true in any environment.

### DIFF
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -901,8 +901,15 @@ class RawDataParser
             // Cross-Reference
             $xref = $this->decodeXref($pdfData, $startxref, $xref);
         } else {
-            // Cross-Reference Stream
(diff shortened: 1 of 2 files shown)


GreyWyvern:
I would like to use **Issue95_ANSI.pdf** to create a few new assertions in `testDecodeText()` for my other PR #634. So if we can get this one merged first, that would be much appreciated! :)

k00ni (approved):
Thank you! Is it ready to merge?

GreyWyvern:
Yes, all good. Thanks!

PR #640, opened 2023-09-19
Better octal and hex-entity decode

GreyWyvern:
# Type of pull request

* [ ] Documentation update
* [ ] Something else

# About

A follow-up and improvement over #627.

Octal strings can include series of backslashes of arbitrary length. If there is an odd number of backslashes, a following octal code is valid, but if there's an even number, the following octal code should not be translated.

Previously PdfParser would only account for two backslashes directly preceding an octal code. A commit from in-progress PR #634 extended this to three which probably covers 99.99% of all cases. This change ups that to 100% in that there could be a string with any number of backslashes in a row, and codes will be correctly translated.

Also update decodeEntities() to use a preg_replace_callback() instead of the bulkier preg_split() + foreach loop. Make sure it matches all hexadecimal digits including a-f.

Add new tests for both of these.

     By the way, you don't have to provide a full fledged PDF file to demonstrate a fix. Instead a unit test may be sufficient sometimes, please have a look at [FontTest](https://github.com/smalot/pdfparser/blob/master/tests/PHPUnit/Unit/FontTest.php#L40) for example code.

### DIFF
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -349,16 +349,20 @@ class Font extends PDFObject
      */
     public static function decodeOctal(string $text): string
     {
-        $parts = preg_split('/(?<!\\\\)(\\\\[0-7]{1,3})/s', $text, -1, \PREG_SPLIT_NO_EMPTY | \PREG_SPLIT_DELIM_CAPTURE);
-        $text = '';
+        // Replace all double backslashes \\ with a special string
+        $text = strtr($text, ['\\\\' => '[**pdfparserdblslsh**]']);
 
-        foreach ($parts as $part) {
-            if (preg_match('/^\\\\[0-7]{1,3}$/', $part)) {
-                $text .= \chr(octdec(trim($part, '\\')));
-            } else {
-                $text .= str_replace(['\\\\', '\\(', '\\)'], ['\\', '(', ')'], $part);
-            }
-        }
+        // Now we can replace all octal codes without worrying about
+        // escaped backslashes
+        $text = preg_replace_callback('/\\\\([0-7]{1,3})/', function ($m) {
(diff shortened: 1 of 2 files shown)


k00ni (approved):
Nice and clean. Thank you again. :+1:

src/Smalot/PdfParser/Font.php
```diff
@@ -368,16 +372,9 @@ public static function decodeOctal(string $text): string
...
-                $text .= \chr(hexdec(trim($part, '#')));
-            } else {
-                $text .= $part;
-            }
-        }
+        $text = preg_replace_callback('/#([0-9a-f]{2})/i', function ($m) {
+            return \chr(hexdec($m[1]));
+        }, $text);
 
         return $text;
```
k00ni, line 379:
```suggestion
        return preg_replace_callback('/#([0-9a-f]{2})/i', function ($m) {
            return \chr(hexdec($m[1]));
        }, $text);
```

src/Smalot/PdfParser/Font.php
```diff
@@ -349,16 +349,20 @@ public static function decodeHexadecimal(string $hexa, bool $add_braces = false)
...
+            return \chr(octdec($m[1]));
+        }, $text);
+
+        // Unescape any parentheses
+        $text = str_replace(['\\(', '\\)'], ['(', ')'], $text);
+
+        // Replace instances of the special string with a single backslash
+        $text = str_replace('[**pdfparserdblslsh**]', '\\', $text);
 
         return $text;
```
k00ni, line 367:
```suggestion
        return str_replace('[**pdfparserdblslsh**]', '\\', $text);
```

k00ni:
Ready for merge?

GreyWyvern:
> Ready for merge?

Yes. All done. 👍

Issue #619, opened 2023-07-26
Calculating text width and incorrect result array paths


gidzr:
Hey there

Thanks so much for writing this wrapper. It's probably the only decent one out there for pdf parsing.

I've come across a few issues from the readme and usage against the behaviour of the project.

Firstly - I've been trying to work out how to get width and height of the bounding box for text (not the page width/height).

I came across your 'calculateTextWidth' function, trying it per your Usage instructions and attempting to do it my own way, but nothing works.

Btw, my pdfFile is a standard PDF created from an MSWord docx (text based), using the SaveAs pdf with MSWord. It comprises of 4 lines of text only. A very basic industry standard PDF.

I've also configed the parser 
use Smalot\PdfParser\Config;
$config = new Config();
$config->setDataTmFontInfoHasToBeIncluded(true);
use Smalot\PdfParser\Parser;
$parser = new Parser([], $config);

/////////////////////////////////////////////////////////////////////////////////////////////////////////////
////1.  When running your usage approach (https://github.com/smalot/pdfparser/blob/master/doc/Usage.md), I get a resulting error:
$pdf = $parser->parseFile($pdfFile);
$data = $pdf->getPages()[0]->getDataTm();
$text = $data[0][1];
$font = reset($pdf->getFonts());     <------- line 299
$width = $font->calculateTextWidth($text);

A PHP Error was encountered
Severity: Notice
Message: Only variables should be passed by reference
Line Number: 299

/////////////////////////////////////////////////////////////////////////////////////////////////////////////
////2.  When running your usage approach (https://github.com/smalot/pdfparser/blob/master/doc/Usage.md), I get a resulting error:
$pdf = $parser->parseFile($pdfFile);
$data = $pdf->getPages()[0]->getDataTm();
$fonts = $pdf->getFonts();
$font_id = $data[0][2]; //R7
$font = $fonts[$font_id];     <------- line 310
$text = $data[0][1];
$width = $font->calculateTextWidth($text);

A PHP Error was encountered
Severity: Warning
Message: Undefined array key "F1"
Line Number: 310

/////////////////////////////////////////////////////////////////////////////////////////////////////////////
////3.  When I adjust all the approach based on the arrays I'm actually getting out of the parser
$pdf = $parser->parseFile($pdfFile);
$data = $pdf->getPages()[0]->getDataTm();
$fonts = $pdf->getPages()[0]->getFonts();
$font_id = $data[0][2]; //R7
$font = $fonts[$font_id];
[...]

PR #644, opened 2023-09-29
Usage.md: change example code to avoid PHP notice

k00ni:
# Type of pull request

* [x] Documentation update

# About

Usage documentation contains buggy code. The following notice is thrown when used:

> Only variables should be passed by reference

because of the following code part:

> $font = reset($pdf->getFonts());

## Acknowledgement

Reported by @gidzr in #619


k00ni:
Ping @j0k3r

j0k3r (approved):

PR #645, opened 2023-09-29
Font.php: hot fix for calculateTextWidth in case Widths key is not set

k00ni:
# Type of pull request

# About

Hot fix for Font.php to avoid PHP warning in case the Widths-key is not set in `$details`.

     By the way, you don't have to provide a full fledged PDF file to demonstrate a fix. Instead a unit test may be sufficient sometimes,
     please have a look at [FontTest](https://github.com/smalot/pdfparser/blob/master/tests/PHPUnit/Unit/FontTest.php#L40) for example code.

### Acknowledgement

Bug reported in #619 by @gidzr.

### DIFF
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -283,7 +283,9 @@ class Font extends PDFObject
     {
         $index_map = array_flip($this->table);
         $details = $this->getDetails();
-        $widths = $details['Widths'];
+
+        // TODO: Temporary fix, in the rare case the Widths-key is not set in $details array.
+        $widths = $details['Widths'] ?? [];
 
         // Widths array is zero indexed but table is not. We must map them based on FirstChar and LastChar
         $width_map = array_flip(range($details['FirstChar'], $details['LastChar']));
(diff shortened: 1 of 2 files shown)


src/Smalot/PdfParser/Font.php
```diff
@@ -283,7 +283,9 @@ public function calculateTextWidth(string $text, array &$missing = null): ?float
     {
         $index_map = array_flip($this->table);
         $details = $this->getDetails();
-        $widths = $details['Widths'];
+
+        // TODO: Temporary fix, in the rare case the Widths-key is not set in $details array.
```
j0k3r, line 287: Why it's temporary? Do you plan to create a better fix?

k00ni: No, but because I was not able to find the origin of the error I wanted to at least mark it for future refactoring.

What would you suggest here?

j0k3r: Well regarding the various code of `getDetails()` we can never be sure that a key is defined. So maybe, we should always handle the case where the given key (`Widths` here) is not defined.

k00ni:
I refined the wording of my comment. Such changes might be needed in other places, but for now that should be sufficient.

j0k3r (approved):

PR #669, opened 2024-01-24
Baseencoding fallback

### LINKED ISSUE #665: Attempting to parse a PDF with a form field with two dots in it causes pdfparser to throw an exception
- PHP Version: 8.3.1
 - PDFParser Version: 2.7.0

### Description:
Attempting to parse a PDF with a form field with two dots in it causes pdfparser to throw an exception. Adobe Acrobat reads the PDF just fine as does `qpdf test.pdf --qdf test.qdf`.

### PDF input
[test.pdf](https://github.com/smalot/pdfparser/files/13881204/test.pdf)

The expected output is for it not to throw an exception. The actual output is an exception:

```
Fatal error: Uncaught Smalot\PdfParser\Exception\EncodingNotFoundException: Missing encoding data for: ""
```

### Code
```
<?php
require __DIR__ . '/vendor/autoload.php';

$pdf = $parser->parseFile('test.pdf');

$objects = $pdf->getObjects();
foreach ($objects as $obj) {
	print_r($obj->getDetails());
}
```

GreyWyvern:
When a document doesn't include a `BaseEncoding` header, `StandardEncoding` should be assumed as the default instead of an empty string.

# Type of pull request

# About

Some documents which are short-and-sweet may not include a `BaseEncoding` header. In this case, the PDF Reference 1.7 describes this encoding as a default.

If the result of checking for the `BaseEncoding` returns an empty string, use `StandardEncoding` as the value instead. Resolves #665.


GreyWyvern:
PHP CS Fixer is complaining about indentation in `Document.php`, `PDFObject.php` and `RawData\RawDataParser.php`. Files I didn't even modify. :( Running PHP CS Fixer on my local (Windows) machine doesn't find these issues either.

k00ni:
I merged #670 into master which fixes these coding style issues.

src/Smalot/PdfParser/Encoding.php
```diff
@@ -145,6 +145,12 @@ protected function getEncodingClass(): string
     {
         // Load reference table charset.
         $baseEncoding = preg_replace('/[^A-Z0-9]/is', '', $this->get('BaseEncoding')->getContent());
+
+        // Check for empty BaseEncoding field value
+        if ('' == $baseEncoding) {
```
k00ni, line 150: Based on https://www.php.net/manual/en/function.preg-replace.php, `preg_replace` can return `string`, `array` and `null`. Shouldn't we check for `[]` and `null` at this point too?

GreyWyvern: I could perhaps change the test to `if (empty($baseEncoding)) { ... }` ? How's that?

k00ni: I would change the whole part to:

```php
// Check for empty BaseEncoding field value
if (is_string($baseEncoding) && 0 < strlen($baseEncoding)) {
    $className = '\\Smalot\\PdfParser\\Encoding\\'.$baseEncoding;
} else {
    $baseEncoding = 'StandardEncoding';
}
```
Now we dont have to worry about null and empty arrays. I don't like `empty` personally, because it does too much magic and one may not be aware of its side "effects" (e.g. no warning if a variable doesn't exist). With this approach we are very explicit what we want and in all other cases we use `StandardEncoding`. 

However we do this code part, we must handle the null- and empty array case.

GreyWyvern: Since the `$className = ...` statement is right below this, it would be simpler to use:

```php
if (!is_string($baseEncoding) || 0 == strlen($baseEncoding)) {
  $baseEncoding = 'StandardEncoding';
}
```

k00ni: My bad, I forgot to adapt one line. In the following would be the whole function with my adaptions:

```php
protected function getEncodingClass(): string
{
    // Load reference table charset.
    $baseEncoding = preg_replace('/[^A-Z0-9]/is', '', $this->get('BaseEncoding')->getContent());

    // use base encoding class if a non-empty string is provided
    if (is_string($baseEncoding) && 0 < strlen($baseEncoding)) {
        $className = '\\Smalot\\PdfParser\\Encoding\\'.$baseEncoding;
    } else {
        $className = '\\Smalot\\PdfParser\\Encoding\\StandardEncoding';
    }

    if (!class_exists($className)) {
        throw new EncodingNotFoundException('Missing encoding data for: "'.$baseEncoding.'".');
    }

    return $className;
}
```

But your approach is fine too, mine was just a minor tweak. Feel free to use what you want.

src/Smalot/PdfParser/Encoding.php
```diff
@@ -145,6 +145,12 @@ protected function getEncodingClass(): string
     {
         // Load reference table charset.
         $baseEncoding = preg_replace('/[^A-Z0-9]/is', '', $this->get('BaseEncoding')->getContent());
+
+        // Check for empty BaseEncoding field value
+        if (!\is_string($baseEncoding) || 0 == \strlen($baseEncoding)) {
```
k00ni, line 150: Just to make sure: if `$baseEncoding` is `null` or an `[ ]`, only the first part of the if-clauses is being used and the second part is not used, correct? The reason why I am asking, `strlen` triggers a warning or type error, if it receives a non-string as parameter.

GreyWyvern: Yes. If `!is_string()` is `true`, the second part of the expression isn't tested since it isn't needed. The only way the `strlen()` function is reached is if the first expression evaluated to `false` which means it *must* be a string.

k00ni (approved):

k00ni:
Thank you!

Issue #668, opened 2024-01-22
preg_match(): compilation failed: regular expression is too large to offset 143690


lonelyrider44:
This happens with v2.8.0.
2.7.0 works fine

JohnMirro:
I get this issue too with v 2.8.0

k00ni:
> This happens with v2.8.0. 2.7.0 works fine

What? Where? Versions? I can't see where the problem is and why. Please provide information about your setup, a stack trace of the error (or at least file line) and a PDF/example string which triggers the exception.

GreyWyvern:
Likely the result of binary content slipping through to the `formatContent()` function. The regexp looks for balanced parentheses in document content (where balance is required), but a binary stream can loop into a huge futile regexp.

Any chance you could post the offending PDFs, @lonelyrider44 or @JohnMirro ?

JohnMirro:
@GreyWyvern sorry, I lost those PDF

cxammar:
Yup, exactly same issue here when uploading payrolls as pdf

GreyWyvern:
> [InvoicesMN0052-2320230303093445.pdf](https://github.com/smalot/pdfparser/files/14043822/InvoicesMN0052-2320230303093445.pdf)

Thanks for this!

So yes, this is a matter of binary content (an image, or something else) being passed to `formatContent()`. `formatContent()` does a check for binary content, but only *after* the problematic regexp. Unfortunately it's required to be this way because the check for binary content may interpret string content as binary, which is perfectly valid. This is the whole reason why strings are removed, then put back.

Either there needs to be a more thorough check for binary before `formatContent()` is called, or we need an alternate way to detect if a content stream is binary *before* the regexp is called.

One way may be to bulk remove *everything* from the content stream from the first parenthesis (start of the first supposed string) to the end of the stream, which would require only a minimal regexp. Everything left would *have* to be plain text abiding by a predictable regexp to show validity. If any binary bytes exist in what's left, the stream could safely be discarded.

If anyone has any other ideas, let me know.

GreyWyvern:
> One way may be to bulk remove _everything_ from the content stream from the first parenthesis (start of the first supposed string) to the end of the stream, which would require only a minimal regexp. Everything left would _have_ to be plain text abiding by a predictable regexp to show validity. If any binary bytes exist in what's left, the stream could safely be discarded.

Something like this, I think, would cover all the bases:

```php
[...]

PR #676, opened 2024-02-14
Check for binary content in formatContent() before a problematic regexp

GreyWyvern:
# Type of pull request

# About

Some PDF files manage to serve binary content (a font, or an image, etc.) to the `formatContent()` method which only accepts text document streams.

Since `(strings)` in PDF document streams can contain binary content (that gets decoded by a font library, for instance) we must test the document stream for text-only content EXCLUDING `(string)` content.

Previously the check for text-only content was done *after* a regexp removing the `(strings)` from the stream, recursively extending the regexp to check for balanced parentheses. However for a sufficiently long binary stream, this might create a regexp long enough to cause a PHP error.

The solution is to move the check for binary content *before* this `(string)`-removing regexp. Simplify the binary check by truncating the document stream at the first open parenthesis, which indicates the start of a `(string)`, then testing what remains for valid UTF-8.

This makes the later check for binary content unnecessary and it has been removed. Resolves issue #668.

Future work on this issue should be done to determine why PdfParser is creating PDFObject objects from binary data in the first place.


src/Smalot/PdfParser/PDFObject.php
```diff
@@ -214,6 +214,16 @@ private function formatContent(?string $content): string
             return '';
         }
 
+        // Outside of (String) content in PDF document streams, all
+        // text should conform to UTF-8. Test for binary content by
+        // deleting everything after the first open-parenthesis ( which
+        // indicates the beginning of a string. Then test what remains
+        // for valid UTF-8. If it's not UTF-8, return an empty string
+        // as this $content is most likely binary.
+        if (!mb_check_encoding(preg_replace('/\(.*$/s', '', $content), 'UTF-8')) {
```
k00ni, line 223:
```suggestion
        if (false === mb_check_encoding(preg_replace('/\(.*$/s', '', $content), 'UTF-8')) {
```

k00ni:
I fixed coding style issues in #677, which got merged into master branch.

PR #686, opened 2024-03-06
Prevent zero from being passed to array_chunk()

### LINKED ISSUE #679: Bug in RawDataParser.php when a row has no data
- PHP Version: 8.1
 - PDFParser Version: 2.8.0

### PDF input
The PDF has personal identifying information that I cannot share through a post like this. I could maybe email it directly if necessary.

I'm just trying to call 

### Code
```
                            case 'pdf':
                                try {
                                    $PDF = $parser->parseFile($filepath);
                                    $pdfText = mb_convert_encoding($PDF->getText(), 'UTF-8');
                                    $result .= $pdfText;
                                } catch (\Exception $e) {
                                    \Log::error('Error parsing recommendation PDF text: '.$e->getMessage());
                                    break;
                                }
```

I'm getting this error:

image

In \smalot\pdfpaser\src\Smalot\PdfParser\RawData\RawDataParser.php, the issue seems to be coming from this code:
```
            } else {
                // number of bytes in a row
                $rowlen = array_sum($wb);
                // convert the stream into an array of integers
                $sdata = unpack('C*', $xrefcrs[1][3][0]);
                // split the rows
                $ddata = array_chunk($sdata, $rowlen);
            }
```

I wonder if it should be something like this?
```
            } else {
                // number of bytes in a row
                $rowlen = array_sum($wb);
                if ($rowlen > 0) {
                    // convert the stream into an array of integers
                    $sdata = unpack('C*', $xrefcrs[1][3][0]);
                    // split the rows
                    $ddata = array_chunk($sdata, $rowlen);
                } else {
                    $dddata = [];
                }
            }
```

GreyWyvern:
# Type of pull request

# About

Passing a zero (0) value to the `array_chunk()` function causes an error, and in rare cases, a PDF XRef object may be added to a document with an "empty" `/W [0 0 0]` command. In **RawDataParser.php** this would cause the `$rowlen` variable to be set to zero and cause an error.

Add a simple check to return an empty array in this case. Resolves #679.

It is very difficult to create a unit test for this error as it requires a PDF to be generated with "empty" sections with the specific `/W [0 0 0]` command. As the error occurs at a point when the *entire* document is being considered, we can't just feed PdfParser some test PDF code that's a subsection of a full document. The sample PDF given by the reporter of issue #679 contained personal info that we cannot include in PdfParser and they did not know how to generate a similar file. It is hoped we can merge this PR without a unit test added, since it is, at it's core, just a check for a zero value which should have been in the code already. :)

Thanks to @KeanuTang for the initial analysis and provided code solution, with which I built this PR.

* [ ] Please add at least **one test case** (unit test, system test, ...) to demonstrate that the change is working. If existing code was changed, your tests cover these code parts as well.

### DIFF
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -408,10 +408,15 @@ class RawDataParser
             } else {
                 // number of bytes in a row
                 $rowlen = array_sum($wb);
-                // convert the stream into an array of integers
-                $sdata = unpack('C*', $xrefcrs[1][3][0]);
-                // split the rows
-                $ddata = array_chunk($sdata, $rowlen);
+                if (0 < $rowlen) {
+                    // convert the stream into an array of integers
+                    $sdata = unpack('C*', $xrefcrs[1][3][0]);
+                    // split the rows
+                    $ddata = array_chunk($sdata, $rowlen);
+                } else {
+                    // if the row length is zero, $ddata should be an empty array as well
+                    $ddata = [];
+                }
             }
 
             $sdata = [];


k00ni (approved):
:+1: @GreyWyvern

About the missing unit test: You are right, its very difficult to write a test for this code part. I checked current code coverage and this part is covered by two other tests indirectly (https://scrutinizer-ci.com/g/smalot/pdfparser/code-structure/master/code-coverage/src/Smalot/PdfParser/RawData/RawDataParser.php). So in this case its appropriate.

PR #692, opened 2024-03-14
Account for inaccurate offsets in getXrefData()

### LINKED ISSUE #673: Undefined array key 1,crash on parsing
- PHP Versiov 8.3: 
 - PDFParser Version 2.8: 

### PDF input

[Cracking-the-Coding-Interview-6th-Edition-189-Programming-Questions-and-Solutions.pdf](https://github.com/smalot/pdfparser/files/14177782/Cracking-the-Coding-Interview-6th-Edition-189-Programming-Questions-and-Solutions.pdf)

It crashes in the RawDataPraser, line 890 

`elseif ($startxrefPreg) {
            // startxref found
            $startxref = $matches[1][0];
        }`
$matches is empty array.
### Code

Just the usual stuff.

GreyWyvern:
# Type of pull request

# About

Normally offset pointers to `xref` keywords in a PDF document are exact to the byte. However, in some cases the pointer may point to some whitespace *before* the `xref` keyword. Adobe Acrobat takes these 'errors' in stride, displaying the document anyway, and so should PdfParser.

Clean up the getXrefData() function in **RawDataParser.php**. It now only needs to do one `preg_match_all()` and pushes the caret past whitespace when looking for `xref` keywords.

Use existing **Issue557.pdf** to create a new file: **Issue673.pdf** where the last `/Prev 13486` command has been decremented to `/Prev 13485`. Trying to parse this file would cause an Exception without this fix. Resolves #673.


src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -864,39 +864,39 @@ private function getHeaderValue(?array $headerDic, string $key, string $type, $d
...
+        // Find all startxref tables from this $offset forward
+        $startxrefPreg = preg_match_all(
+            '/(?<=[\r\n])startxref[\s]*[\r\n]+([0-9]+)[\s]*[\r\n]+%%EOF/i',
             $pdfData,
-            $matches,
-            \PREG_OFFSET_CAPTURE,
+            $startxrefMatches,
+            \PREG_SET_ORDER,
             $offset
         );
```
k00ni, line 883: Using `$startxrefMatches` as third parameter and to store result of `preg_match_all` looks weird. It seems to work, but I am wondering which one of these two is stored. Based on your code you assume its the latter. But you also use it as it contains the matches (e.g. line 890).

I would like to use two separate variables here to improve code readability.

src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -864,39 +864,39 @@ private function getHeaderValue(?array $headerDic, string $key, string $type, $d
...
             // Cross-Reference Stream object
-            $startxref = $offset;
-        } elseif ($startxrefPreg) {
-            // startxref found
-            $startxref = $matches[1][0];
+            $startxref = $bumpOffset;
         } else {
-            throw new \Exception('Unable to find startxref');
+            // Use the next startxref from this $offset
+            $startxref = (int) $startxrefMatches[0][1];
```
k00ni, line 899: Can we be sure that `$startxrefMatches[0][1]` is always defined? If not, we could run into an index not defined error.

k00ni (changes requested):
Thank you @GreyWyvern, good work! I have just a few remarks/questions.

src/Smalot/PdfParser/RawData/RawDataParser.php
GreyWyvern: On line 885, there is a non-strict check of the `preg_match_all()` return value `0 == $startxrefPreg` which will catch if there were no matches, or if there was a failure `false`. So to arrive at the final `else` here, `$startxrefMatches` MUST have at least one match in it.

src/Smalot/PdfParser/RawData/RawDataParser.php
GreyWyvern: `$startxrefPreg` is the result of the `preg_match_all()` and `$startxrefMatches` stores the associated matches. So they are already two different variables? Unless I'm reading your comment incorrectly...

k00ni: My bad, all good.

k00ni (approved):
Sorry for the delay. I merged in a recent change to get rid of the coding style issue.

PR #693, opened 2024-03-15
Account for inline images in formatContent()

### LINKED ISSUE #691: Trying to access array offset on value of type null (PDFObject.php line 795)
- PHP Version: 8.3.3
 - PDFParser Version: 2.9.0

### Description:

An exception `Trying to access array offset on value of type null` was thrown on `PDFObject.php` line `795`
`$current_position_cm is null`
image
image

### PDF input

Not allowed to put pdf in public, but can share it privately.

Expected output: to get text
Actural output: Exception was thrown

### Code

```php
            $pdf = $parser->parseFile($pathToPDF);
            $texts = $pdf->getText();
```

GreyWyvern:
# Type of pull request

# About

`formatContent()` now accounts for inline image `BI ... ID ... EI` commands in document streams. Resolves #691.

### DIFF
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -233,31 +233,32 @@ class PDFObject
             // actually occured within a (string) using the following
             // steps:
 
-            // Remove any escaped parentheses from the alleged image
-            // characteristics data
+            // Step 1: Remove any escaped parentheses from the alleged
+            // image characteristics data
             $para = str_replace(['\\(', '\\)'], '', $text[1][0]);
 
-            // Remove all correctly ordered and balanced parentheses
-            // from (strings)
+            // Step 2: Remove all correctly ordered and balanced
+            // parentheses from (strings)
             do {
                 $paraTest = $para;
-                $para = preg_replace('/\(([^)]*)\)/', '$1', $paraTest);
+                $para = preg_replace('/\(([^()]*)\)/', '$1', $paraTest);
             } while ($para != $paraTest);
 
             $paraOpen = strpos($para, '(');
             $paraClose = strpos($para, ')');
 
-            // If the remaining text contains a close parenthesis ')'
-            // AND it occurs before any open parenthesis, then we are
-            // almost certain to be inside a (string)
+            // Check: If the remaining text contains a close parenthesis
+            // ')' AND it occurs before any open parenthesis, then we
+            // are almost certain to be inside a (string)
             if (0 < $paraClose && (false === $paraOpen || $paraClose < $paraOpen)) {
                 // Bump the search offset forward and match again
                 $offsetBI = (int) $text[1][1];
                 continue;
             }
 
-            // Double check that this is actually inline image data by
-            // parsing the alleged image characteristics as a dictionary
+            // Step 3: Double check that this is actually inline image
+            // data by parsing the alleged image characteristics as a
+            // dictionary
             $dict = $this->parseDictionary('<<'.$text[1][0].'>>');
 
             // Check if an image Width and Height are set in the dict
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -274,6 +274,12 @@ class PDFObject
                     $content,
                     1
                 );
+            } else {
+                // If there was no valid dictionary, or a height and width
+                // weren't specified, then we don't know what this is, so
+                // just leave it alone; bump the search offset forward and
+                // match again
+                $offsetBI = (int) $text[1][1];
             }
         }
 
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -214,6 +214,20 @@ class PDFObject
             return '';
         }
 
+        // Find all inline image content and replace them so they aren't
+        // affected by the next steps
+        $pdfInlineImages = [];
+        while (preg_match('/\sID\s(.+?)\sEI(?=\s|$)/', $content, $text)) {
+            $id = uniqid('IMAGE_', true);
+            $pdfInlineImages[$id] = $text[1];
+            $content = preg_replace(
+                '/'.preg_quote($text[0], '/').'/',
+                '^^^'.$id.'^^^',
+                $content,
+                1
+            );
+        }
+
         // Outside of (String) content in PDF document streams, all
         // text should conform to UTF-8. Test for binary content by
         // deleting everything after the first open-parenthesis ( which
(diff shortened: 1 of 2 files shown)


tests/PHPUnit/Integration/PDFObjectTest.php
```diff
@@ -284,6 +284,19 @@ public function testFormatContent(): void
...
+        $cleaned = $formatContent->invoke(
+            $this->getPdfObjectInstance(new Document()),
+            'q 65.30 0 0 18.00 412 707 cm BI /W 544 /H 150 /BPC 1 /IM true /F [/A85 /Fl] ID Gb"0F_$L6!$j/a\$:ma&h\'JnJJ9S?O_EA-W+%D^ClCH=FP3s5M-gStQm\'5/hc`C?<Q)riWgtEe:Po0dY_-er6$jM@#?n`E+#(sa"0Gk3&K>CqL(^pV$_-er6Ik`"-1]Q ;~> EI Q /F002 10.00 Tf 0.00 Tw 0 g'
+        );
+
+        // PdfParser should not be fooled by Q's in inline image data;
+        // Only one 'Q' command should be found
+        $commandQ = preg_match_all('/Q\r\n/', $cleaned);
+
+        $this->assertEquals(1, $commandQ);
```
k00ni, line 299: Please move it to a separate test case to make the test case easier to read.

k00ni (changes requested):
@GreyWyvern :+1:

k00ni:
I really appreciate you taking the time!

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -224,6 +224,59 @@ private function formatContent(?string $content): string
...
+                ];
+                $content = preg_replace(
+                    '/'.preg_quote($text[0][0], '/').'/',
+                    '^^^'.$id.'^^^',
+                    $content,
+                    1
+                );
+            }
+        }
+
```
k00ni, line 279: We should move this part to a separate function? It will be easier to test later on. Maybe the "restore" part too.

k00ni (changes requested):

src/Smalot/PdfParser/PDFObject.php
k00ni: I understand. Please continue as you see fit for now. In the end code readability could might be improved if we move parts like that in a separate, private function in PDFObject, or not. Will see.

GreyWyvern:
So, the last thing left here that the code wouldn't cover is a *proper* inline image, that *doesn't* have a proper image-properties dictionary with a width and height. The code in this PR then skips over it, but the potential is there for such an inline image (probably *very* rare if it happens at all) to contain binary content that can potentially cause errors in the way PdfParser interprets the document stream. (Like unbalanced Q/q etc.)

We can:
- Just accept it as is; the document with such an inline image is malformed anyways. There should be no expectation of an error-free parsing in such a case.
- Not check for the height and width in the dictionary at all, and just accept all `BI ... ID ... EI` sequences outside of strings as "valid" inline images. This allows the possibility (miniscule?) of finding false-positive inline image sequences.

I've no data to back it up, but I believe the second case, where `formatContent()` finds a `BI ... ID ... EI` sequence outside of a string, but **in error**, is a probably rarer than an inline image dictionary not containing a height and width. But then again I could be wrong!

Regardless, I would recommend keeping the dictionary check just in case. If it gets released and users find the array-access error again, then we can always remove it. In this case, this PR is ready to be taken out of draft status as-is.

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -224,6 +224,65 @@ private function formatContent(?string $content): string
...
+            // Remove any escaped parentheses from the alleged image
+            // characteristics data
+            $para = str_replace(['\\(', '\\)'], '', $text[1][0]);
+
+            // Remove all correctly ordered and balanced parentheses
+            // from (strings)
+            do {
+                $paraTest = $para;
+                $para = preg_replace('/\(([^)]*)\)/', '$1', $paraTest);
+            } while ($para != $paraTest);
```
k00ni, line 245: Can we be sure this never can become an endless loop?

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -224,6 +224,65 @@ private function formatContent(?string $content): string
...
+                    '^^^'.$id.'^^^',
+                    $content,
+                    1
+                );
+            } else {
+                // If there was no valid dictionary, or a height and width
+                // weren't specified, then we don't know what this is, so
+                // just leave it alone; bump the search offset forward and
+                // match again
+                $offsetBI = (int) $text[1][1];
```
k00ni, line 282: Maybe something like the following would prevent array-access errors (to somet extent)?

```suggestion
                $offsetBI = (int) $text[1][1] ?? 1;
```

The `?? 1` is some dummy value, it should be replaced with a default value which makes sense, but only if there is a chance that `$text[1][1]` can be undefined.

src/Smalot/PdfParser/PDFObject.php
```diff
@@ -224,6 +224,65 @@ private function formatContent(?string $content): string
...
+        while (preg_match('/\sBI\s(\/.+?)\sID\s(.+?)\sEI(?=\s|$)/s', $content, $text, \PREG_OFFSET_CAPTURE, $offsetBI)) {
+            // Attempt to detemine if this instance of the 'BI' command
+            // actually occured within a (string) using the following
+            // steps:
+
+            // Remove any escaped parentheses from the alleged image
+            // characteristics data
+            $para = str_replace(['\\(', '\\)'], '', $text[1][0]);
+
+            // Remove all correctly ordered and balanced parentheses
```
k00ni, line 240: I am only nitpicking here: could you add a numbering to the steps you refering here? I assume the comment in line 236 is for step 1 and comment in line 240 for step 2?

k00ni (commented):
> Regardless, I would recommend keeping the dictionary check just in case. If it gets released and users find the array-access error again, then we can always remove it. In this case, this PR is ready to be taken out of draft status as-is.

In my opinion, if something is not according to the specification, you can go yolo and do whatever you want. As you described in high detail, the only thing we can do with ill-formed PDFs is to try to make the best of it. As a user/developer I surely appreciate if software can handle ill-formed data to some extent. It keeps me sane. On the other hand we are a community which maintains the library in our sparetime, so there must be a balance.

That being said, your arguments make sense and I will follow your advice here @GreyWyvern. Please do the final preparations and mark the PR ready for review.

In the following just a few remarks/suggestions.

src/Smalot/PdfParser/PDFObject.php
GreyWyvern: The `preg_replace()` call is removing parentheses in pairs. Whenever it succeeds, `$para` is updated so it does not equal `$paraTest`. When it fails (does not find anything to replace), then `$para` remains the same and the loop exits. I can't see any way this falls into an endless loop.

src/Smalot/PdfParser/PDFObject.php
GreyWyvern: I think this is a decent idea in theory, but it couldn't be implemented as simply as this.

First of all I don't think there is a case where asking for `preg_match` to return offset values returns successful matches, but *without* the offset values. So is it even necessary?

Second, if you wanted to add it, you couldn't just use a static value because of the potential for falling into an endless loop. The next iteration of the loop would find the same match again, have the offset set to `1` and find the same match again, etc. So if the offset value was somehow missing, you would have to, say, take the first 10 or 20 characters of the match and match it against the document stream to find the offset, then use *that* as the fallback.

I believe it's probably too much trouble to account for, for something that perhaps might never happen.

k00ni:
Thank you very much @GreyWyvern

PR #700, opened 2024-04-05
fix: Return page width and height from document

vitormattos:
# Type of pull request

* [x] Documentation update
* [ ] Something else

# About

Some documents could have pages without MediaBox, at this case we need to get the MediaBox from document and not from a specific page.

### DIFF
--- a/doc/Usage.md
+++ b/doc/Usage.md
@@ -226,7 +226,8 @@ foreach ($pages as $page) {
     $details = $page->getDetails();
     // The page could haven't the Mediabox, then we get this information from document
     if (!isset($details['MediaBox'])) {
-        $details = reset($pdf->getObjectsByType('Pages'))->getHeader()->getDetails();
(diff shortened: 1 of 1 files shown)


k00ni:
Thank you, good addition to the documentation @vitormattos

doc/Usage.md
```diff
@@ -224,6 +224,11 @@ $pages = $pdf->getPages();
 $mediaBox = [];
 foreach ($pages as $page) {
     $details = $page->getDetails();
+    // The page could haven't the Mediabox, then we get this information from document
```
k00ni, line 227: The comment doesn't reflect the code in my opinion. Here is an alternative:

```suggestion
    // If Mediabox is not set in details of current $page instance, get details from the header instead
```

vitormattos: Thanks! English is not my main language.

k00ni (approved):
:+1:

PR #704, opened 2024-04-22
Strengthen check for UTF-8 conformity in formatContent()

### LINKED ISSUE #668: preg_match(): compilation failed: regular expression is too large to offset 143690

GreyWyvern:
# Type of pull request

# About

In some cases a binary string may pass as valid UTF-8 to the `mb_check_encoding(..., 'UTF-8')` function. Use a comprehensive regexp from the W3 group instead to be **certain** we aren't trying to parse binary content in `formatContent()`. In addition to `(strings)`, also check for the beginning of `ID` inline image content sections, which may also contain binary. Resolves #668.

Reference: https://www.w3.org/International/questions/qa-forms-utf-8.en

### DIFF
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -214,13 +214,31 @@ class PDFObject
             return '';
         }
 
-        // Outside of (String) content in PDF document streams, all
-        // text should conform to UTF-8. Test for binary content by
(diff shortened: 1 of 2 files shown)


k00ni (approved):
@GreyWyvern I appreciate you took the time to improve code readability, by extending the comments and moving the `preg_replace` call out of the if-clause. Thanks!

PR #711, opened 2024-05-13
Fix for adjacent escaped slashes and escaped parentheses in strings

### LINKED ISSUE #709: preg_match(): Compilation failed: regular expression is too large at offset 38605
pdfparser version:2.10.0

I have about 600,000 pdf files, all of which use pdfparser for text extraction.This kind of problem was shown to have been solved on 704, probably because of the Chinese coding problem, and now it appears again, please help to solve it, thank you.
[1710747436_65f7ef2ccfac97cc01c0803eda73f23a732316a07e2ab5f2c43ec0e162ac4.pdf](https://github.com/smalot/pdfparser/files/15244269/1710747436_65f7ef2ccfac97cc01c0803eda73f23a732316a07e2ab5f2c43ec0e162ac4.pdf)
[1710812373_65f8ecd57a825b030437578010bf1e1aa0ae31669f11f5fae857a58f22bbf.pdf](https://github.com/smalot/pdfparser/files/15244270/1710812373_65f8ecd57a825b030437578010bf1e1aa0ae31669f11f5fae857a58f22bbf.pdf)
[1710898905_65fa3ed9d61a7e6d996a0d361fbfe8efa9620742d2e0560f19c51edefd6f2.pdf](https://github.com/smalot/pdfparser/files/15244271/1710898905_65fa3ed9d61a7e6d996a0d361fbfe8efa9620742d2e0560f19c51edefd6f2.pdf)

[1710124871_65ee6f473b7eddbd7437d2803d8f95dff1747721dc88d4035d46343d9766e.pdf](https://github.com/smalot/pdfparser/files/15244268/1710124871_65ee6f473b7eddbd7437d2803d8f95dff1747721dc88d4035d46343d9766e.pdf)

GreyWyvern:
# Type of pull request

# About

The current `(string)` replacement regexp in `formatContent()` only backchecked two characters for escaped slashes, so if an escaped slash immediately preceded an escaped parenthesis, the script would incorrectly interpret it as an escaped slash and an unescaped parenthesis. This would lead to the loop never finding the "end" of the string (for an open parenthesis) or finding the end of the string prematurely (for a close parenthesis).

Perform a string replace to get rid of all escaped slashes and then escaped parentheses; they aren't needed when just checking for balanced, unescaped parentheses. Also add removing slashes to the inline images section above for the same reason.

Resolves #709.

### DIFF
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -251,9 +251,9 @@ class PDFObject
             // actually occured within a (string) using the following
             // steps:
 
-            // Step 1: Remove any escaped parentheses from the alleged
-            // image characteristics data
-            $para = str_replace(['\\(', '\\)'], '', $text[1][0]);
+            // Step 1: Remove any escaped slashes and parentheses from
+            // the alleged image characteristics data
+            $para = str_replace(['\\\\', '\\(', '\\)'], '', $text[1][0]);
 
             // Step 2: Remove all correctly ordered and balanced
             // parentheses from (strings)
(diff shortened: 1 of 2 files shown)


k00ni:
@huihuangjiuai Does this fix #709 for you?

k00ni (approved):
:+1: Short and clean. 

I only moved the test code into a separate function to keep code readability up, otherwise test code concerning different issues is mixed, which makes debugging/maintenance more difficult in the future.

PR #716, opened 2024-05-27
Fix hexadecimal decoding (fixes #715)

### LINKED ISSUE #715: Closing round bracket encoded in hexadecimal format breaks parsing
- PHP Version: 8.2.10
 - PDFParser Version: master (4b86c6636d086ca7ea4780c07c2d7390321982b5)

### Description:
Closing round bracket encoded in hexadecimal format breaks parsing - string is truncated.
String is truncated here:
https://github.com/smalot/pdfparser/blob/4b86c6636d086ca7ea4780c07c2d7390321982b5/src/Smalot/PdfParser/Element/ElementString.php#L62-L74
Because this is my first contact with pdfparser I probably have no competence to provide a safe patch.

### Test
```
public function testHexadecimalEncodedBracket(): void
{
    $document = new Document();

    $testString = '()';
    $content = '<< /Contents <'.bin2hex($testString).'> >>';
    $header = Header::parse($content, $document);
    $this->assertEquals($testString, (string) $header->get('Contents'));
}
```

Test should pass, but returns:
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'()'
+'('

krzyc:
# Type of pull request

# About

Fix for #715 as requested by @k00ni. Tests are passing but I am not confident that this will not break pdfparser because I lack knowledge of internal work of library and PDF format.

### DIFF
--- a/src/Smalot/PdfParser/Element/ElementHexa.php
+++ b/src/Smalot/PdfParser/Element/ElementHexa.php
@@ -47,12 +47,11 @@ class ElementHexa extends ElementString
         if (preg_match('/^\s*\<(?P<name>[A-F0-9]+)\>/is', $content, $match)) {
             $name = $match['name'];
             $offset += strpos($content, '<'.$name) + \strlen($name) + 2; // 1 for '>'
-            // repackage string as standard
-            $name = '('.self::decode($name).')';
-            $element = ElementDate::parse($name, $document);
+            $name = self::decode($name);
+            $element = ElementDate::parse('('.$name.')', $document);
 
             if (!$element) {
-                $element = ElementString::parse($name, $document);
+                $element = new self($name);
             }
 
             return $element;
--- a/src/Smalot/PdfParser/Element/ElementString.php
+++ b/src/Smalot/PdfParser/Element/ElementString.php
@@ -52,6 +52,51 @@ class ElementString extends Element
     }
 
     /**
+     * Part of parsing process to handle escaped characters.
+     * Note, most parameters are passed by reference.
+     *
+     * Further information in PDF specification (page 53):
+     * https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.7old.pdf
+     */
+    private static function handleEscapedCharacters(string &$name, int &$position, string &$processedName, string $char): void
+    {
+        // escaped chars
+        $nextChar = substr($name, 0, 1);
+        switch ($nextChar) {
+            // end-of-line markers (CR, LF, CRLF) should be ignored
+            case "\r":
+            case "\n":
+                preg_match('/^\\r?\\n?/', $name, $matches);
+                $name = substr($name, \strlen($matches[0]));
+                $position += \strlen($matches[0]);
+                break;
+                // process LF, CR, HT, BS, FF
+            case 'n':
+            case 't':
+            case 'r':
+            case 'b':
+            case 'f':
+                $processedName .= stripcslashes('\\'.$nextChar);
+                $name = substr($name, 1);
+                ++$position;
+                break;
+                // decode escaped parentheses and backslash
+            case '(':
+            case ')':
+            case '\\':
+            case ' ': // TODO: this should probably be removed - kept for compatibility
+                $processedName .= $nextChar;
+                $name = substr($name, 1);
+                ++$position;
+                break;
+                // TODO: process octal encoding (but it is also processed later)
+                // keep backslash in other cases
+            default:
+                $processedName .= $char;
+        }
+    }
+
+    /**
      * @return bool|ElementString
      */
     public static function parse(string $content, ?Document $document = null, int &$offset = 0)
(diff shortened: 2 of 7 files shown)


krzyc:
@k00ni I have fixed PHP 7.x incompatibility.
Regarding testSpecialCharsEncodedAsHex test failed - I have still not found a method to fix it - I have added this test because during additional inspection I have discovered that there is more related problems with round brackets.
But I think that PR is already in state that allows us to verify whether my approach is correct.

k00ni:
Is there a reference in the PDF specification you referring to with this changes in general?

krzyc:
I have rewritten element string parser and now tests pass.
I have two observations:
- backslash followed by space should return backslash and space, but this triggers fail of tests, so I kept old behavior (returning space only) for compatibility - to be discussed.
- octal parsing should be probably moved to parsing loop, but I kept current behavior - to be discussed.
I am awaiting your feedback.

k00ni:
Thanks for your effort! I try to give you feedback as soon as possible.

src/Smalot/PdfParser/Element/ElementString.php
```diff
@@ -51,6 +51,51 @@ public function equals($value): bool
...
+                break;
+                // decode escaped parentheses and backslash
+            case '(':
+            case ')':
+            case '\\':
+            case ' ': // TODO: this should probably be removed - kept for compatibility
+                $processedName .= $nextChar;
+                $name = substr($name, 1);
+                ++$position;
+                break;
```
k00ni, line 91: What do you mean with "kept for compatibility"? To what?

src/Smalot/PdfParser/Element/ElementHexa.php
```diff
@@ -47,12 +47,11 @@ public static function parse(string $content, ?Document $document = null, int &$
...
-            // repackage string as standard
-            $name = '('.self::decode($name).')';
-            $element = ElementDate::parse($name, $document);
+            $name = self::decode($name);
+            $element = ElementDate::parse('('.$name.')', $document);
 
             if (!$element) {
-                $element = ElementString::parse($name, $document);
+                $element = new self($name);
             }
```
k00ni, line 55: Please give me a short summary what you try to accomplish here.

Why removing the reference to `$document`?

k00ni (commented):
@krzyc Again, thanks for your effort. I dove into your code and the PDF specification 1.7 and holy cow, the spec has over 1300 pages. This topic is not easy but I think I got what you try to achieve with your patch as well as what the spec says in this regard. But further feedback is welcomed as always, CC: @j0k3r @GreyWyvern.

---

I took the liberty to change some of you code. Here is a summary of my changes:
* refactored some code parts
* added references and some comments
* added a few tests

(for further details, please see my commit notes)

**About the tests:** I copied a few example texts from the spec (page 54 + 55) to see if PDFParser fails without your fixes. It does most of the time. Only [this one](https://github.com/smalot/pdfparser/pull/716/files#diff-3b58988938ced1f0b6c99c9f181bec3ef4a461cd4b4834bf5be3d316d249196eR152) works pre- and post patch.

What is great is the fact that it now parses strings like ([ref](https://github.com/smalot/pdfparser/pull/716/files#diff-3b58988938ced1f0b6c99c9f181bec3ef4a461cd4b4834bf5be3d316d249196eR146-R147)):

> (Strings may contain balanced parentheses ( ) and
> special characters (*!&}^% and so on).)

Pre-patch PDFParser would have only got until `Strings may contain balanced parentheses (`.

I can image there are further edge cases which still might not work. But we should aim for the following: It can get new features/capabilities as long as it retains its current ones.

To your knowledge, are there any cases left which worked before but don't now?

src/Smalot/PdfParser/Element/ElementHexa.php
j0k3r: Also why switching from `ElementString` to `ElementHexa`? Is it to match others existing behavior from other elements classes?

GreyWyvern:
After an initial review, this appears to be an issue solely with ElementString::parse. There are zero changes required to the **ElementHexa.php** file IMHO. ElementHexa::parse sends ElementString::parse a perfectly valid string, and it's ElementString that proceeds to mess it up.

The only reason ElementHexa is involved at all is because it's a vector that _allows_ this sort of tricky (but still valid!) string to reach ElementString, whereas normally ElementString would never see such strings.

ElementString::parse expects all internal parentheses to be escaped, and I would wager that in >99% of PDFs, they are. But according to the PDF Reference, balanced, unescaped parentheses are allowed. So this is a matter of _just_ updating ElementString::parse to handle balanced, unescaped parentheses. I haven't fully examined the updated code in ElementString::parse in this PR, but as long as it does that, there should be no other edits necessary to **ElementHexa.php**.

k00ni:
What are your plans here @krzyc?

As mentioned, the change in ElementHexa does not seem to be required.

GreyWyvern:
I've been thinking about this some more and it's a little more complicated than it seems at first glance.

ElementString accepts (and currently expects) strings with escaped slashes and parentheses. 99% of the time strings from PDF document streams will be escaped, **BUT** unescaped, balanced parentheses are allowed. Unescaped backslashes are NOT allowed; according to the PDF reference if they aren't followed by a valid escape character, they should be ignored.

And that is the problem with strings from ElementHexa. It passes a completely unescaped (but valid) string, but if it contains literal backslashes, then further parsing it will remove them or result in an incorrect special character.

Strings that originate from ElementHexa should be passed to ElementString to create the appropriate object, but `ElementString::parse()` should probably not be called since the string is already unescaped. Or if necessary, a new function to process already unescaped strings can be added.

k00ni:
I don't want to let this PR become stale, but my time is limited currently. How can we salvage this? Open for suggestions.

CC @GreyWyvern @j0k3r

k00ni:
Closed due to inactivity and no response from the author.

PR #719, opened 2024-06-05
Continuous-integration.yml: let workflow run on each push event

k00ni:
# Type of pull request

* [x] Something else

# About

Our current CI workflow only runs when a commit was pushed to the master branch or in a PR. This change lets it run every time something was committed, regardless which branch was used. It also allows us to create side branches and test things online, eliminating the need to do that locally. Before the change, CI workflow didn't trigger when a commit was added to a side-branch.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -3,8 +3,6 @@ name: "CI"
 on:
   pull_request:
   push:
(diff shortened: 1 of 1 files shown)


j0k3r (approved):

k00ni:
@j0k3r Thanks for the quick review.

PR #722, opened 2024-06-24
Merge XMP Metadata if dc:format tag not found

### LINKED ISSUE #721: Title and other properties not read with getDetails for some files
- PHP Version: 8.1
 - PDFParser Version: 2.10.0

### Description:
I recently used pdfparser to load 2500 documents in to Drupal. I pulled the properties with getDetails() to populate the Drupal fields.

About 50 documents did not get the right title and other properties, even though I can see them when I look at the files with a text editor and every other PDF tool I used saw them (eCopy, Chrome, Linux document viewer, Funnelback search engine).

I've attached a zip with some files that failed, and a few that succeeded, in case that helps track this down.

### PDF input

Attached.

Input is in attached file. Output expected to include all properties, but many are missing.

I have never seen an error of any kind.

### Code

```
public function getMetadataFromFile($file) {
  $metadata = "";
  $parser = new Parser();

  $pdf = $parser->parseFile($file);
  $metadata = $pdf->getDetails();

  return $metadata;
}
```

I have as similar function that uses parseContent and the problem is there, too. (I used parseConent to import existing files from an old system, and parseFile for new uploads.)
[Documents.zip](https://github.com/user-attachments/files/15890957/Documents.zip)

GreyWyvern:
# Type of pull request

# About

Previously `extractXMPMetadata()` would check for the existence of a `dc:format` tag with an `application/pdf` MIME-type value before allowing found XMP metadata to be merged with the other document details.

If the tag doesn't exist, merge the metadata anyway. If it _does_ exist, only _then_ check to see if it has the `application/pdf` MIME-type. Resolves #721.

### DIFF
--- a/src/Smalot/PdfParser/Document.php
+++ b/src/Smalot/PdfParser/Document.php
@@ -287,7 +287,7 @@ class Document
             }
 
             // Only use this metadata if it's referring to a PDF
-            if (isset($metadata['dc:format']) && 'application/pdf' == $metadata['dc:format']) {
+            if (!isset($metadata['dc:format']) || 'application/pdf' == $metadata['dc:format']) {
                 // According to the XMP specifications: 'Conflict resolution
                 // for separate packets that describe the same resource is
                 // beyond the scope of this document.' - Section 6.1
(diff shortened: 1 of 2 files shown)


k00ni (approved):
:+1: Good job

_Just for future reference, here is the related specification document: https://github.com/adobe/XMP-Toolkit-SDK/blob/main/docs/XMPSpecificationPart1.pdf_

PR #723, opened 2024-06-26
CI: added PHP 8.4

k00ni:
# Type of pull request

* [x] New feature (involves code and configuration changes)

# About

~~**PHP 8.4** is still in alpha so there is no need to hurry. I wanted to have it integrated already so we can observe changes and how PDFParser behaves.~~


j0k3r (approved):
Good idea 👍🏼

k00ni:
Github got some hiccups... Saw these errors when re-run the tests. First it seems they all failed but in the end almost all went through.

Bildschirmfoto vom 2024-09-25 16-48-49

k00ni:
**Summary:**
- Nothing spectacular anymore
- One adaption required to avoid a warning for `array_flip` ([FYI](https://github.com/smalot/pdfparser/pull/723/files#diff-27bc594b24bd5e2779e8d81ee79810d0ffda03f200f0d53be007a44a9d2cb2de))
- PHPUnit complains about some invalid attributes (*I don't really care about that, because its neither PDFParser nor quality related and they changed the config again and again from version to version. We have to use 3 versions for now to cover PHP 7.2 until 8.4, so its normal that we stumble upon deprecations*)

**In the following the output of PHPUnit when running on PHP 8.4 RC2:**

```
Runtime:       PHP 8.4.0RC2
Configuration: /var/www/html/phpunit.xml

...............................................................  63 / 187 ( 33%)
............................................................... 126 / 187 ( 67%)
.............................................................   187 / 187 (100%)

Time: 00:14.655, Memory: 605.75 MB

There was 1 PHPUnit test runner warning:

1) Test results may not be as expected because the XML configuration file did not pass validation:

  Line 5:
  - Element 'include': This element is not expected.
 ```

CC @j0k3r Any objection or can we merge it?

PR #727, opened 2024-07-25
fix: check that the previous xref is not the just processed xref

tkegan:
# Type of pull request

# About

A customer uploaded several files to my employer's product which cause one of our backend processes using smalot/pdfparser (thanks) to bomb with a fatal out of memory error.  I attempted to employ the [configuration](https://github.com/smalot/pdfparser/blob/master/doc/CustomConfig.md#option-setdecodememorylimit--setretainimagecontent-manage-memory-usage) documented. This did not work. Based on metadata this PDF is a scan made on some copier, pdf version 1.7. I did a bit of tracing the problem and found that memory was exhausted in an infinite loop. It seems the copier adds `Prev 0` in the trailing portion of a xref. The proposed change checks for this unique case where previous is the just processed xref and ignores it. This works for my employer's product.

     By the way, you don't have to provide a full fledged PDF file to demonstrate a fix. Instead a unit test may be sufficient sometimes,
     please have a look at [FontTest](https://github.com/smalot/pdfparser/blob/master/tests/PHPUnit/Unit/FontTest.php#L40) for example code.


tests/PHPUnit/Integration/RawData/RawDataParserTest.php
```diff
@@ -194,4 +194,24 @@ public function testGetXrefDataIssue673(): void
 
         self::assertStringContainsString('6 rue des Goutais', $text);
     }
+
+    /**
+     * Handle self referencing xref
+     *
+     * It seems that some pdf creators output `Prev 0` when there is no previous
+     * xref
```
k00ni, line 202:
```suggestion
     * It seems that some PDF creators output `Prev 0` when there is no previous xref.
```

k00ni (commented):
Thank you for the PR. This really sound like an edge case. Did you experience anything like that before @j0k3r @GreyWyvern?

At first glance the changes look good. I ran the code locally and it fails without the fix.

k00ni:
Thank you @tkegan

PR #737, opened 2024-09-26
Simplified Coding Style checks: PSR12 replaces Symfony, risky not allowed anymore

k00ni:
# Type of pull request

* [x] Something else

# About

Coding style checks are a pain in the butt for a while and are interfering with new pull requests frequently. Recent changes even lead to syntax errors because they are not compatible with older PHP versions ([example](https://github.com/smalot/pdfparser/actions/runs/11046573037/job/30686157896?pr=723)). Investing the time to keep the code up to date with them isn't worth it (at least to me). With this PR a stable and not-ever-changing coding style set is used (`PSR12`). Its also a standard and well documented (https://www.php-fig.org/psr/psr-12/). I hope this eliminates or at least reduces the number of coding style errors in the future while at the same time guarantees some level of consistent coding styles.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -14,10 +14,8 @@ $finder = Finder::create()
 $config = new Config();
 $config
     ->setFinder($finder)
-    ->setRiskyAllowed(true)
     ->setRules([
-        '@Symfony' => true,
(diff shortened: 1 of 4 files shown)


j0k3r (approved):
Good idea!

Issue #761, opened 2025-02-06
getDataTm is not returning all the text


ridgey-dev:
- PHP Version: 8.2.27
 - PDFParser Version: 2.11.0

### PDF input
[pdf-with-text.pdf](https://github.com/user-attachments/files/18689208/pdf-with-text.pdf)

### Expected output & actual output
The PDF contains the word `text`,  but the `getDataTm` does not return this text for the second page.
`[
  0 => array:2 [
    0 => array:6 [
      0 => "1"
      1 => "0"
      2 => "0"
      3 => "1"
      4 => "36.266"
      5 => "754.031"
    ]
    1 => "Test"
  ]
  1 => array:2 [
    0 => array:6 [
      0 => "1"
      1 => "0"
      2 => "0"
      3 => "1"
      4 => "34.016"
      5 => "701.653"
    ]
    1 => ""
  ]
]`

!Note that: The `getTextArray` **is** returning `text`. The problem has to do something with `getTextArray` returning an empty string for the second page.

`[
  0 => "Test"
  1 => ""
  2 => "text"
]`

These are the data commands:

`[
  0 => array:3 [
    "t" => ""
    "o" => "BT"
    "c" => ""
  ]
  1 => array:3 [
    "t" => ""
    "o" => "Td"
    "c" => "36.266 754.031"
  ]
  2 => array:3 [
    "t" => "/"
    "o" => "Tf"
    "c" => "F2 18.0"
  ]
  3 => array:3 [
    "t" => "["
    "o" => "TJ"
    "c" => array:1 [
      0 => array:3 [
        "t" => "("
        "o" => "TJ"
        "c" => "Test"
      ]
    ]
  ]
  4 => array:3 [
    "t" => ""
    "o" => "ET"
    "c" => ""
  ]
  5 => array:3 [
    "t" => ""
    "o" => "q"
    "c" => ""
  ]
  6 => array:3 [
    "t" => ""
    "o" => "cm"
    "c" => "187.717 0 0 90.000 371.297 715.624"
  ]
  7 => array:3 [
    "t" => ""
    "o" => "Q"
    "c" => ""
  ]
  8 => array:3 [
    "t" => ""
    "o" => "BT"
    "c" => ""
  ]
  9 => array:3 [
    "t" => ""
    "o" => "Td"
    "c" => "34.016 701.653"
  ]
  10 => array:3 [
    "t" => "/"
    "o" => "Tf"
    "c" => "F1 12.0"
  ]
[...]

PR #762, opened 2025-02-12
getTextArray: Add trim check in Do command

ridgey-dev:
# Type of pull request

* [ ] Documentation update
* [ ] Something else

# About
See: https://github.com/smalot/pdfparser/issues/761
Summary: If a document contains a Do command with an empty string in the content, if returns not corresponding text with the Text Matrix that is returned by using `getDataTm`.


ridgey-dev:
@k00ni What is the progress of this PR? Are we waiting for something to merge it?

k00ni:
I am currently very busy and hope I can give more feedback until the end of this month. But to be quite frank, I have no knowledge in this topic. It would help if you could add a link to the specification which covers this case?

Maybe @j0k3r, @GreyWyvern or @unixnut might have some thoughts about this, which they can share in the meantime.

ridgey-dev:
I just created a fix for the specific PDF (see PR). I didn't found a specification that addresses this issue.

Hopefully someone can look at it as it's a blocker for a project I'm working on!

k00ni:
@unixnut [mentioned](https://github.com/smalot/pdfparser/issues/761#issuecomment-2686884101):
> I think changing getTextArray() is the wrong approach. This will break functionality for people who expect elements with empty strings.

If that is the case I would decline this PR. Because of my limited time I can't look into this matter any further. But would be more than open to read any (technical) arguments or/and links which show that this change will not lead to a breaking change.

There might be an argument to use a switch in the `Config` class, but that feels kinda nasty at the moment.

We use semantic version and our users trust the behavior of PDFParser, even though I am not sure how rare this case might be in the wild.

@j0k3r what do suggest we do here?

k00ni:
Feel free to comment here or create a new issue if something changed in the situation.

PR #763, opened 2025-02-25
Fix incorrect parsing of bfrange (#631)

### LINKED ISSUE #631: Allowed memory exhausted when parse the PDF file.
- PHP Version: 7.4.33
 - PDFParser Version: 2.5.0

### Description:
Trying to parse this PDF always result in Allowed memory exhausted error. 

Error: Allowed memory size of 1077936128 bytes exhausted (tried to
allocate 335544320 bytes) in
...../smalot/pdfparser/src/Smalot/PdfParser/Font.php,
line 223

Set up PHP memory limit to 4GB did not help either.
I have also tried to setDecodeMemoryLimit to lower but still had the same memory issue. Setting Decode memory limit prevent the error only when I set it to 1000 or lower. So maybe it should be set in MB and not in bytes, or there is an bug in the code.
 

### PDF input

[test_pdf.pdf](https://github.com/smalot/pdfparser/files/12291120/test_pdf.pdf)

Parser should either parse the text from the PDF, or return empty string or some exception and not memory error.

### Code

```
$config = new Config();
$url = 'path_to_PDF_folder/test_pdf.pdf';
$config->setDecodeMemoryLimit(1000000);
$pdf = $parser->parseFile($url);
```

nsfisis:
# Type of pull request

* [ ] Documentation update
* [ ] Something else

# About

fixes #631

Previously, the regular expression for single-offset bfrange mapping unintentionally matched portions of the bracketed array form. This caused extremely large code range, leading to out of memory.

Now, we capture the `<from>` and `<to>` ranges first and then distinguish between single-offset (`<xxxx>`) and array (`[<xxxx> <xxxx> ...]`) forms. This ensures that the single-offset regex won't accidentally match bracketed segments.

# TODO

- [x] Write tests.

Question:
In order to write test code, I need to generate a PDF that contains arbitrary `bfrange` sections. However, I’m afraid I don’t know how to do that. Do you have any suggestions on how to create such a file?

### DIFF
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -216,11 +216,13 @@ class Font extends PDFObject
             // Support for multiple bfrange sections
             if (preg_match_all('/beginbfrange(?P<sections>.*?)endbfrange/s', $content, $matches)) {
                 foreach ($matches['sections'] as $section) {
-                    // Regexp to capture <from>, <to>, and either <offset> or [...] items.
-                    // - (?P<from>...) Source range's start
-                    // - (?P<to>...)   Source range's end
-                    // - (?P<dest>...) Destination range's offset or each char code
-                    //                 Some PDF file has 2-byte Unicode values on new lines > added \r\n
+                    /**
+                     * Regexp to capture <from>, <to>, and either <offset> or [...] items.
+                     * - (?P<from>...) Source range's start
+                     * - (?P<to>...)   Source range's end
+                     *  - (?P<dest>...) Destination range's offset or each char code
+                     *                 Some PDF file has 2-byte Unicode values on new lines > added \r\n
+                     */
                     $regexp = '/<(?P<from>[0-9A-F]+)> *<(?P<to>[0-9A-F]+)> *(?P<dest><[0-9A-F]+>|\[[\r\n<>0-9A-F ]+\])[ \r\n]+/is';
 
                     preg_match_all($regexp, $section, $matches);
@@ -230,7 +232,7 @@ class Font extends PDFObject
                         $char_to = hexdec($matches['to'][$key]);
                         $dest = $matches['dest'][$key];
 
-                        if (preg_match('/^<(?P<offset>[0-9A-F]+)>$/i', $dest, $offset_matches)) {
+                        if (1 === preg_match('/^<(?P<offset>[0-9A-F]+)>$/i', $dest, $offset_matches)) {
                             // Support for : <srcCode1> <srcCode2> <dstString>
                             $offset = hexdec($offset_matches['offset']);
 
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -239,11 +239,13 @@ class Font extends PDFObject
                             for ($char = $char_from; $char <= $char_to; ++$char) {
                                 $this->table[$char] = self::uchr($char - $char_from + $offset);
                             }
-
                         } else {
                             // Support for : <srcCode1> <srcCodeN> [<dstString1> <dstString2> ... <dstStringN>]
                             $strings = [];
-                            preg_match_all('/<(?P<string>[0-9A-F]+)> */is', $dest, $strings);
+                            $matched = preg_match_all('/<(?P<string>[0-9A-F]+)> */is', $dest, $strings);
+                            if (false === $matched || 0 === $matched) {
+                                continue;
+                            }
 
                             foreach ($strings['string'] as $position => $string) {
                                 $parts = preg_split(
@@ -252,6 +254,9 @@ class Font extends PDFObject
                                     0,
                                     \PREG_SPLIT_NO_EMPTY | \PREG_SPLIT_DELIM_CAPTURE
                                 );
+                                if (false === $parts) {
+                                    continue;
+                                }
                                 $text = '';
                                 foreach ($parts as $part) {
                                     $text .= self::uchr(hexdec($part));
--- a/src/Smalot/PdfParser/Font.php
+++ b/src/Smalot/PdfParser/Font.php
@@ -216,45 +216,46 @@ class Font extends PDFObject
             // Support for multiple bfrange sections
             if (preg_match_all('/beginbfrange(?P<sections>.*?)endbfrange/s', $content, $matches)) {
                 foreach ($matches['sections'] as $section) {
-                    // Support for : <srcCode1> <srcCode2> <dstString>
-                    $regexp = '/<(?P<from>[0-9A-F]+)> *<(?P<to>[0-9A-F]+)> *<(?P<offset>[0-9A-F]+)>[ \r\n]+/is';
+                    // Regexp to capture <from>, <to>, and either <offset> or [...] items.
+                    // - (?P<from>...) Source range's start
+                    // - (?P<to>...)   Source range's end
+                    // - (?P<dest>...) Destination range's offset or each char code
+                    //                 Some PDF file has 2-byte Unicode values on new lines > added \r\n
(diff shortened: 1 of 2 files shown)


k00ni:
> In order to write test code, I need to generate a PDF that contains arbitrary bfrange sections. However, I’m afraid I don’t know how to do that. Do you have any suggestions on how to create such a file?

You don't need a real PDF file, instead create a `Font` instances and set them up depending on your test case. If you want to test how it handles `bfrange` sections, you can create different strings with correct and faulty data and inject it in the instance. You might need to use other class instances too to build a working case, but it seems reasonable and doable.

Here are a few starting points which might help:
* https://github.com/smalot/pdfparser/blob/master/tests/PHPUnit/Unit/FontTest.php
* https://github.com/smalot/pdfparser/blob/master/tests/PHPUnit/Integration/FontTest.php#L165
* https://github.com/smalot/pdfparser/blob/master/tests/PHPUnit/Integration/PagesTest.php

nsfisis:
Added tests: `tests/PHPUnit/Integration/FontTest.php`

Ready for review!

src/Smalot/PdfParser/Font.php
```diff
@@ -216,45 +216,46 @@ public function loadTranslateTable(): array
             // Support for multiple bfrange sections
             if (preg_match_all('/beginbfrange(?P<sections>.*?)endbfrange/s', $content, $matches)) {
                 foreach ($matches['sections'] as $section) {
-                    // Support for : <srcCode1> <srcCode2> <dstString>
-                    $regexp = '/<(?P<from>[0-9A-F]+)> *<(?P<to>[0-9A-F]+)> *<(?P<offset>[0-9A-F]+)>[ \r\n]+/is';
+                    // Regexp to capture <from>, <to>, and either <offset> or [...] items.
+                    // - (?P<from>...) Source range's start
+                    // - (?P<to>...)   Source range's end
+                    // - (?P<dest>...) Destination range's offset or each char code
+                    //                 Some PDF file has 2-byte Unicode values on new lines > added \r\n
```
k00ni, line 223:
```suggestion
                    /**
                     * Regexp to capture <from>, <to>, and either <offset> or [...] items.
                     * - (?P<from>...) Source range's start
                     * - (?P<to>...)   Source range's end
                     *  - (?P<dest>...) Destination range's offset or each char code
                     *                 Some PDF file has 2-byte Unicode values on new lines > added \r\n
                     */
```

src/Smalot/PdfParser/Font.php
```diff
@@ -216,45 +216,46 @@ public function loadTranslateTable(): array
...
                         $char_from = hexdec($from);
                         $char_to = hexdec($matches['to'][$key]);
-                        $offset = hexdec($matches['offset'][$key]);
+                        $dest = $matches['dest'][$key];
 
-                        for ($char = $char_from; $char <= $char_to; ++$char) {
-                            $this->table[$char] = self::uchr($char - $char_from + $offset);
-                        }
-                    }
+                        if (preg_match('/^<(?P<offset>[0-9A-F]+)>$/i', $dest, $offset_matches)) {
```
k00ni, line 233:
```suggestion
                        if (1 === preg_match('/^<(?P<offset>[0-9A-F]+)>$/i', $dest, $offset_matches)) {
```

src/Smalot/PdfParser/Font.php
```diff
@@ -216,45 +216,46 @@ public function loadTranslateTable(): array
...
-                            );
-                            $text = '';
-                            foreach ($parts as $part) {
-                                $text .= self::uchr(hexdec($part));
+                        } else {
+                            // Support for : <srcCode1> <srcCodeN> [<dstString1> <dstString2> ... <dstStringN>]
+                            $strings = [];
+                            preg_match_all('/<(?P<string>[0-9A-F]+)> */is', $dest, $strings);
+
+                            foreach ($strings['string'] as $position => $string) {
```
k00ni, line 246: Can we be sure that `$strings['string']` key is always set?

src/Smalot/PdfParser/Font.php
```diff
@@ -216,45 +216,46 @@ public function loadTranslateTable(): array
...
+                                $parts = preg_split(
+                                    '/([0-9A-F]{4})/i',
+                                    $string,
+                                    0,
+                                    \PREG_SPLIT_NO_EMPTY | \PREG_SPLIT_DELIM_CAPTURE
+                                );
+                                $text = '';
+                                foreach ($parts as $part) {
```
k00ni, line 254: In case `preg_split` returns `false` PHP will complain. Maybe add an if-clause before the foreach-loop to check if `$parts` is valid?

tests/PHPUnit/Integration/FontTest.php
```diff
@@ -238,6 +238,68 @@ public function testLoadTranslateTable(): void
         $this->assertEquals('y', $table[92]);
     }
 
+    /**
+     * Tests loadTranslateTable with a bfrange definition that lists every destination character.
+     *
+     * @see https://github.com/smalot/pdfparser/issues/631
+     */
+    public function testLoadTranslateTableIssue631(): void
```
k00ni, line 246: Thanks for investing the time to add a test!

This seems to test the handling of elements, which couldn't be handled before. Would you mind elaborate a bit more on the this test? Is it worth thinking about further tests for other (edge) cases?

k00ni (changes requested):
First of all, thanks for your PR. Sorry for the delayed feedback, I am super busy currently.

There are only a few small remarks and suggestions. I hope you have some time left so we can finish this together. Looking forward to your repsonse.

k00ni:
_Just for future reference:_ info about bfrange handling in different PDF specifications. Because we support different versions it might be interesting what was defined and when.

- **PDF Specification 1.3**: see page 326 and later (https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.3.pdf)
- **PDF Specification 2.0**: see page 349 and later (https://developer.adobe.com/document-services/docs/assets/5b15559b96303194340b99820d3a70fa/PDF_ISO_32000-2.pdf)

src/Smalot/PdfParser/Font.php
nsfisis: I added check if `preg_match_all()` succeeds.

src/Smalot/PdfParser/Font.php
nsfisis: I added check if `preg_split()` succeeds.

tests/PHPUnit/Integration/FontTest.php
nsfisis: I added more valid or invalid test cases.

src/Smalot/PdfParser/Font.php
```diff
@@ -216,45 +216,53 @@ public function loadTranslateTable(): array
             // Support for multiple bfrange sections
             if (preg_match_all('/beginbfrange(?P<sections>.*?)endbfrange/s', $content, $matches)) {
                 foreach ($matches['sections'] as $section) {
-                    // Support for : <srcCode1> <srcCode2> <dstString>
-                    $regexp = '/<(?P<from>[0-9A-F]+)> *<(?P<to>[0-9A-F]+)> *<(?P<offset>[0-9A-F]+)>[ \r\n]+/is';
+                    /**
+                     * Regexp to capture <from>, <to>, and either <offset> or [...] items.
+                     * - (?P<from>...) Source range's start
+                     * - (?P<to>...)   Source range's end
+                     *  - (?P<dest>...) Destination range's offset or each char code
```
k00ni, line 223:
```suggestion
                     * - (?P<dest>...) Destination range's offset or each char code
```

k00ni:
Thank you @nsfisis

Issue #756, opened 2025-01-15
Possible startxrefPreg extension


xBambey:
Some PDFs in a project could not be read by the parser.

After a closer examination of the binary data, it was noticed that there is often a space before the reference byte.

After a brief search on the Internet, I could not find any information as to whether this space may be included or not. Perhaps someone here who is more familiar with the subject knows more.

By inserting an optional space in the RegEx at this point, the PDF is recognized again.

RegEx would then look like the following: 

`'/(?<=[\r\n])startxref[\s]*[\r\n]+[\s]*([0-9]+)[\s]*[\r\n]+%%EOF/i'`

https://github.com/smalot/pdfparser/blob/f44ada017eac4f607ffeb1caca96a2347d48f38f/src/Smalot/PdfParser/RawData/RawDataParser.php#L884-L891

unixnut:
Other files can cause the `Uncaught Exception: Unable to find startxref` error too.

getXrefData() uses a regex that requires a newline before the offset value.  This means that some PDF files (e.g. containing "startxref 1746580") won't parse.

Example PDF details seen in production:
Creator:        DocuPrint CM225/228
PDF version:    1.7

A different example will be attached soon.

A combination of my working fix for the problem mentioned in this comment, plus the one above, is:
`'/(?<=[\r\n])startxref(?:[\s]*[\r\n]+[\s]*|\s+)([0-9]+)[\s]*[\r\n]+%%EOF/i'`

unixnut:
> My current solution to the problem is
> 
> ```
> '/(?<=[\r\n])startxref[\s\r\n]+([0-9]+)[\s\r\n]+%%EOF/i'
> ```

This seems too general to me, e.g. allows multiple newlines between `startxref` statement and offset.  Also would allow the eof token on the same line as the offset.

@xBambey what did you think of my regex that allows for either newlines or spaces, plus optional leading/trailing spaces around the offset?

xBambey:
@unixnut your regex looks good. I kept my solution as general as possible because I don't want to keep making further adjustments.

Therefore there is also the possibility that the eof token can be on the same line as the offset.

I know that this doesn't normally happen, but there is the extremely rare case in whis a generation in the last functions of the PDF can throw an error, which cout actually cause this to happen.

unixnut:
I think the `startxref` statement and the eof marker are always produced together and it's safer to look for the eof marker on a new line.

[...]

PR #764, opened 2025-02-27
Bugfix for "Unable to find startxref" exception

unixnut:
# Type of pull request

* [Y] Bug fix (involves code and configuration changes)
* [ ] Documentation update
* [ ] Something else

# About

Addresses inadequacies in handling startxref statement before EOF marker. 

Currently, getXrefData() uses a regex that requires a newline before the offset value.

This fix allows:

- space before the reference offset.
- keyword and offset on the same line (e.g. containing "startxref 1746580").

For more detail see #756 

# TO DO

- tests 
- example files


k00ni:
I just want to check how you are doing. Any plans when to finish this work?

PR #776, opened 2025-07-25
CI: windows-2019 => windows 2022

k00ni:
# Type of pull request

# About

Fixes failing Windows related tests.

FYI: https://github.com/actions/runner-images/issues/12045#issue-2997151424

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -202,7 +202,7 @@ jobs:
 
   windows-tests:
     name: Windows-Tests with PHP ${{ matrix.php-versions }}
-    runs-on: windows-2019
+    runs-on: windows-2022
 
     strategy:
       fail-fast: true


k00ni:
Something ain't right ...

Bildschirmfoto vom 2025-07-25 14-42-06

There is no output and it runs way too long. God, I hope there is no other problem below the surface. Lets see how it looks later on.

j0k3r:
Yeah it's weird, it looks stuck ..

k00ni:
Our CI doesn't run through anymore. I assume, but don't know for sure, if its related to the Windows tests. Maybe they get stuck and therefore the rest of the pipeline is not being executed. Or there is something else.

**Has anyone an idea what to do now?**
If there is no solution soon I would rather remove the Windows tests from the workflow file.

rupertj:
I think the issue is this: https://github.com/actions/runner-images/issues/11101

The Ubuntu ubuntu-20.04 runner needs replacing with a newer version. I've tried ubuntu-24.04 in my fork and the test started quickly and passed: https://github.com/rupertj/pdfparser/actions

Issue #782, opened 2025-10-30
PDFObject::getTextArray() shouldn't include XObject Forms.


rupertj:
This is a follow up to #733, but could also fix some other issues, like #671.

In #733, I wrote: "I'm not sure if it's appropriate to skip Form and PS types here or not". I now think skipping Form XObjects is the right thing to do. NB that a Form XObject isn't a form in the sense of a Form to capture data. It's a Form as in a Shape. (Section 8.10.1 of https://opensource.adobe.com/dc-acrobat-sdk-docs/standards/pdfstandards/pdf/PDF32000_2008.pdf#page=217 helped my understanding here.)

I'd seen the consistent off-by-one error mentioned in #671 in the PDFs I'm working with. I have some code that's using the position of text to try to find the text inside a given annotation, and it was frequently finding the text to the left of the given annotation.

Each of my PDFs has a Form early on in each page's data, which I believe is the use case of "a form XObject may serve as the template for an entire page" mentioned in the spec.

Skipping the Form when creating the text array, much like we did for Images in the fix for #733, removes this off by one error for me.

MR and examples to follow.

PR #783, opened 2025-10-30
Ignore Form as well as Image XObjects when assembling the text array for a PDFObject.

rupertj:
Fix for #782

### DIFF
--- a/tests/PHPUnit/Unit/PDFObjectTest.php
+++ b/tests/PHPUnit/Unit/PDFObjectTest.php
@@ -11,6 +11,7 @@ use Smalot\PdfParser\Element\ElementArray;
 use Smalot\PdfParser\Header;
 use Smalot\PdfParser\Page;
 use Smalot\PdfParser\PDFObject;
+use Smalot\PdfParser\XObject\Form;
 use Smalot\PdfParser\XObject\Image;
 
 class PDFObjectTest extends TestCase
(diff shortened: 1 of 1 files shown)


k00ni (changes requested):
Thank you for taking the time to extend the test coverage for your changes!

However, I noticed that the changes in this PR modify existing test data (see [diff here](https://github.com/smalot/pdfparser/pull/783/files#diff-9ed339b02acb8464a0b8fca7c91b7cfb5264673b386cf2ea17977485bb530f85L44-R46)). This could introduce unintended side effects for the existing test logic.

**To avoid this risk, could you please:**

- Move your test code into a separate test method
- If helpful, you can copy an existing test as a template and adapt it for your new functionality.

This way, we ensure that the existing tests remain unaffected and your new functionality is tested in isolation.

Once that's done, your PR is good to go. By the way, your description in https://github.com/smalot/pdfparser/issues/782 was very helpful and provided good insights into the issue 👍

rupertj:
That change from "Imo" to "Im0" was just correcting a typo in the existing test. I didn't spot that I got that wrong when I wrote it.

I could revert that line and submit it as a separate PR if you like? I think keeping the new test coverage in the same method as the existing coverage makes sense, as they're testing the same bit of code.

k00ni:
Sorry for the delayed response. 

I follow your arguments, it looks good to me. The documentation provided in #782 was very helpful.

PR #784, opened 2025-11-01
Add case for 'endobj' in PDF parsing

NamelessCoder:
Fixes an issue with parsing some PDF documents which contain this instruction. I have no idea what it means.

# Type of pull request

* [ ] Documentation update
* [ ] Something else

# About

I am working on parsing a HUGE library of PDF files and some of those files contain an `endobj` header element
which isn't handled by the parser, and this results in an exception. The files do parse correctly if this header
element is simply ignored. It doesn't seem to have any practical use.

### DIFF
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -320,6 +320,7 @@ class Parser
 
             case 'endstream':
             case 'obj': // I don't know what it means but got my project fixed.
+            case 'endobj': // I don't know what this means either, but it is found in some PDF documents I tried to parse.
             case '':
                 // Nothing to do with.
                 return null;


k00ni (commented):
Changing code without tests that cover the changes might lead to unintended side effects. We therefore only allow code changes without tests under rare circumstances.

But lets take a step back and get an overview about the underlying issue:

1. I assume you receive an `Exception` (Invalid type) without this code change?
2. Can you provide a PDF which triggers this problem and is free of charge so we can use it in the test suite?
3. If you are not able to produce/provide a usable PDF (e.g. because it contains sensitive data) we have to think about a basic test case which triggers your code. 

Don't get me wrong, I know that a faulty library can disrupt a project in a significant way but just because a PDF fails to parse doesn't mean the library needs a change. I hope we can find a way to "harden" the library but also add at least 1 test to cover the change.

PR #787, opened 2025-11-24
Fix memory exhaustion caused by circular references in PDF xref chains

k00ni:
# Type of pull request

# About

This PR fixes xref-handling when it comes to circular references. A malformed PDF can lead to memory exhaustion because PDFParser enters an endless loop at some point. These fixes should prevent that.

It allows me to work with a bunch of "broken" PDFs, but I might have overlook some things. It would be great if someone else could have a look. Any feedback is appreciated (maybe @GreyWyvern @j0k3r?) :rocket:


k00ni:
> Then again, I'm pretty sure none of those PDFs were of the malformed type to have caused this error.

You would have noticed :sweat_smile:

Thank you very much for taking the time.

Issue #788, opened 2026-01-15
Page getText() returns blank contents


gerardnll:
- PHP Version: 8.4.14
 - PDFParser Version: 2.12.3

### Description:

Parsing in version 2.12.2 works fine, updating to the last one returns empty content when doing ->getText() to a page. I haven't touched the implementation.

### PDF input

I can provide the PDF privately.

k00ni:
I hope its not a regression. It is most likely due to the changes of #783. @rupertj do have an idea?

@gerardnll Could you go into the vendor folder in your local installation and in `smalot/pdfparser` change the file `src/Smalot/PdfParser/PDFObject.php` by reverting [this line](https://github.com/smalot/pdfparser/pull/783/changes#diff-76265e7497ce3b784c16b9d7aa78a903a393dfaaa4b138bde412f47f081ef145R793) and try if the problem persists?

MrMitch:
@rupertj PDF sent. I've added you as a collaborator on the minimal reproducer repository since i already built-it beforehand. There's a full description in the README, including the missing text. The PDF is in the repository as well.

PR #789, opened 2026-02-06
When assembling the text array for an object, skip Forms that don't contain any text, instead of all Forms.

rupertj:
Fix for #788.

### DIFF
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -803,7 +803,7 @@ class PDFObject
 
                         // If the PDFObject is a Form and doesn't have any text,
                         // skip it.
-                        if (($xobject instanceof Form) && ($objectText === ' ')) {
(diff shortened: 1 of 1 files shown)


src/Smalot/PdfParser/PDFObject.php
```diff
@@ -788,16 +788,26 @@ public function getTextArray(?Page $page = null): array
...
-                            $text[] = $xobject->getText($page);
+                        if (\in_array($xobject->getUniqueId(), self::$recursionStack, true)) {
+                            break;
+                        }
+
+                        $objectText = $xobject->getText($page);
+
+                        // If the PDFObject is a Form and doesn't have any text,
+                        // skip it.
+                        if (($xobject instanceof Form) && ($objectText === ' ')) {
```
k00ni:
```suggestion
                        if ($xobject instanceof Form && $objectText === ' ') {
```
No parentheses required when using  a combination of simple checks.

PR #779, opened 2025-08-08
Include PHP 8.5 in CI

k00ni:
# Type of pull request

* [x] Something else

# About

Includes PHP 8.5 in CI so we already see if there are issues with the new PHP version.

PR #793, opened 2026-04-07
Fix chr() deprecation in ASCII85 last-tuple decoding

Ciki:
## Problem

PR #779 fixed the `chr()` deprecation warning in the main ASCII85 decoding loop but missed the **"last tuple" switch/case block** (handling incomplete groups at the end of encoded data).

The same issue applies — bit-shifted values (`$tuple >> 16`, `$tuple >> 8`) can exceed 255, triggering:

```
chr(): Providing a value not in-between 0 and 255 is deprecated, this is because a byte value must be in the [0, 255] interval.
```

Observed in production when parsing PDFs with ASCII85-encoded streams that have a partial trailing group.

## Fix

Apply `& 0xFF` bitmask to extract individual bytes before passing to `chr()`, consistent with the approach used in the main loop fix from #779.

`& 0xFF` is the standard bitwise idiom for extracting a single byte from a multi-byte integer (equivalent to `% 256` but more idiomatic).

## Changes

```php
// Before
$decoded .= \chr($tuple >> 24).\chr($tuple >> 16).\chr($tuple >> 8);

// After
$decoded .= \chr(($tuple >> 24) & 0xFF).\chr(($tuple >> 16) & 0xFF).\chr(($tuple >> 8) & 0xFF);
```

Same pattern applied to `case 3` and `case 2`.

### DIFF
--- a/samples/bugs/PullRequest793.pdf
+++ b/samples/bugs/PullRequest793.pdf
@@ -1,0 +1,34 @@
+%PDF-1.4
+%‚„œ”
+1 0 obj
+<< /Type /Catalog /Pages 2 0 R >>
+endobj
+2 0 obj
+<< /Type /Pages /Kids [3 0 R] /Count 1 >>
+endobj
+3 0 obj
+<< /Type /Page /Parent 2 0 R /MediaBox [0 0 612 792]
+   /Contents 4 0 R /Resources << /Font << /F1 5 0 R >> >> >>
+endobj
+4 0 obj
+<< /Filter /ASCII85Decode /Length 81 >>
+stream
+6<#'\7PQ#?1*BP.+>khq2_Zp.<+I+"5uU-B8N8RMCghC,/Tc,SCh4`-G%G]+Ci=M?FCfN8.3MT)+@T6~>
+endstream
+endobj
+5 0 obj
+<< /Type /Font /Subtype /Type1 /BaseFont /Helvetica >>
+endobj
+xref
+0 6
+0000000000 65535 f 
+0000000015 00000 n 
+0000000064 00000 n 
+0000000121 00000 n 
+0000000250 00000 n 
+0000000404 00000 n 
+trailer
+<< /Size 6 /Root 1 0 R >>
+startxref
+474
+%%EOF
--- a/src/Smalot/PdfParser/RawData/FilterHelper.php
+++ b/src/Smalot/PdfParser/RawData/FilterHelper.php
@@ -209,7 +209,7 @@ class FilterHelper
                     if (255 < $tuple >> 24) {
                         $chr24Part = \chr(($tuple >> 24) % 256);
                     } else {
-                        $chr24Part = \chr($tuple >> 24);
+                        $chr24Part = \chr(($tuple >> 24) & 0xFF);
                     }
 
                     if (255 < $tuple) {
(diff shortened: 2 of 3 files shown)


k00ni:
@Ciki Thank you for the PR.

If I remember correctly, the [changes](https://github.com/smalot/pdfparser/pull/779/changes#diff-934dc8e182208bab6b10fa04c3b576ce4e20d94a04354248bcc395fd0cd06f36R193-R221) in #779 where suggested somewhere to fix PHP 8.5 related deprecations. I have limited knowledge in this area so I had to check your information. The following two pages describe bit-wise operations quite a bit:

1. https://medium.com/android-news/java-when-to-use-n-8-0xff-and-when-to-use-byte-n-8-2efd82ae7dd7
2. https://oscarliang.com/what-s-the-use-of-and-0xff-in-programming-c-plus-p/

Code looks good. Could you provide one of the PDFs which triggered these so we can use it in our test suite?

I will keep this open for a few days in case someone else wants to comment. Or do you need it merged right away?

Ciki:
@k00ni Thanks for the review!

The original PDF that triggered this contained client data, so I generated a minimal reproduction PDF instead.

**File:** [ascii85-last-tuple-overflow.pdf](https://github.com/user-attachments/files/26629410/ascii85-last-tuple-overflow.pdf)


It's a valid single-page PDF (657 bytes) with one ASCII85Decode-filtered content stream. The stream is crafted so that the last ASCII85 group (4-char incomplete tuple) produces values exceeding 255 when bit-shifted:

- `$tuple` = 541414470
- `$tuple >> 24` = 32 (OK)
- `$tuple >> 16` = 8261 (> 255 → **chr() deprecation**)
- `$tuple >> 8` = 2114900 (> 255 → **chr() deprecation**)

With the fix (`& 0xFF` masking), these become `chr(32)`, `chr(69)`, `chr(84)` — all valid byte values.

No rush on merging — take your time. We're running a fork in the meantime.

k00ni:
@Ciki I added your PDF to the test samples and a mini test which loads it. Without your fixes one would see the deprecations.

There were a few warnings (at least in PHP 8.5, [example](https://github.com/smalot/pdfparser/actions/runs/24074162328/job/70219155945?pr=793#step:6:25)):

```
Runtime:       PHP 8.5.4
Configuration: /home/runner/work/pdfparser/pdfparser/phpunit.xml

...............................................................  63 / 192 ( 32%)
............................................................... 126 / 192 ( 65%)
.................W.........W................................... 189 / 192 ( 98%)
...                                                             192 / 192 (100%)

Time: 00:16.086, Memory: 605.75 MB

2 tests triggered 11 PHP warnings:

1) /home/runner/work/pdfparser/pdfparser/src/Smalot/PdfParser/Font.php:145
The float 1.013347617134381E+19 is not representable as an int, cast occurred

Triggered by:

* PHPUnitTests\Integration\ParserTest::testIssue621 (2 times)
  /home/runner/work/pdfparser/pdfparser/tests/PHPUnit/Integration/ParserTest.php:302

* PHPUnitTests\Integration\ParserTest::testParseFile (2 times)
  /home/runner/work/pdfparser/pdfparser/tests/PHPUnit/Integration/ParserTest.php:58

2) /home/runner/work/pdfparser/pdfparser/src/Smalot/PdfParser/Font.php:145
The float 9.039039706125167E+20 is not representable as an int, cast occurred

Triggered by:

* PHPUnitTests\Integration\ParserTest::testIssue621 (2 times)
  /home/runner/work/pdfparser/pdfparser/tests/PHPUnit/Integration/ParserTest.php:302

* PHPUnitTests\Integration\ParserTest::testParseFile (2 times)
  /home/runner/work/pdfparser/pdfparser/tests/PHPUnit/Integration/ParserTest.php:58

[...]
```

Those are not because of your changes as far as I can tell, therefore ignored for now.

Thank you for taking the time preparing the pull request. It was a joy not having to ask for every detail but to get it prepared and on point.

k00ni (approved):

PR #801, opened 2026-04-24
Guard low-memory parsing by bounding Flate decode and image payload retention

vitormattos:
## Summary
- derive a conservative effective Flate decode cap when decodeMemoryLimit is not explicitly configured
- apply the same cap to the compress.zlib fallback path
- skip retaining image stream payloads when PHP memory is constrained to avoid cumulative OOM during parsing
- add focused regression coverage in DocumentIssueFocusTest for PullRequest457.pdf

## Reproduction
Before this patch, the command below fails with fatal OOM at memory_limit=128M:

```bash
docker run --rm -v "$PWD":/work -w /work smalot-pdf-bench:local \
  php -d memory_limit=128M -r 'require "vendor/autoload.php"; $parser = new Smalot\\PdfParser\\Parser(); $doc = $parser->parseFile("samples/bugs/PullRequest457.pdf"); echo count($doc->getPages()), PHP_EOL;'
```

After this patch it prints 28.

## Tests
- make run-phpunit ARGS="tests/PHPUnit/Integration/DocumentIssueFocusTest.php --filter testParseFileWithLargeFlateStreams"
- make run-phpunit ARGS="tests/PHPUnit/Integration/PageTest.php --filter testGetTextPullRequest457"

## PDF Source
- File: samples/bugs/PullRequest457.pdf
- Source URL: N/A (existing repository fixture reused for this fix)

### DIFF
--- a/src/Smalot/PdfParser/RawData/FilterHelper.php
+++ b/src/Smalot/PdfParser/RawData/FilterHelper.php
@@ -264,10 +264,12 @@ class FilterHelper
      */
     protected function decodeFilterFlateDecode(string $data, int $decodeMemoryLimit): ?string
     {
+        $effectiveDecodeMemoryLimit = $this->getEffectiveDecodeMemoryLimit($decodeMemoryLimit);
+
         // Uncatchable E_WARNING for "data error" is @ suppressed
         // so execution may proceed with an alternate decompression
         // method.
-        $decoded = @gzuncompress($data, $decodeMemoryLimit);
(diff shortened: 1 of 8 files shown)


vitormattos:
Superseded by the RawDataParser consolidation chain in the fork.

This fix (flate decode memory guard, MemoryLimit helper, test stabilization) is included in vitormattos/pdfparser#31, stacked directly on the `fix/invalid-object-reference-tolerant-parser` consolidation branch.

Supersedes smalot/pdfparser#801

PR #798, opened 2026-04-24
fix: tolerate startxref offset inside xref keyword

vitormattos:
Summary
- Fixes parsing when `startxref` points one byte inside the `xref` keyword (`ref`) or includes leading whitespace drift.
- Uses a normalized xref offset before deciding xref table vs xref stream path.
- Adds an integration regression fixture and test for the failing corpus file.

Reproduction
- Fixture parses with `pdfinfo` but failed in parser with `Invalid object reference for $obj.`.
- New test: `testParseFileWhenStartxrefPointsToLeadingWhitespaceInXrefStream`.

Changes
- `RawDataParser::getXrefData()` now computes and uses a whitespace-adjusted `startxref` offset.
- Adds one-byte tolerance for malformed offsets that land on `ref` instead of `xref`.
- Adds fixture: `samples/bugs/PullRequestStartxrefWhitespaceXrefStream.pdf`.
- Adds regression assertion in integration tests.\n\nPDF Source
- Corpus: veraPDF-corpus
- Fixture origin: PDF_A-1b/6.1 File structure/6.1.2 File header/veraPDF test suite 6-1-2-t01-fail-a.pdf
- Source URL: https://github.com/veraPDF/veraPDF-corpus/blob/staging/PDF_A-1b/6.1%20File%20structure/6.1.2%20File%20header/veraPDF%20test%20suite%206-1-2-t01-fail-a.pdf
- Local copy used in tests: samples/bugs/PullRequestStartxrefWhitespaceXrefStream.pdf

PR #803, opened 2026-04-24
Fix parsing when startxref points near xref keyword

vitormattos:
## Bug
A valid PDF from the veraPDF corpus failed with:

`Exception: Invalid object reference for $obj.`

`pdfinfo` reports `Pages: 1`, so the document should be parsed successfully.

## Root Cause
In this file, `startxref` points to a byte adjacent to the `xref` keyword (off by one).
The parser required an exact `xref` match at `startxref`, then incorrectly fell back to cross-reference stream parsing.
That path tried to resolve an invalid object reference and raised the exception.

## Fix
- Normalize the `startxref` probe offset by skipping leading PDF whitespace.
- Tolerate the case where `startxref` points one byte inside `xref` (`ref`) by probing one byte back.
- Use the normalized offset consistently for deciding between xref table and xref stream parsing.

## Tests
- Added regression fixture: `samples/bugs/PullRequest794.pdf`
- Added integration test: `DocumentIssueFocusTest::testParseFileWhenStartxrefPointsNearXrefKeyword`

The new test fails on `master` and passes with this patch.

## Source PDF
- veraPDF corpus test file:
  https://raw.githubusercontent.com/veraPDF/veraPDF-corpus/master/PDF_A-2b/6.6%20Metadata/6.6.2%20Metadata%20streams/6.6.2.3%20Schemas/6.6.2.3.2%20Extension%20schemas/veraPDF%20test%20suite%206-6-2-3-2-t01-pass-c.pdf


vitormattos:
Closing this as duplicate of #798 to keep discussion/history in a single PR.

PR #804, opened 2026-04-24
fix: recover hybrid xref offsets in pdf.js issue17147

vitormattos:
## Summary
- add regression fixture samples/bugs/PullRequest804-pdf.js.pdf from pdf.js issue17147
- add integration test covering parsing when a hybrid trailer points to /XRefStm and /Prev
- harden xref parsing to recover when startxref points near, but not exactly at, the xref keyword
- avoid type errors when xref-stream lookup starts on trailer dictionaries and follow trailer offsets instead

## Root cause
issue17147.pdf contains a hybrid-reference structure where trailer metadata references both /XRefStm and /Prev, and one xref offset points near xref table entries rather than exactly at xref.

The parser raised a type error (array passed as object reference) and then failed with Unable to find xref.

## Fix details
- in decodeXref(), follow /XRefStm trailers so hybrid references are decoded
- in decodeXrefStream(), guard object-reference type and recover from trailer dictionaries by following /XRefStm and /Prev
- in getXrefData(), search a small backward window for xref when offsets land shortly after the keyword

## Validation
- red/green regression on tests/PHPUnit/Integration/DocumentIssueFocusTest.php with filter testParseFileWithArrayXrefObjectReferenceInStream
- full run of tests/PHPUnit/Integration/DocumentIssueFocusTest.php
- direct parse smoke-check of issue17147.pdf returns 1 page

All validation was run in Docker (php:8.3-cli) because host PHP is unavailable.

## Source PDF
- https://github.com/mozilla/pdf.js/blob/master/test/pdfs/issue17147.pdf


vitormattos:
Superseded by the RawDataParser consolidation chain in the fork.

This fix (hybrid xref offset recovery, XRefStm handling, backward xref scan) is included in vitormattos/pdfparser#32, stacked on `fix/invalid-object-reference-tolerant-parser`.

PR #795, opened 2026-04-24
Improve PDF.js corpus resilience with parser-native page recovery, dimensions API, and regression coverage

vitormattos:
## Objective
Strengthen parser reliability on malformed real-world PDFs and validate outcomes directly by parser readability (not by external tool comparison).

## What this PR changes
- Fixes page-tree traversal and recovery paths to handle malformed/cyclic/repeated Kids references more safely.
- Improves page box resolution and fallback behavior for malformed or missing page boxes.
- Adds robust handling for readable-encrypted edge cases (files marked as encrypted but still readable).
- Expands regression coverage with additional PDF.js fixtures, including broken xref/stream/encryption-marked cases.
- Introduces a native page dimensions API:
  - `Page::getDimensions($boxName = 'CropBox')`
  - `Document::getPagesDimensions($boxName = 'CropBox')`
- Simplifies usage documentation to use the new native dimensions API instead of workaround logic.

## Validation approach
- Parser-centric validation over the Mozilla PDF.js corpus.
- Each file is classified only by parser readability:
  - `Readable (plain)`
  - `Readable (encryption-marked)`
  - `Unreadable by parser`

## Results (PDF.js corpus: 930 files / 929 unique hashes)

| Status | Baseline (master) | This branch | Δ |
|---|---:|---:|---:|
| ✅ Readable (plain) | 870 | 922 | +52 |
| 🔐 Readable (encryption-marked) | 2 | 6 | +4 |
| ❌ Unreadable by parser | 58 | 2 | -56 |
| **Total** | **930** | **930** | **-** |
| parser_success_rate | 93.7634% | 99.7849% | +6.0215 pp |

## Additional validation
- Full PHPUnit suite remains green: **273 tests**, **1649 assertions**.
- VeraPDF corpus baseline validation remains stable.

## Notes
- This PR does **not** claim full decryption support for explicit user-password protected PDFs.
- It does improve handling of encryption-marked files that are readable in practice.

PR #806, opened 2026-04-24
fix: guard cyclic page tree traversal

vitormattos:
## Bug fixed in this PR
Deep page-tree traversal could enter infinite recursion when `/Kids` contained a cycle.

## Fixture(s) and source
- `samples/bugs/PullRequest806-pdf.js.pdf`
  - https://github.com/mozilla/pdf.js/blob/master/test/pdfs/Pages-tree-refs.pdf

## Note to maintainer
You can merge this PR directly, or use it only as a code-review helper and merge the aggregator PR (#809).

### DIFF
--- a/samples/bugs/PullRequest806-pdf.js.pdf
+++ b/samples/bugs/PullRequest806-pdf.js.pdf
@@ -1,0 +1,84 @@
+%PDF-1.7
+
+1 0 obj
+  << /Type /Catalog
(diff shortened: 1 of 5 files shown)


vitormattos:
Consolidated into #795 to avoid repeated cross-PR conflicts and keep a single review lane. Closing this PR as part of branch unification strategy.

PR #808, opened 2026-04-25
fix: recover PDFs without a valid startxref stanza

vitormattos:
## Summary
- recover xref parsing when no valid `startxref` stanza is present
- tighten `strpos` comparisons to avoid false positives at offset `0`
- add fallback xref reconstruction from object headers plus trailer metadata
- add a regression for `issue19800.pdf`

## Why
Some malformed PDFs have no valid `startxref` stanza but still contain enough trailer/object-header information to recover the document.

## Validation
- parser sanity check in container:
  - `issue19800.pdf` => `pages=1`
- focused PHPUnit:
  - `testParseFileWithoutStartxrefButWithTrailerRoot`

### DIFF
--- a/samples/bugs/PullRequest809-pdf.js.pdf
+++ b/samples/bugs/PullRequest809-pdf.js.pdf
@@ -1,0 +1,90 @@
+%PDF-1.4
+1 0 obj
+<<
+ /Type /Catalog
+ /Outlines 2 0 R
+ /Pages 3 0 R
+>>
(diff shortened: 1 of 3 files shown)


vitormattos:
Superseded by the consolidation chain in the fork: vitormattos/pdfparser#35\n\nIncludes missing-startxref recovery and refactor, integrated on top of PR807 stack with fixture/test reorganization.

PR #810, opened 2026-04-25
fix: tolerate startxref offset near xref keyword

vitormattos:
## Summary

Tolerate `startxref` offsets that land on whitespace immediately before the `xref` keyword, or one byte into `xref`.

## Why

Some malformed PDFs do not point `startxref` exactly at the beginning of `xref`, even though the cross-reference table is still recoverable.

Without this tolerance, parsing can fail with `Invalid object reference for $obj.` on files that should still be recoverable.

## Validation

Validated against representative regressions from the corpora used in the larger PDF.js integration work:

- PDF.js: `outlines_for_editor.pdf`
- VeraPDF: `veraPDF test suite 6-6-2-3-2-t01-pass-c.pdf`
- VeraPDF: `veraPDF test suite 6-1-2-t01-fail-a.pdf`

## Notes

This is a focused fix and can be reviewed independently from the aggregate integration PR.


vitormattos:
Superseded by the consolidation chain in the fork: vitormattos/pdfparser#36\n\nThis includes PR810 parser fix and regression coverage, integrated on top of PR808 stack with fixture/test consolidation.

PR #811, opened 2026-04-25
fix: recover missing root object when xref stream is incomplete

vitormattos:
## Summary
- add fallback xref recovery when trailer /Root reference is missing from decoded xref entries
- scan indirect object headers in the PDF body and merge only missing object offsets
- add regression fixture and integration test for PDF.js issue18986

## Why
Some malformed xref streams decode without the /Root object entry. In this case the parser returned zero pages with no exception because Catalog/Page objects were never loaded.

## Validation
- focused integration test (Docker): testMalformedXrefStreamMissingRootEntryStillParsesPage passes
- dataset harness repro on issue18986.pdf: parser_pages changed from 0 to 1 (status remains pdfinfo_error because pdfinfo still fails on malformed input)

## Notes
Maintainers can choose either path:
- merge this PR individually, or
- merge the aggregated draft integration PR instead.


vitormattos:
Superseded by the consolidation chain in the fork: vitormattos/pdfparser#37\n\nThis includes the missing-/Root xref stream recovery integrated on top of PR810 stack with fixture/test consolidation.

PR #809, opened 2026-04-25
integration: consolidated PDF.js parsing resilience fixes

vitormattos:
## PDF.js validation: aggregated fixes

**Objective:** Validate smalot/pdfparser against the Mozilla PDF.js corpus by comparing parser output with [pdfinfo](https://poppler.freedesktop.org/documentation.html) (Poppler reference implementation).

**Validation approach:** Run parser on each PDF, extract page count. Compare against pdfinfo page count. Files match = ok. Mismatch or parser error = issue to fix.

**Results ([PDF.js corpus](https://github.com/mozilla/pdf.js), 930 files / 929 unique hashes):**

| Status | Baseline | This branch | Δ |
|--------|----------|-------------|---|
| ✅ ok | 865 | 909 | +44 |
| ❌ parser_error | 41 | 0 | -41 |
| ⚠️ both_error | 13 | 0 | -13 |
| ℹ️ pdfinfo_error | 7 | 20 | +13 |
| 🔀 mismatch | 3 | 0 | -3 |
| **Total** | **929** | **929** | **—** |

**Success rate:** 93.1% → 97.8% (+4.7pp)

**Deduplication note:** the validation state is keyed by SHA-256, so two byte-identical files (`empty.pdf` and `empty#hash.pdf`) collapse into one effective corpus entry. The directory contains 930 PDFs, but the aggregate counts operate on 929 unique hashes.

**Note on `pdfinfo_error` (+13):** this is a classification shift, not a parser regression. These files are cases where parser-side failures were reduced while pdfinfo still reports malformed-input errors (e.g. Poppler syntax or xref failures). The parser now successfully reads more PDFs than pdfinfo can process.

**Additional validation:**
- [VeraPDF corpus](https://github.com/veraPDF/veraPDF-corpus): **2,907 PDFs** validated with 100% pass rate on baseline tests

**Review options:**

Maintainer can choose either path:

1. **Review and merge individual fixes**: Each focused fix reviewed and merged separately (granular approach)
2. **Fast-track integration PR**: Remove draft status and merge this consolidated PR (aggregated approach)

Both paths achieve the same end state. Choose based on review capacity and preference.

PR #814, opened 2026-04-26
fix(rawdata): guard flate decode memory and add MemoryLimit helper

vitormattos:
## Bug fixed in this PR
In low-memory scenarios, Flate stream decoding could become unstable or fail due to unsafe decode limits. This PR adds memory guards and a dedicated helper to make decoding behavior predictable.

## Fixture(s) and source
- This PR does not add a new PDF fixture.
- It validates behavior using existing fixtures already in the repository (for example `samples/bugs/PullRequest457.pdf`).
- Repository fixture link: https://github.com/smalot/pdfparser/blob/master/samples/bugs/PullRequest457.pdf

## Note to maintainer
You can merge this PR directly, or use it only as a code-review helper and merge the aggregator PR (#809).


vitormattos:
Consolidated into #795 to avoid repeated cross-PR conflicts and keep a single review lane. Closing this PR as part of branch unification strategy.

PR #816, opened 2026-04-27
fix(rawdata): consolidate malformed xref/startxref and page-tree recovery

vitormattos:
## Summary
Improve RawData and parser resilience for malformed PDF.js-style documents by combining xref/startxref recovery, tolerant object/header parsing, and page-tree/catalog recovery fixtures.

## Included Coverage
- Recover malformed or missing startxref/xref table placement.
- Tolerate malformed xref-stream offsets and nearby indirect object headers.
- Improve header parsing tolerance for malformed tokens and ObjStm edge-cases.
- Keep page-tree/catalog extraction robust on malformed Kids/catalog structures.
- Add and validate regression fixtures from PDF.js and related malformed corpora.

## Updated Areas
- `src/Smalot/PdfParser/RawData/RawDataParser.php`
- `src/Smalot/PdfParser/Parser.php`
- `src/Smalot/PdfParser/Pages.php`
- `tests/PHPUnit/Integration/RawData/RawDataParserTest.php`
- `tests/PHPUnit/Integration/DocumentIssueFocusTest.php`
- `samples/bugs/rawdata/*`

## Validation
- `tests/PHPUnit/Integration/RawData/RawDataParserTest.php`
- `tests/PHPUnit/Integration/DocumentIssueFocusTest.php`

PR #818, opened 2026-04-29
Parser ObjStm hardening and regression tests

vitormattos:
## Description
Incremental update to fix/rawdata-recovery-stack branch adding:
- Parser.php: ObjStm malformed preamble hardening with defensive null checks
- ParserTest.php: regression test for malformed ObjStm parsing

These changes complement existing rawdata recovery logic and prevent parser crashes on corrupted object streams.

## Testing
- Regression test passes: testParseFileWithMalformedObjectStreamPreamble
- Fixture: samples/bugs/bug1978317.pdf

## Related
Builds on existing fix/rawdata-recovery-stack commits for xref/object stream recovery.


vitormattos:
Closing: These commits should be part of PR #816 (fix/rawdata-next-xref-trailer-recovery) instead, as they modify the same PHP files (RawDataParser.php, Parser.php, RawDataParserTest.php). Multiple PRs modifying the same files will cause merge conflicts when the maintainer tries to merge individually. Will consolidate into PR #816 instead.

PR #819, opened 2026-04-29
RawData xref recovery tolerance for malformed offsets

vitormattos:
## Description
Incremental update to fix/recover-invalid-xref-offsets branch adding:
- RawDataParser.php: isRecoverableXrefLookupException() helper and try-catch guards around XRefStm/Prev traversal
- RawData integration tests: testParseFileWithInvalidXrefOffsetFromPdfJsCorpus for REDHAT fixture
- Fixtures: REDHAT-1531897-0.pdf, PullRequest812-issue7229.pdf

This allows graceful recovery when xref offset chains are malformed without aborting parse.

## Testing
- Regression test passes: testParseFileWithInvalidXrefOffsetFromPdfJsCorpus
- Fixtures from PDF.js corpus demonstrating xref offset recovery

## Related
Builds on existing fix/recover-invalid-xref-offsets commits for invalid offset tolerance.


vitormattos:
Closing: These commits should be part of PR #816 (fix/rawdata-next-xref-trailer-recovery) instead, as they modify the same PHP files (RawDataParser.php, RawDataParserTest.php). Multiple PRs modifying the same files will cause merge conflicts when the maintainer tries to merge individually. Will consolidate into PR #816 instead.

Issue #802, opened 2026-04-24
Parsing form object that is referenced by Do command


DominikDostal:
- PHP Version: 8.1.24
 - PDFParser Version: 2.12.5

### Description:

User of the website im working on apparently used the website https://online2pdf.com to convert a word document into a pdf.
This website seems to generate PDFs in a weird way:
Instead of just putting the text into the page object as content it generated a form object containing the text and referenced this object in the page content with the `Do` command.

Currently PDFParser doesnt support the `Do` command and also can't correctly extract content (DataTm) from form objects.

### PDF input

[Test.pdf](https://github.com/user-attachments/files/27051642/Test.pdf)

### Expected output & actual output
Expected:
Array containing the text `Test` and the coordinates of this textarea

Actual:
```
array (size=0)
  empty
```

### Code
```php
$filepath = "C:/Users/dd/Desktop/Test.pdf";
$filecontent = file_get_contents($filepath);
$config = new Config();
$config->setDataTmFontInfoHasToBeIncluded(true);
$parser = new Parser([], $config);
$pdf = $parser->parseContent($filecontent);
$pages = $pdf->getPages();
foreach ($pages as $page) {
    var_dump($page->getDataTm());
}
```

PR #820, opened 2026-05-04
Implement 'Do' command and support Form referenced by 'Do' commands

DominikDostal:
# Type of pull request

* [x] New feature (involves code and configuration changes)
* [ ] Documentation update
* [ ] Something else

# About

Added support for `Do` command for Issue #802

# Discuss the draft

Opened as Draft for now to show how I managed to make it work with my test pdf from #802 

I think it's best to discuss the changes I had to make first, to find out if there is maybe a better way to do it:
1) `$extractedTexts` can be empty if all Text is inside the object referenced by `Do` command
-> Added a check to only skip all next commands if extracted text wasnt empty to begin with
2) `extractRawData` and `getTextArray` takes content from the wrong property for Form objects
Page object gets it with `$content = $this->get('Contents');` which takes it from the `elements` array defined in the Header object
Form object content can be access with:
```
$header = new Header([], $this->document);
$content = new PDFObject($this->document, $header, $this->content, $this->config);
```
I copied this part from another location and found that it worked. Im pretty sure we can create this object somewhere in the constructor of form or the factory, but sadly i didnt manage to find the place where i should do that.

### DIFF
--- a/src/Smalot/PdfParser/Page.php
+++ b/src/Smalot/PdfParser/Page.php
@@ -532,6 +532,9 @@ class Page extends PDFObject
                 case 'cm':
                     $extractedData[] = $command;
                     break;
+                case 'Do':
+                    $extractedData[] = $command;
+                    break;
                     /*
                      * ET
                      * End a text object, discarding the text matrix
@@ -733,7 +736,7 @@ class Page extends PDFObject
             // If we've used up all the texts from getTextArray(), exit
             // so we aren't accessing non-existent array indices
             // Fixes 'undefined array key' errors in Issues #575, #576
-            if (\count($extractedTexts) <= \count($extractedData)) {
+            if (\count($extractedTexts) > 0 && \count($extractedTexts) <= \count($extractedData)) {
                 break;
             }
             $currentText = $extractedTexts[\count($extractedData)];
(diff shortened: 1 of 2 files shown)


k00ni:
@DominikDostal Thank you for taking the time. I am currently very busy, but try to get back to you in more detail in the next weeks.

In the meantime, would you be so kind and check https://github.com/smalot/pdfparser/pull/795, if it covers code you also changed? If that is the case I would see that we further discuss implementation details.

DominikDostal:
> @DominikDostal Thank you for taking the time. I am currently very busy, but try to get back to you in more detail in the next weeks.

Dont worry about it, take your time. I managed to make it work for my project, just wanted to share my findings with the community :)

> In the meantime, would you be so kind and check #795, if it covers code you also changed? If that is the case I would see that we further discuss implementation details.

I looked through it, but my PR doesnt seem to have any overlapping code with #795

k00ni:
@DominikDostal could you cite the sources you based your changes on? (PDF specification, ...)

DominikDostal:
> @DominikDostal could you cite the sources you based your changes on? (PDF specification, ...)


For the Do command itself I based it on https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/PDF32000_2008.pdf
image

Just a simple reference to another XObject that is supposed to be added at the current location.

As for the changes I made in Form.php:
I think the problem I fixed here was that the raw data wasnt extracted from the object correctly.

Changes between Page.php and Form.php side by side (extracted for better readability):
image

Somehow I couldn't find out where exactly the parsing of these 2 objects deviates that make such changes necessary, maybe you know more about that?

k00ni (commented):
@DominikDostal To be quite frank, I am not the right person to review and discuss this at the level it would require. My knowledge of the PDF specification is too limited for that, and at the moment I also don't have the time or resources to build up that expertise in a way that would allow me to give you a reliable review.

You wrote:

> I think it's best to discuss the changes I had to make first, to find out if there is maybe a better way to do it

I understand this PR as a proposal that works for your use case, but where the implementation details and the best overall approach are still open for discussion. If that is a fair reading, then I do not feel confident enough to judge which approach would be the correct one in terms of the PDF specification and the internal architecture of this library.

I also have a couple of technical concerns regarding the current implementation itself. I am aware that this is (to some extent) test code to have something we can talk about. However, it still raises questions about ordering, transformation state handling, recursion, and the amount of logic duplicated in `Form.php`.

More broadly, this PR and also #795 make me realize that the current way this library is maintained is no longer sufficient for changes which touch PDF specific details. There is simply not enough specification knowledge, implementation confidence, and available time on my side to review and maintain contributions like this responsibly.

That is unfortunate, and I am sorry that I cannot give this PR the level of review it would deserve.

Given that situation, I currently see two realistic outcomes:
1. someone with deeper PDF-spec and parser expertise steps in and reviews or continues this work, or
2. this PR cannot be merged here, even if the general direction may be valid.

I suggest you look into https://github.com/PrinsFrank/pdfparser, which is a great library and gives better results in my projects than this library.

PR #823, opened 2026-06-04
Fix Font::getDetails() TypeError when Encoding is indirect reference to PDFObject

### LINKED ISSUE #822: Font::getDetails() throws TypeError when Encoding is indirect reference to PDFObject
- PHP Version: 8.x (reproduced on 8.1–8.5)
- PDFParser Version: 2.12.4+

### Description:

`Font::getDetails()` throws a `TypeError` when a font's `/Encoding` entry is stored as an indirect object reference (`N 0 R`) that resolves to a plain `PDFObject` — i.e. an encoding dictionary that lacks `/Type /Encoding`.

This is a valid PDF structure per PDF spec Table 5.11: an encoding dictionary may carry only a `/Differences` array and omit `/Type` and `/BaseEncoding`.

`Header::resolveXRef()` correctly resolves the reference to a `PDFObject`. However, `Font::getDetails()` then executes `(string) $this->get('Encoding')`, which fails because `PDFObject` has no `__toString()` method.

**Error:**
```
Object of class Smalot\PdfParser\PDFObject could not be converted to string
in Font.php:97
```

### PDF input

Minimal reproducer (included in the fix PR under `samples/bugs/EncodingAsIndirectPDFObject.pdf`).

The relevant PDF structure:
```
4 0 obj
<< /Type /Font /Subtype /Type1 /BaseFont /Helvetica /Encoding 5 0 R >>
endobj

5 0 obj
<< /Differences [32 /space 65 /A /B /C /D /E /F /G /H /I /J] >>
endobj
```

**Expected:** `Font::getDetails()` returns `['Encoding' => 'Ansi', ...]` (or the `BaseEncoding` name if present) without throwing.

**Actual:** `TypeError: Object of class Smalot\PdfParser\PDFObject could not be converted to string`

### Code

```php
$pdf    = $parser->parseFile('EncodingAsIndirectPDFObject.pdf');

foreach ($pdf->getPages() as $page) {
    foreach ($page->getFonts() as $font) {
        $font->getDetails(); // throws TypeError
    }
}
```

kernie:
# Type of pull request

* [ ] Documentation update
* [ ] Something else

# About

fixes #822

`Font::getDetails()` threw `"PDFObject could not be converted to string"` when a font's `/Encoding` entry was an indirect reference resolving to a plain `PDFObject` (encoding dictionary without `/Type /Encoding`). This is a valid PDF structure per spec Table 5.11.

**Root cause:** `Font::getDetails()` applied an unconditional `(string)` cast to the result of `$this->get('Encoding')`. `Header::resolveXRef()` can return a `PDFObject` (which has no `__toString()`), causing a `TypeError` in PHP 8+.

**Fix:** Check the type before casting:
- `PDFObject` (covers both `Encoding` subclass and plain encoding dictionaries): extract `BaseEncoding` name; fall back to `'Ansi'` if absent — consistent with `Encoding::getDetails()['BaseEncoding']`.
- `Element` (direct name like `/WinAnsiEncoding`): cast as before.
- `null` (no entry): return `'Ansi'` as before.

**Changes:**
- `src/Smalot/PdfParser/Font.php` — fix in `getDetails()`
- `tests/PHPUnit/Unit/FontTest.php` — 5 unit tests covering all branches
- `tests/PHPUnit/Integration/FontTest.php` — integration test with sample PDF
- `samples/bugs/EncodingAsIndirectPDFObject.pdf` — minimal reproducer (720 B)

**AI disclosure**
This PR was developed with Claude Code assistance. All code has been reviewed as best I can. I tested the code in my local installation.

### DIFF
--- a/samples/bugs/EncodingAsIndirectPDFObject.pdf
+++ b/samples/bugs/EncodingAsIndirectPDFObject.pdf
@@ -1,0 +1,48 @@
+%PDF-1.4
+1 0 obj
+<< /Type /Catalog /Pages 2 0 R >>
+endobj
+
+2 0 obj
+<< /Type /Pages /Kids [3 0 R] /Count 1 >>
+endobj
+
+3 0 obj
+<< /Type /Page /Parent 2 0 R
+   /MediaBox [0 0 612 792]
+   /Resources << /Font << /F1 4 0 R >> >>
+   /Contents 6 0 R >>
+endobj
+
+4 0 obj
+<< /Type /Font
+   /Subtype /Type1
+   /BaseFont /Helvetica
+   /Encoding 5 0 R >>
+endobj
+
+5 0 obj
+<< /Differences [32 /space 65 /A /B /C /D /E /F /G /H /I /J] >>
+endobj
+
+6 0 obj
+<< /Length 37 >>
+stream
+BT /F1 12 Tf 100 700 Td (Hello) Tj ET
+endstream
+endobj
+
+xref
+0 7
+0000000000 65535 f 
+0000000009 00000 n 
+0000000059 00000 n 
+0000000117 00000 n 
+0000000253 00000 n 
+0000000349 00000 n 
+0000000429 00000 n 
+trailer
+<< /Size 7 /Root 1 0 R >>
+startxref
+517
+%%EOF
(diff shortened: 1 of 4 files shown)


k00ni:
Thank you @kernie! I hope there is a German holiday soon, so I have some time to get on this :sweat_smile: Its looks very polished already at first glance.

k00ni:
Thank you again. Sorry for the late reply.

PR #824, opened 2026-06-18
Improve Memory efficiency

splitbrain:
# Type of pull request

* [X] New feature (involves code and configuration changes)
* [ ] Documentation update
* [ ] Something else

# About

This branch combines three independent, complementary changes that lower peak memory consumption when parsing PDFs while keeping extracted text and document details byte-for-byte identical.

* 263a01b5517c616199fd015ebf75aeadfaa5abf5 build PDF objects incrementally -> improves memory during parsing
* ca9090a92dee496018e222cc7718d35f37b34336 chunk section building -> improves memory during text extraction
* bfa950cead3feb69aaab9ce6082f7ea0edeff58b optionally store content on disk instead in memory -> lowers retained memory after parsing

Each commit message contains the measured improvement of the commit on its own against master. Each individual commit should be easy to review (not many changes). My repository contains a branch for each of these commits separately if you'd rather only merge one or two of them - the commits slightly overlap, this PR addresses the merge conflicts for you.

The commit message of the final merge (2e56042dc3f3129b8dd1ad6f81030ae3ac6a56e9) has memory measurements of all three combined, resulting in gains between roughly 35 to 60%.

- [x] **Tests included** - `tests/PHPUnit/Unit/ContentSpoolTest.php`, plus `Config` unit + integration tests asserting spool-on/off produce identical output. Full suite passes
- [x] **Docs updated** - new `setContentSpooling` row + section in `doc/CustomConfig.md`.
- [x] **Coding style (PHP-CS-Fixer)** - no issues

### DIFF
--- a/src/Smalot/PdfParser/RawData/RawDataParser.php
+++ b/src/Smalot/PdfParser/RawData/RawDataParser.php
@@ -945,16 +945,21 @@ class RawDataParser
     }
 
     /**
-     * Parses PDF data and returns extracted data as array.
+     * Normalize the raw PDF data and decode the cross-reference/trailer data.
+     *
+     * Returns the xref/trailer data together with the normalized PDF data so
+     * callers (e.g. Parser) can inspect the trailer (for instance to detect
+     * encryption) and then stream the objects one at a time via
+     * getObjectsStream() instead of materializing them all at once.
      *
      * @param string $data PDF data to parse
      *
-     * @return array array of parsed PDF document objects
+     * @return array{0: array, 1: string} [$xref, $pdfData]
      *
      * @throws EmptyPdfException if empty PDF data given
      * @throws MissingPdfHeaderException if PDF data missing `%PDF-` header
      */
-    public function parseData(string $data): array
+    public function parseHeaderAndXref(string $data): array
     {
         if (empty($data)) {
             throw new EmptyPdfException('Empty PDF data given.');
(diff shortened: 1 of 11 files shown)


src/Smalot/PdfParser/RawData/RawDataParser.php
```diff
@@ -945,16 +945,21 @@ protected function getXrefData(string $pdfData, int $offset = 0, array $xref = [
...
      * @param string $data PDF data to parse
      *
-     * @return array array of parsed PDF document objects
+     * @return array{0: array, 1: string} [$xref, $pdfData]
      *
      * @throws EmptyPdfException if empty PDF data given
      * @throws MissingPdfHeaderException if PDF data missing `%PDF-` header
      */
-    public function parseData(string $data): array
+    public function parseHeaderAndXref(string $data): array
```
k00ni, line 962: Thank you for your PR. Just a quick note: It might be worth further discussing if `RawDataParser` (and other internal classes) are considered part of the public API but regardless, changing a function and its return value is something we need to talk about. Could you elaborate on that change please? I am not sure yet if I am OK with this change.

splitbrain: Basically parsedata was split into two parts: the header parsing and the object parsing. the latter now yields one object at a time - enabling more efficient memory handling. 

if you consider parsedata a public api, I can readd it wrapping the two new methods - it's like 5 lines or so. but it will be unused by the library itself.

k00ni:
Could you please close this PR and split it up into 3 PRs which reflect the structure you talked about? This would help me to decide if its manageable or not.

splitbrain:
Okay I submitted all three commits as separate PRs. TBH I think they only make sense in combination and if you want them all three, you land here again ;-)

I did not readd the parseData() method yet. Let me know if you want it here or on #827 or not at all.

I let you close whichever PRs you don't want.

k00ni:
Thank you again for all these changes. I will only take the ones (#828, #827) which refine/fix code without introducing new stuff (such as in #829).

PR #825, opened 2026-06-18
Fix float-to-int cast warning in Font::uchr on PHP 8.1+

splitbrain:
# Type of pull request

* [ ] Documentation update
* [ ] Something else

# About

When a CMap contains an oversized hex value, hexdec() returns a float larger than PHP_INT_MAX. Casting it to int in Font::uchr() raised a "not representable as an int" warning on PHP 8.1+. Such a value can never be a valid Unicode code point, so it is now treated as a missing character.

Fixes warnings during running the tests

* [x] Tests and Coding Style checks run

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,21 +1,33 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- https://phpunit.de/manual/current/en/appendixes.configuration.html -->
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.5/phpunit.xsd" backupGlobals="false" bootstrap="vendor/autoload.php" colors="true" processIsolation="false" stopOnFailure="false" cacheDirectory=".phpunit.cache" backupStaticProperties="false" displayDetailsOnTestsThatTriggerWarnings="true">
-  <php>
-    <ini name="error_reporting" value="-1"/>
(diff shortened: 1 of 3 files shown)


k00ni:
Thank you @splitbrain!

PDF parsing can be goofy sometimes, so even if that seems like a small fix, I would like to back it up. Could you add a basic test which triggers the behavior when running PHP 8.1? What is the current return value without this fix? Could it impact expected parsing results?

splitbrain:
I will look into it later. however the warnings were triggered by the existing testsuite for me on php 8.5. looking at the code I believe they should have triggered since 8.1 but I have not verified that.

splitbrain:
I added a specific test for the behavior. See 6053be6 for details.

I was indeed wrong about this being a PHP 8.1 change - I thought it was related to [Deprecate implicit non-integer-compatible float to int conversions](https://wiki.php.net/rfc/implicit-float-int-deprecate), but it's actually [Adding warnings to implicit fallible conversions](https://wiki.php.net/rfc/warnings-php-8-5) introduced in PHP 8.5.

>  What is the current return value without this fix?

As far as I understand it's not 100% defined and depends on the system architecture. Most cases would be a negative integer (which is not a valid Unicode point) - see the linked RFC above.

> Could it impact expected parsing results?

From how I understand the code, only positively. We now return a clear "unknown gylph" while before an invalid code point was returned.

k00ni:
Replaced by #841, because I had no permissions to push own commits into `cosmocode:fix/uchr-float-cast-warning`.

PR #827, opened 2026-06-19
Build PDF objects incrementally to lower peak memory

splitbrain:
Part 1 of #824 

parseData() used to decode every indirect object into a fully built raw object graph - deeply nested arrays that inflate the source roughly 8-10x - and return the whole thing before Parser turned any of it into PDFObjects. At that moment the entire raw graph and the PDF string were held simultaneously, which is the peak for most documents.

Replace parseData() with parseHeaderAndXref() plus a getObjectsStream() generator. Parser now builds each PDFObject and releases the raw structure before pulling the next one, so the full raw graph is never materialized, and the PDF string is freed before text extraction.

Backward compatibility note: the public RawDataParser::parseData() method is removed (it had no remaining callers). Third party consumers (I don't think there are any) that called it directly should switch to parseHeaderAndXref() + getObjectsStream(), or iterate the latter into an array for the old [$xref, $objects] return shape.

Measured peak memory (getText() over the whole file), vs master:

  samples/bugs/PullRequest457.pdf        147 MB -> 88 MB  (-40%)
  samples/DocumentWithLotsOfObjects.pdf   99 MB -> 62 MB  (-38%)

Documents whose peak occurs during text extraction rather than parsing (e.g. Issue356.pdf, Issue391.pdf) are unchanged. Extracted text is identical in all cases.

### DIFF
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -99,24 +99,36 @@ class Parser
      */
     public function parseContent(string $content): Document
     {
-        // Create structure from raw data.
-        list($xref, $data) = $this->rawDataParser->parseData($content);
+        // Normalize the raw data and decode the cross-reference/trailer table.
+        list($xref, $pdfData) = $this->rawDataParser->parseHeaderAndXref($content);
+
+        // The original (possibly un-trimmed) input is no longer needed; drop it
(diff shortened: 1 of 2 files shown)


k00ni:
Replaced by #842, because I had no permissions to push own commits into `cosmocode:mem/pipeline-objects`.

PR #828, opened 2026-06-19
Chunked getSectionsText() to lower peak memory

splitbrain:
Part 2 of #824 

getSectionsText() ran the formatted content stream through a single preg_split() into an array of every line, then kept only the handful needed for text positioning. A vector-graphics-heavy page can format to hundreds of thousands of lines while yielding almost no text, so that array was the peak of the whole text-extraction phase.

Split the formatted stream in, line-aligned 1 MB chunks instead. Each chunk is then preg_split(), filtered and then discarded. Where the source stream is smaller than 1 MB, there is no change in behaviour.

Output is byte-for-byte identical, and there is virtually no throughput cost.

Measured peak memory (parseFile + getText on the whole file) vs master, on a graphics heavy document:

  samples/bugs/Issue356.pdf   79.4 MB -> 59.2 MB  (-25%)

Other documents are unchanged (PullRequest457.pdf 146.9 MB, DocumentWithLotsOfObjects.pdf 99.1 MB, Issue391.pdf 10.4 MB).

### DIFF
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -429,71 +429,91 @@ class PDFObject
     {
         $sections = [];
 
-        // A cleaned stream has one command on every line, so split the
-        // cleaned stream content on \r\n into an array
-        $textCleaned = preg_split(
-            '/(\r\n|\n|\r)/',
-            $this->formatContent($content),
-            -1,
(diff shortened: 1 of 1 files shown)


k00ni:
Replaced by #840, because I had no permissions to push own commits into `cosmocode:mem/stream-sections`.

PR #829, opened 2026-06-19
Add opt-in content spooling to lower retained memory

splitbrain:
Part 3 of #824 

Add Config::setContentSpooling(): when enabled, each parsed object's decoded stream content is written to a single temporary file (ContentSpool) and read back on demand via PDFObject::getContent(), instead of being held in memory for the document's lifetime. The temp file is removed automatically once the document is destroyed.

This lowers the memory retained after parsing and during text extraction; extracted text and document details are byte-for-byte identical with the option on or off (verified across the sample suite).

Measured memory retained after parsing (getText output identical), vs master:

  samples/bugs/PullRequest457.pdf        71 MB -> 30 MB  (-58%)
  samples/DocumentWithLotsOfObjects.pdf  55 MB -> 32 MB  (-41%)
  samples/bugs/Issue356.pdf              25 MB ->  2 MB  (-93%)
  samples/bugs/Issue391.pdf               5 MB ->  2 MB  (-62%)

Note: on documents whose peak occurs while the raw object graph is being built, spooling alone does not lower the peak - it pairs with incremental object construction (#827), which removes that earlier peak.

### DIFF
--- a/doc/CustomConfig.md
+++ b/doc/CustomConfig.md
@@ -26,6 +26,7 @@ The `Config` class has the following options:
 | `setPdfWhitespaces`      | String  | `\0\t\n\f\r `   |                                                                                                                                                      |
 | `setPdfWhitespacesRegex` | String  | `[\0\t\n\f\r ]` |                                                                                                                                                      |
 | `setRetainImageContent`  | Boolean | `true`          | If parsing fails due to memory exhaustion, you can set the value to `false`. This will reduce memory usage, although it will no longer retain image content. |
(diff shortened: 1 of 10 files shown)


k00ni:
@splitbrain Sorry, but I am going to decline this one. The project is under limited maintenance, and we explicitly rule out new features and performance projects ([CONTRIBUTING.md](https://github.com/smalot/pdfparser/blob/master/CONTRIBUTING.md)). Unlike #827 and #828, which are internal optimizations without new API surface, this PR adds a feature that has to be maintained from now on: a new config option, a new class, new public methods on `Document` and `PDFObject`, and temp file handling.

Issue #712, opened 2024-05-15
Possibility to improve parsing perfomance by not using uniqid() and add an early return?


bernemann:
I parse PDFs where strings are encoded in the bracket notation like

`[(Xxxx)1.7(a)1.2(tt )-10.1(\374)1.1(be)1.4(r )-8.5(Xxxx)-7.8(xx)-9.5(xxx)2.2(xxx)-1.1(x)-9(xxx)2( )]TJ`

I have an idea to improve parsing performance which reduced parsing time from 1.65 seconds to 0.54 seconds in my example: 

First, I observed a significant performance increase when calling `Page::getTextArray()` on documents with strings like this with a simpler placeholder in `PDFObject::formatContent()` like

 `$id = "S_$i";`  where `$i` is a simple counting integer instead of `$id = uniqid('STRING_', true);`

Reason for this are the latter numerous calls to `str_replace()` in order to replace back the placeholders with the original content. The complex placeholder generated by `uniqid()` seems to slow down the calls. 

Is there any reason for using this complex placeholder or can it be replaced by a much simpler one?

Second, I recognized 2 (more or less) subsequent calls to `PDFObject->getTextArray($this)` in `Page::getTextArray()` around line where the first one might immediatlly return: 

https://github.com/smalot/pdfparser/blob/a19d5550531796aa1d2d351a00066e422c848123/src/Smalot/PdfParser/Page.php#L346-L350

Is there a reason for not returning in the `try` case and instead return the same call later

https://github.com/smalot/pdfparser/blob/a19d5550531796aa1d2d351a00066e422c848123/src/Smalot/PdfParser/Page.php#L365

?

Thank you for your time!

k00ni:
Sounds reasonable and interesting so please create a pull request and lets discuss there.

> Is there any reason for using this complex placeholder or can it be replaced by a much simpler one?

Without a deeper look into the code, it doesn't seem necessary to use an expensive function such as `uniqid()` here.

GreyWyvern:
The reason for the Unique ID is that the simpler the replacement string, the higher the chance it may appear elsewhere in the document and cause the wrong text to be restored later. For example, `@@@S_2@@@` has a higher chance of appearing as regular document stream text than `@@@STRING_b64b8a7ad6d4c@@@` etc. Maybe not all *that* much higher, but still higher.

That being said, if the `uniqid()` calls can be replaced with a system that uses less CPU, but still keeps the complexity of the replacement string, I'm all for that.

[...]

PR #831, opened 2026-07-04
Avoid quadratic placeholder replacement in PDFObject::formatContent()

bernemann:
# Type of pull request

* [ ] Documentation update
* [X] Performance

# About

This is a follow-up to the performance issue I opened in https://github.com/smalot/pdfparser/issues/712.

As described there, parsing gets slow on PDFs whose text is encoded as bracket-notation TJ arrays like [(Xxxx)1.7(a)1.2(tt )-10.1(\374)...], because every string operand becomes a placeholder in PDFObject::formatContent() and is then replaced back with its own str_replace() over the whole content stream. With many small operands this becomes quadratic and dominates the parsing time.

Instead of shortening the placeholder (which I first suggested, but which could collide with real stream content), I kept the uniqid() placeholders and only changed how they are handled:

First, the placeholders are now restored in a single strtr() pass instead of one str_replace() per placeholder. strtr() does all replacements in one traversal, so the cost no longer grows with the number of operands.

Second, the initial extraction now uses strpos() + substr_replace() instead of preg_quote() + preg_replace(), which avoids compiling a regex for every operand.

In my measurements this brings the affected PDFs down from about 8.3 seconds to about 2.1 seconds (roughly 3.5-4x), while documents that were already fast stay the same. The extracted text from getTextArray() is byte-identical before and after on all my sample files.

Thank you for your time!

### DIFF
--- a/src/Smalot/PdfParser/PDFObject.php
+++ b/src/Smalot/PdfParser/PDFObject.php
@@ -320,12 +320,18 @@ class PDFObject
                 // Replace the string with a unique placeholder
                 $id = uniqid('STRING_', true);
                 $pdfstrings[$id] = $text[0];
-                $content = preg_replace(
-                    '/'.preg_quote($text[0], '/').'/',
-                    '@@@'.$id.'@@@',
-                    $content,
-                    1
-                );
+                // Replace the first literal occurrence without compiling a regex
+                // per string operand (preg_quote + preg_replace is quadratic for
+                // kerned TJ arrays with many operands).
+                $stringPos = strpos($content, $text[0]);
+                if (false !== $stringPos) {
+                    $content = substr_replace(
+                        $content,
+                        '@@@'.$id.'@@@',
+                        $stringPos,
+                        \strlen($text[0])
+                    );
+                }
 
                 // Reset to search for the next string
                 $attempt = '(';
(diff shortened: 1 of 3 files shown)


k00ni:
Thank your for the PR.

Could you provide some test code which demonstrates the statement:

> In my measurements this brings the affected PDFs down from about **8.3 seconds to about 2.1** seconds (roughly 3.5-4x), while documents that were already fast stay the same.

Ideally the check code is located somewhere in https://github.com/smalot/pdfparser/tree/master/tests/Performance. We already have a bunch of performance tests.

---

Could you also add at least one PHPUnit test which demonstrates that your new code works as expected?

bernemann:
Thanks for the review. I added tests for the `formatContent()` changes:

**Unit tests** (PDFObjectTest.php)

- Kerned TJ array split into many small string operands, reassembled in the right order.
- A `<< >>` BDC dictionary wrapping a text block, making sure it doesn't swallow the following text.
- A string containing balanced unescaped parentheses, checking it survives extraction.

**Performance test** (KernedTjArrayFormatContentTest.php)

Builds a content stream with 20,000 kerned string operands and extracts its text. The old restoration ran one full-content `str_replace()` per placeholder, the new code does a single `strtr()` pass. 
 
Rough numbers:
 
| operands | old (per-placeholder `str_replace`) | new (single `strtr`) |
|---------:|------------------------------------:|---------------------:|
| 5,000    | ~0.9s                               | ~0.1s                |
| 10,000   | ~3.2s                               | ~0.4s                |
| 20,000   | ~10s                                | ~1.5s                |

I have a standalone script to reproduce the figures which I left out of the commit. I can provide it if requested.

k00ni:
Thank you again @bernemann for adding the tests. I measured `formatContent()` with 20,000 operands: 6.1 s on master, 0.52 s with your changes. The output is identical to master for all 59 sample files.

I pushed one commit on top before merging:

- The extraction of strings still grew quadratically (each operand caused a search from the start and a copy of the stream). It is linear now: your performance scenario with 60,000 operands went from 10.6 s to 0.64 s.
- `KernedTjArrayFormatContentTest` was not registered in `runPerformanceTests.php`, so it never ran. It does now, and it fails if no text gets extracted.
- The unit tests share a helper, and I added a test for edge cases of `formatContent()` (escaped and nested parentheses, unterminated string, end-of-line markers in strings, dictionaries containing strings).

PR #836, opened 2026-09-01
Avoid recursive matching for object stream indexes

### LINKED ISSUE #835: Object stream parsing can exhaust PCRE JIT and read a missing capture
## Description

`Parser::parseObject()` separates an object-stream index from its body with a nested repeated regular expression, ignores the result of `preg_match()`, and then unconditionally reads capture 3.

On PHP 8.5 with PCRE 10.47, a synthetic object stream containing about 4,000 object/offset pairs makes the match return `false` with `JIT stack limit exhausted`. The subsequent array access raises `Undefined array key 3`.

The expression also eventually reaches the recursion limit with JIT disabled, so changing PCRE limits is not a durable solution.

## Minimal reproduction

The existing `ParserSub::exposedParseObject()` test helper can reproduce this without a PDF fixture by passing an `ObjStm` structure containing 4,000 synthetic `object-number offset` pairs followed by simple `null` objects.

## Expected behavior

Use the object-stream dictionary's required `/N` and `/First` entries to bound and separate the index, parse exactly `N` object/offset pairs with non-recursive logic, and reject malformed counts or offsets with a deterministic exception.

No PDF content or customer data is needed for the regression.

NiekNijland:
Fixes #835.

The object-stream parser currently separates the object index with a nested, recursive PCRE pattern. Large valid indexes can exhaust the PCRE JIT stack; preg_match then returns false and the parser reads missing captures.

This change:
- uses the required ObjStm N and First metadata to separate the index from object bodies;
- tokenizes the bounded index with a non-recursive whitespace split;
- validates counts, numeric values, offsets, and duplicate offsets before parsing embedded objects;
- preserves existing object-stream behavior for valid PDFs.

Verification:
- regression fails before the change with Undefined array key warnings and 0 parsed objects;
- focused object-stream tests: 3 tests, 6 assertions;
- full PHPUnit suite: 195 tests, 1189 assertions;
- PHP-CS-Fixer dry-run: clean;
- PHPStan: clean.

### DIFF
--- a/src/Smalot/PdfParser/Parser.php
+++ b/src/Smalot/PdfParser/Parser.php
@@ -177,24 +177,37 @@ class Parser
                     $content = isset($part[3][0]) ? $part[3][0] : $part[1];
 
                     if ($header->get('Type')->equals('ObjStm')) {
-                        $match = [];
+                        $numberOfObjects = $this->objectStreamInteger($header, 'N');
+                        $firstObjectOffset = $this->objectStreamInteger($header, 'First');
 
-                        // Split xrefs and contents.
-                        preg_match('/^((\d+\s+\d+\s*)*)(.*)$/s', $content, $match);
-                        $content = $match[3];
+                        if ($firstObjectOffset > \strlen($content)) {
+                            throw new \UnexpectedValueException('Object stream First offset exceeds its content length.');
+                        }
 
-                        // Extract xrefs.
-                        $xrefs = preg_split(
-                            '/(\d+\s+\d+\s*)/s',
-                            $match[1],
-                            -1,
-                            \PREG_SPLIT_NO_EMPTY | \PREG_SPLIT_DELIM_CAPTURE
+                        if ($numberOfObjects > $firstObjectOffset) {
+                            throw new \UnexpectedValueException('Object stream N exceeds its index length.');
+                        }
+
+                        $xrefs = $this->parseObjectStreamIndex(
+                            substr($content, 0, $firstObjectOffset),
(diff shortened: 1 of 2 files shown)


k00ni:
@NiekNijland first of all, sorry for the late response and thank you for the PR.

The bug is real and your approach via `/N` and `/First` is the right one. It is even worse than described in #835: from about 4,000 pairs on (2,000 on PHP 7.1) all objects of the object stream get lost silently.

Your approach also fixes a second bug you didn't mention: if integer objects follow the index, master takes them for pairs. `samples/DocumentWithLotsOfObjects.pdf` has such an object stream (`12815_0`, `/N 200`, master reads 274 pairs), which results in shifted objects and broken fonts.

I pushed a commit which replaces most of the implementation though, for two reasons:

- **Exceptions.** I know you asked for deterministic exceptions in the issue, but an exception in `parseObject()` aborts `parseContent()`, so one odd object stream costs the whole document. I compared 21 malformed, but readable object streams: master reads all of them, your version throws in 13 (e.g. `/N` off by one, `/First` as real number, a comment behind the index). This library rather returns something incomplete than nothing, and I don't want to change that for existing users (because we are in a legacy-kinda mode).
- **Size.** 8 private methods with a character based tokenizer and its own overflow check are a lot to maintain for this fix. It is one non-recursive `preg_match_all()` now.

Further information in the related commit message.

If you hit this with a real document, I'd be glad to hear whether this version handles it. I will leave this PR open for short while before I merge it. Unless you have an objection.

